PR #57, opened 2026-04-01
feat: CFN auto-generated names follow AWS {stack}-{logicalId}-{SUFFIX…

Nahuel990:
- Added _physical_name() helper with service-specific casing and length limits
  - S3: lowercase max 63, SQS: max 80, DynamoDB: max 255, Lambda/IAM/EB: max 64
  - Fixes CDK stacks that omit explicit names producing untraceble cfn-xxx names

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,17 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ## [1.1.15] — 2026-04-01
 
+### Added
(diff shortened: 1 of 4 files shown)


Nahuel990:
Resource Type | Max Length | Notes / Gotchas
-- | -- | --
S3 Bucket | 63 characters | Lowercase only, no underscores, must be globally unique. LocalStack might allow slightly invalid names, but AWS will reject.
Lambda Function | 64 characters | Includes letters, numbers, hyphens. LocalStack may allow longer names in testing, AWS won’t.
DynamoDB Table | 255 characters | Names are case-sensitive. Some LocalStack versions truncate automatically.
CloudFormation Stack | 128 characters | Stack names can’t include certain symbols.
IAM Roles / Policies | 64–128 characters | Policies often have path restrictions.
API Gateway Rest API | 128 characters | LocalStack may allow longer names, AWS rejects.

https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-cloudformation-stack.html

PR #71, opened 2026-04-02
fix: SNS Subscribe now applies Attributes (RawMessageDelivery, FilterPolicy)

### LINKED ISSUE #70: SNS Subscribe ignores Attributes parameter (RawMessageDelivery, FilterPolicy)
The _subscribe function in sns.py does not parse the Attributes parameter from the Subscribe request. This means subscription attributes like RawMessageDelivery and FilterPolicy passed during subscription creation are silently ignored.

**Current behavior:**

The `_subscribe` function hardcodes RawMessageDelivery: "false" (line 218) and never reads the Attributes.* entries from the request params:

```
  sub = {
      ...
      "attributes": {
          ...
          "RawMessageDelivery": "false",  # always false
      },
  }
```

When the AWS SDK sends a subscribe request with attributes like:

Attributes.entry.1.key=RawMessageDelivery&Attributes.entry.1.value=true
Attributes.entry.2.key=FilterPolicy&Attributes.entry.2.value={"event":["MyEvent"]}

These are ignored. The only way to set them is via a separate SetSubscriptionAttributes call after subscribing.

**Expected behavior:**

The _subscribe function should parse Attributes.* from the request params and merge them into the subscription attributes, similar to how `_set_subscription_attributes` already handles them (line 338).

Something like:

```
  # After building the sub dict, apply any Attributes passed in the request
  for key in params:
      if key.startswith("Attributes.entry."):
          # parse and apply
```

santiagodoldan:
Closes #70

### DIFF
--- a/ministack/services/sns.py
+++ b/ministack/services/sns.py
@@ -215,6 +215,13 @@ def _subscribe(params):
         },
     }
 
+    i = 1
+    while _p(params, f"Attributes.entry.{i}.key"):
+        key = _p(params, f"Attributes.entry.{i}.key")
+        val = _p(params, f"Attributes.entry.{i}.value")
+        sub["attributes"][key] = val
+        i += 1
+
     topic["subscriptions"].append(sub)
     _sub_arn_to_topic[sub_arn] = topic_arn
     _refresh_subscription_counts(topic)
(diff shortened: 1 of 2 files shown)


Nahuel990:
Hello I'm merging this for speed, but I will do a minor correction before push to v 18  to filter by allowed_attrs to match the validation in _set_subscription_attributes

   val = _p(params, f"Attributes.entry.{i}.value")
        sub["attributes"][key] = val
        i += 1
        
        
        Thanks!

PR #85, opened 2026-04-03
fix: align Route53 record-set ordering and pagination cursors

jgrumboe:
While running my own Terraform modules against Ministack, I encountered an error that Terraform would run in an endless loop when creating `aws_route53_record` resources.

This PR fixes two Route53 `ListResourceRecordSets` behavior mismatches with AWS that were blocking Terraform flows:
- **Correct DNS name ordering for start cursors**
  - Route53 sorts by reversed labels (for example, `com.example.www.`), not plain string order.
  - Updated listing sort/filter logic to use reversed-label ordering for `StartRecordName` + `StartRecordType`.
  - This prevents parent-zone records from being incorrectly returned ahead of an exact child-name start point.
- **Correct pagination cursor semantics**
  - When `IsTruncated=true`, `NextRecordName` / `NextRecordType` must point to the **first record of the next page**, not the last record of the current page.
  - Fixed cursor generation so `MaxItems=1` no longer returns self-referential `NextRecord*` values.
## Test coverage added/updated
- `test_route53_list_resource_record_sets_start_name_uses_reversed_label_order`
- `test_route53_list_resource_record_sets_truncated_next_record_uses_next_page_start`
- `test_route53_list_resource_record_sets_pagination_advances_with_next_record_cursor`
These tests verify ordering, correct next-page cursor generation, and end-to-end forward pagination progress.

Disclaimer: Yes, I used opencode to create this PR. I tested it locally with Terraform.

### DIFF
--- a/ministack/services/route53.py
+++ b/ministack/services/route53.py
@@ -134,6 +134,12 @@ def _normalise_name(name: str) -> str:
     return name if name.endswith(".") else name + "."
 
 
+def _name_sort_key(name: str) -> tuple[str, ...]:
+    """Route53 sorts names by labels reversed (e.g. com.example.www.)."""
+    labels = _normalise_name(name).rstrip(".").split(".")
+    return tuple(reversed(labels))
+
+
 # ─── default records (SOA + NS) ───────────────────────────────────────────────
 
 _DEFAULT_NS = [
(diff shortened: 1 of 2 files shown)


Nahuel990:
:O What a sin to use AI :D no worries, thanks very much for this! I am merging I will tag it alongside v 25.

Nahuel990 (approved):

jgrumboe:
> :O What a sin to use AI :D no worries, thanks very much for this! I am merging I will tag it alongside v 25.

Thanks very much.
I thought it's worth mentioning that I used AI, without it I would have needed a lot more time to understand the codebase. But it also means that the output is sometimes garbage, even if it looks correct. 😉 

Great project! Keep it up!

Issue #92, opened 2026-04-03
Add support for AWS CloudFront Service


Nikhiladiga:
### AWS Service name

AWS CloudFront

### Operations needed

- CreateDistribution
- GetDistribution
- GetDistributionConfig
- ListDistributions
- UpdateDistribution
- DeleteDistribution
- CreateInvalidation
- ListInvalidations
- GetInvalidation

### Use case

Local development and CI/CD testing of CloudFront distribution CRUD, cache invalidations, and origin configuration without hitting real AWS.

Implementation details:
- REST/XML service file (`ministack/services/cloudfront.py`) dispatching on `(method, path)` using regex matching (same pattern as Route53)
- XML responses with namespace `http://cloudfront.amazonaws.com/doc/2020-05-31/`
- ETag-based optimistic concurrency (issued on create/get, required on update/delete)
- DistributionNotDisabled (409) guard on delete while enabled
- PreconditionFailed (412) on ETag mismatch
- In-memory state for distributions and invalidations
- Python stdlib only — no new dependencies
- 12 integration tests covering all 9 operations + error cases

### Would you like to contribute this?

- [x] Yes, I'd like to implement this service and open a PR
- [ ] I can help review/test an implementation

Nahuel990:
Released in v1.1.26. Please reopen if issue persist

PR #93, opened 2026-04-03
#92: [FEAT] Add CloudFront service

Nikhiladiga:
https://github.com/Nahuel990/ministack/issues/92
--

Adds CloudFront API emulation with 9 operations:

- Distribution CRUD (CreateDistribution, GetDistribution, GetDistributionConfig, ListDistributions, UpdateDistribution, DeleteDistribution)
- Invalidations (CreateInvalidation, ListInvalidations, GetInvalidation)
- ETag-based optimistic concurrency control (If-Match header enforcement)
- Proper error handling (NoSuchDistribution 404, PreconditionFailed 412, DistributionNotDisabled 409, MalformedXML 400)
- REST/XML API under /2020-05-31/ path prefix
- In-memory storage with full reset support

Tested locally: 
- all 12 boto3 integration tests passed
- curl verified. 
- Includes 12 integration tests covering distributions, invalidations, updates with ETag, delete-enabled guard, and ETag mismatch.

### DIFF
--- a/tests/test_services.py
+++ b/tests/test_services.py
@@ -17254,3 +17254,194 @@ def test_persist_kms_roundtrip():
     assert key_id in _kms._keys
     assert _kms._keys[key_id]["Description"] == "persist-key"
     _kms._keys.pop(key_id)
+
+# ========== CloudFront ==========
+
+_CF_DIST_CONFIG = {
+    "CallerReference": "cf-test-ref-1",
+    "Origins": {
+        "Quantity": 1,
+        "Items": [
+            {
+                "Id": "myS3Origin",
+                "DomainName": "mybucket.s3.amazonaws.com",
+                "S3OriginConfig": {"OriginAccessIdentity": ""},
+            }
+        ],
+    },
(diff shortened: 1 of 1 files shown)


Nahuel990:
Due to upcoming weekend I'll review this in a few hours. So far looks good! Thank you!!

Nahuel990:
Hi @Nikhiladiga really appreciate the enthusiasm! I’ll work on putting together a list of issues. So far, our coverage looks strong. Given the number of downloads compared to the relatively low number of open issues, that’s a good sign.
As you may have seen, we have close to 1,000 tests in place, not just for individual services, but also for integration and end-to-end scenarios.

I think the focus will be now in Cloudformation (which is the only service that has a dedicated folder instead of a py file due to its complexity. 

To have an idea look at what the LocalStack Community offered before the license change (you can still pointing at v4.14.0) and we will aim to match that with minor improvements. Many people try to go straight to Bedrock but that will easily raise our image size in around 600Mb so unless it's requested by a lot of people (300+) I don't see the point.

PR #151, opened 2026-04-06
Fix Step Functions timestamp wire compatibility

jayjanssen:
## Why
MiniStack Step Functions responses currently return ISO timestamp strings. AWS SDK clients expect JSON timestamp numbers for these fields, so Go SDK v2 and botocore can fail to deserialize responses before workflows even start.

## What
- Normalize Step Functions timestamp fields at the response boundary for state machine, execution, history, sync, stop, and activity APIs
- Keep internal Step Functions state unchanged and apply the wire-format fix centrally in `handle_request`
- Add SDK-facing regression tests covering the affected timestamp-bearing APIs

## Risk Assessment
Low. The change is isolated to Step Functions response serialization, and the internal interpreter and persistence behavior are unchanged.

## References
- Verified with `python3 -m py_compile ministack/services/stepfunctions.py tests/test_services.py`
- Verified with `pytest tests/test_services.py -v --tb=short -k 'sfn or stepfunctions'` (48 passed)

Generated with Codex

### DIFF
--- a/ministack/services/stepfunctions.py
+++ b/ministack/services/stepfunctions.py
@@ -131,6 +131,55 @@ if _restored:
     restore_state(_restored)
 
 
+_TIMESTAMP_RESPONSE_FIELDS = {
+    "creationDate",
+    "redriveDate",
+    "startDate",
+    "stopDate",
+    "timestamp",
+    "updateDate",
+}
+
+
+def _timestamp_response_value(value):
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks @jayjanssen for this implementation! I will merge it and release it as part of v 42

PR #177, opened 2026-04-07
EventBus fixes: Add UpdateEventBus, numeric timestamps, and policy enhancements

jgrumboe:
To fully support `aws_cloudwatch_event_*` Terraform resources I've fixed the following:

- Switch CreationTime/LastModifiedTime to numeric epoch timestamps; coerce legacy ISO strings on restore
  The Terraform error was: `Error: reading EventBridge Event Bus (platforms): operation error EventBridge: DescribeEventBus, https response error StatusCode: 200, RequestID: c10b13d6-1d4a-46e7-bdcb-6328199af655, deserialization failed, failed to decode response body, expected Timestamp to be a JSON Number, got string instead` 
- Implement UpdateEventBus handler and register it in request dispatch
- Return EventBus Policy as a JSON string (empty when absent)
- Make PutPermission accept optional StatementId (auto-generate) according to AWS API spec
- Update related tests to assert numeric timestamps and adjust expectations

### DIFF
--- a/ministack/services/eventbridge.py
+++ b/ministack/services/eventbridge.py
@@ -21,7 +21,7 @@ import os
 import re
 import threading
 import time
-from datetime import datetime, timezone
+from datetime import datetime
 
 from ministack.core.responses import error_response_json, json_response, new_uuid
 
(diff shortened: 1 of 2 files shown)


Nahuel990:
Massive PR! Thanks @jgrumboe merging and releasing as part of v49!

Nahuel990 (approved):

PR #214, opened 2026-04-09
feat: use botocore serializer for query-protocol SFN aws-sdk dispatch

jayjanssen:
## Problem

The query-protocol `aws-sdk:*` dispatcher in Step Functions flattens JSON parameters naively into query params. This breaks when SDK model parameter names differ from wire-format names — which is the case for most query-protocol services.

Examples:
- `DbSubnetGroupName` (SDK) → `DBSubnetGroupName` (wire)
- `DbClusterParameterGroupName` → `DBClusterParameterGroupName`
- `SubnetIds` list → `SubnetIds.SubnetIdentifier.N` (not `SubnetIds.member.N`)

## Solution

Use botocore's serializer, which has the full AWS service model metadata, to correctly serialize query-protocol requests. This handles:

1. **Parameter name mapping** via case-insensitive matching from SFN SDK-style names to botocore model names
2. **List member naming** — botocore uses the correct model-specific member names (e.g. `SubnetIdentifier`) instead of generic `member`
3. **Action name casing** — PascalCase from camelCase SFN ARN names (subsumes #204)

Also updates `_parse_member_list` in rds.py to handle both `Prefix.member.N` (direct API) and `Prefix.<MemberName>.N` (botocore serializer) list formats.

Falls back to naive flattening if botocore can't handle a service.

## Dependencies

Adds `botocore` as a runtime dependency (already a transitive dep via `docker` package).

## Testing

- All 1099 existing tests pass
- Verified against a Terraform provider acceptance test suite: parameter groups, subnet groups, and cluster creation all work correctly through SFN dispatch.

### DIFF
--- a/ministack/services/rds.py
+++ b/ministack/services/rds.py
@@ -1969,16 +1969,33 @@ def _parse_tags(params):
 
 
 def _parse_member_list(params, prefix):
-    """Parse Prefix.member.N style lists from query params."""
+    """Parse list params in either Prefix.member.N or Prefix.<MemberName>.N format.
+
+    The member.N format is used by direct AWS CLI/SDK calls. The <MemberName>.N
+    format is produced by botocore's serializer when dispatched via Step Functions
+    aws-sdk integrations (e.g. SubnetIds.SubnetIdentifier.N).
+    """
     items = []
     i = 1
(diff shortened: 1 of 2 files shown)


jayjanssen:
working on an alternative

PR #218, opened 2026-04-09
feat: SFN response fidelity — botocore parser, key naming convention, intrinsics

jayjanssen:
## Summary

Improves SFN `aws-sdk` integration response fidelity to match real AWS Step Functions behavior. This is part of an effort to pass Terraform acceptance tests for an Aurora provisioner that uses SFN state machines with aws-sdk integrations.

### Botocore response parser for query-protocol
- Query-protocol XML responses are now deserialized through botocore's response parser (`_parse_query_response_via_botocore`), producing correctly typed values (int, bool) and proper SDK member names instead of raw XML element names.
- Removes the duplicate hand-rolled XML parser functions in favor of the botocore parser.

### SFN key naming convention
- API response keys like `DBClusters`, `DBClusterArn` are converted to Java SDK V2 convention (`DbClusters`, `DbClusterArn`) via `_api_name_to_sfn_key` / `_convert_keys_to_sfn_convention`.
- Applied to both query-protocol and JSON-protocol aws-sdk dispatchers.
- This fixed an infinite `DescribeDBClusters` polling loop where the caller expected `DbClusters` but received `DBClusters`.

### New intrinsic functions
- `States.ArrayGetItem(array, index)` — extract element from array by index
- `States.Array(val1, val2, ...)` — construct array from arguments  
- `States.ArrayLength(array)` — return length of array

### Bug fixes
- `States.TaskFailed` in Retry/Catch blocks now acts as a catch-all error matcher (matches any Task error), consistent with AWS behavior.
- `datetime` objects from botocore response parser are serialized to ISO-8601 strings.
- Map state `ItemSelector` paths prefixed with `$` now resolve against the Map's effective input, not the individual item.
- RDS `ModifyDBCluster` accepts `EnableHttpEndpoint` parameter (stub, accepted and ignored).

### Tests
- Updated SFN RDS test assertions to use SDK convention key names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Added
+- **States.ArrayGetItem, States.Array, States.ArrayLength intrinsics** — SFN state machines using `States.ArrayGetItem(array, index)`, `States.Array(val1, val2, ...)`, and `States.ArrayLength(array)` now execute correctly.
+- **Botocore response parser for query-protocol aws-sdk dispatch** — query-protocol responses are now deserialized through botocore's response parser, producing correctly typed values (int, bool) and proper SDK member names instead of raw XML element names.
+- **SFN key naming convention (`_convert_keys_to_sfn_convention`)** — API response keys like `DBClusters` are now converted to Java SDK V2 convention (`DbClusters`) matching real AWS SFN behavior. Applied to both query-protocol and JSON-protocol aws-sdk dispatchers.
+
+### Fixed
+- **`States.TaskFailed` treated as catch-all** — `Retry` and `Catch` blocks matching `States.TaskFailed` now catch any Task error, matching AWS behavior where it acts as a wildcard error matcher.
+- **`datetime` objects in botocore responses** — botocore response parser returns `datetime` objects for timestamp fields; these are now serialized to ISO-8601 strings before JSON encoding.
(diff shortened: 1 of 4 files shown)


Nahuel990:
@jayjanssen Thanks for the contribution! We’re planning to move the repository to the `ministackorg` org on Saturday, so we won’t be adding new features tomorrow.

I’ll merge any remaining PRs, and from Sunday we can resume working on enhancements to MiniStack.

Tomorrow we will only do bugfix in case of any

Nahuel990:
@jayjanssen Thanks for the thorough work on this. We cherry-picked everything except the botocore parts into
  v1.1.60:                                                                                                        
          
  Included:                                                                                                       
  - _api_name_to_sfn_key / _convert_keys_to_sfn_convention (fixes your infinite DescribeDBClusters polling loop)  
  - Key conversion applied to both query-protocol and JSON-protocol aws-sdk dispatchers                           
  - States.ArrayGetItem, States.Array, States.ArrayLength intrinsics                                              
  - States.TaskFailed catch-all in Retry/Catch blocks                                                             
  - datetime → ISO-8601 serialization                                                                             
  - Map state ItemSelector fix ($ paths resolve against effective input, not individual item)
  - RDS EnableHttpEndpoint stub                                                                                   
  - Updated test assertions for SDK convention key names    
                                                                                                                  
  Not included:                                             
  - _parse_query_response_via_botocore and removal of hand-rolled XML parser — botocore adds ~30% to our Docker   image size and pulls in dependencies for services we don't support yet. We want to keep the install footprint   minimal                                                                                                        
                                                                                                                  
  The key naming fix should unblock your Terraform Aurora provisioner's polling loop. The remaining gap without
  botocore is typed values (int vs string) and single-element XML list parsing, if those become blockers for your
   use case let us know and we can look at lighter alternatives.

jayjanssen:
Not a problem!

PR #232, opened 2026-04-10
fix(lambda): Docker executor mounts provided-runtime code at wrong path

jayjanssen:
## Why

`_execute_function_docker()` fails for `provided`/`provided.al2023` runtimes because it mounts Lambda code only at `/var/task` and overrides CMD to `["/var/task/bootstrap"]`. The AWS RIE entrypoint (`/lambda-entrypoint.sh`) in `public.ecr.aws/lambda/provided:al2023` expects the bootstrap binary at `/var/runtime/bootstrap`, so the container always errors out.

## What

- Mount code at both `/var/task` (user code) and `/var/runtime` (bootstrap) using `docker.types.Mount`, matching real AWS Lambda layout
- Pass `"bootstrap"` as CMD instead of `"/var/task/bootstrap"` so the RIE entrypoint resolves the handler correctly
- Add integration test with a shell-script bootstrap implementing the Lambda Runtime API protocol (gated on `LAMBDA_EXECUTOR=docker`)

## Risk Assessment

Low — only affects Docker executor path for provided runtimes (`LAMBDA_EXECUTOR=docker`). No change to the default native executor or Python/Node.js Docker paths.

## References

- The RIE entrypoint (`/lambda-entrypoint.sh`) unconditionally runs `exec /var/runtime/bootstrap`: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html
- Verified: test fails without fix (`KeyError: statusCode`), passes with fix
- Full suite: 1106 passed (2 pre-existing failures unrelated to this change)

Generated with Amp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Amazing work with the Lambda gate, thanks @jayjanssen !

PR #234, opened 2026-04-10
fix(apigateway): invoke Node.js Lambdas via warm worker in v1/v2 proxy

bognari:
## Summary

API Gateway v1 and v2 `_invoke_lambda_proxy` only dispatched to the warm worker pool for Python runtimes. Node.js Lambdas received a hardcoded `"Mock response"` instead of being executed — even though `lambda_runtime.py` already supports Node.js via `_NODEJS_WORKER_SCRIPT`.

## Changes

- **`ministack/services/apigateway.py`** (v2): `startswith("python")` → `startswith(("python", "nodejs"))`
- **`ministack/services/apigateway_v1.py`** (v1): same fix
- **`tests/test_lambda.py`**: added `test_apigwv2_nodejs_lambda_proxy` — creates a Node.js Lambda behind an API Gateway v2 HTTP API route and verifies the actual handler response is returned

## Motivation

We use MiniStack with Terraform to deploy a Node.js Lambda backend (API Gateway v2 HTTP API + ~25 Lambda functions). Direct `aws lambda invoke` worked correctly, but requests through the API Gateway execute-api endpoint returned `"Mock response"` for all Node.js handlers.

The root cause is a runtime check in both `_invoke_lambda_proxy` functions that only matches `"python"`:

```python
# Before:
if code_zip and func_data["config"]["Runtime"].startswith("python"):

# After:
if code_zip and runtime.startswith(("python", "nodejs")):
```

## Test plan

- [x] `pytest tests/test_lambda.py -v -k "apigwv2_nodejs"` — new test passes
- [x] `pytest tests/test_lambda.py -v` — all 69 existing tests pass, 1 skipped (docker-only)
- [x] Verified end-to-end: Terraform-deployed Node.js Lambda returns real handler response via API Gateway v2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 ### Fixed
+- **API Gateway v1/v2 returns mock response for Node.js Lambdas** — `_invoke_lambda_proxy` in both `apigateway.py` (v2) and `apigateway_v1.py` (v1) only dispatched to the warm worker pool for Python runtimes. Node.js Lambdas received a hardcoded `"Mock response"` instead of being executed, even though the warm worker pool in `lambda_runtime.py` already supports Node.js. Now checks for both `python` and `nodejs` runtimes.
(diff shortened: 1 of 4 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this quick workaround @bognari ! Welcome to Ministack! We will be releasing this as part of v62 version.

PR #236, opened 2026-04-10
feat(lambda): extract layers in warm worker pool

bognari:
## Summary

The warm worker pool (`Worker._spawn` in `lambda_runtime.py`) only extracted the function's code ZIP but ignored attached Lambda Layers. Any dependency shipped via a layer caused `ModuleNotFoundError` at runtime — even though the layer data was correctly stored via `PublishLayerVersion`.

## Changes

**`ministack/core/lambda_runtime.py`:**

- `Worker._spawn()` now iterates `config["Layers"]`, resolves each layer's ZIP via `lambda_svc._resolve_layer_zip()`, and extracts it into a temp directory alongside the function code
- Sets `_LAMBDA_LAYERS_DIRS` env var so the existing worker scripts can find layer packages
- Sets `NODE_PATH` for ESM `import()` compatibility in Node.js handlers
- Adds `sys.path` layer expansion to `_PYTHON_WORKER_SCRIPT`, mirroring the existing logic in `_WRAPPER_SCRIPT` (used by the subprocess executor)

**`tests/test_lambda.py`:**

- Added `test_lambda_warm_worker_uses_layer` — publishes a Python layer with a helper module, creates a function using that layer, and verifies the handler can import and use the layer module

## Motivation

We deploy Node.js Lambda functions with a shared dependency layer (node_modules) via Terraform. Direct `aws lambda invoke` worked, but the warm worker pool did not extract layers — causing `Cannot find package` errors. The subprocess-based executor (`_execute_function_local`) already had full layer support; this PR brings the same capability to the warm worker pool.

## Test plan

- [x] `pytest tests/test_lambda.py -v -k "warm_worker_uses_layer"` — new test passes
- [x] `pytest tests/test_lambda.py -v` — all 69 existing tests pass, 1 skipped (docker-only)
- [x] Verified end-to-end: Terraform-deployed Node.js Lambda with shared dependency layer resolves packages correctly via warm worker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ### Added
 - **SFN query-protocol acronym mapper** — Step Functions `aws-sdk:*` integrations now correctly convert SDK-style parameter names (e.g. `DbSubnetGroupName`) to wire-format names (`DBSubnetGroupName`) for query-protocol services (RDS, EC2, IAM, STS, etc.). Uses a static acronym mapping — no botocore dependency. Contributed by @jayjanssen.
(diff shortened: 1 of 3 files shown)


Nahuel990:
Thanks for this @bognari I will merge this post migration tomorrow and release it in v1.2.0

Nahuel990 (approved):

PR #235, opened 2026-04-10
feat: static acronym mapper for SFN query-protocol param name conversion

jayjanssen:
## Summary

Step Functions `aws-sdk:*` integrations send parameters using Java SDK V2 naming (e.g. `DbSubnetGroupName`) but query-protocol services expect wire-format names (e.g. `DBSubnetGroupName`). This PR adds a lightweight static acronym mapper to convert parameter names before dispatch.

Replaces the botocore-based serializer from #214 (now closed) per feedback from @Nahuel990 — uses a curated acronym set instead of a runtime botocore dependency.

## Changes

- Add `_AWS_ACRONYMS` frozenset with 17 known AWS acronyms (`DB`, `IAM`, `VPC`, `SSL`, `KMS`, etc.)
- Add `_sfn_key_to_api_name()` — reverse of `_api_name_to_sfn_key()`, expands acronyms back to uppercase
- Add `_convert_params_to_api_names()` — recursive dict/list key converter
- Apply conversion in `_dispatch_aws_sdk_query()` before flattening to query params
- CHANGELOG entry

## Design Decisions

- **Excluded `Arn`, `Id`**: These appear as single uppercase chars in wire format (`DBClusterArn`, `KmsKeyId`), not as uppercase runs (`DBClusterARN`)
- **Excluded `Http`, `Https`, `Ec2`**: Mixed-case or digit-containing in wire format, not pure acronym runs
- **Idempotent**: Wire-format names pass through unchanged (single-letter tokens don't match any acronym)

## Tests

- **must-convert**: 16 acronym expansion cases
- **must-not-convert**: 9 names that should be unchanged
- **idempotent**: 5 wire-format names that should pass through
- **round-trip**: 18 `_api_name_to_sfn_key` → `_sfn_key_to_api_name` conversions
- **nested structures**: recursive dict/list conversion
- **integration**: SFN dispatching RDS `CreateDBCluster` with SDK-style param names

All 1125 tests pass (only pre-existing `cbor2` skip).

PR #237, opened 2026-04-10
feat: REST-JSON aws-sdk dispatcher + rdsdata SFN alias + RDS Data API stub fallback

jayjanssen:
## Summary

Add REST-JSON protocol support to the SFN `aws-sdk:*` task dispatcher, enabling Step Functions to call RDS Data API via `arn:aws:states:::aws-sdk:rdsdata:executeStatement`.

This is independent of #235 (acronym mapper) — no shared code or dependencies between them.

### Step Functions (`stepfunctions.py`)
- Register `rdsdata` in `_AWS_SDK_SERVICE_MAP` with `rest-json` protocol and `rds-data` service key
- Add `_REST_JSON_ACTION_PATHS` static map for RDS Data API endpoints (`ExecuteStatement` → `/Execute`, etc.)
- Add `_dispatch_aws_sdk_rest_json()` — path-based routing with JSON body, no botocore dependency
- Route `rest-json` protocol in `_invoke_aws_sdk_integration()`

### RDS Data API (`rds_data.py`)
- Add `_stub_success()` helper for mock environments
- Return stub success (instead of 400) when cluster has no Docker endpoint or DB driver is missing
- Real database errors still propagate as errors (not stubbed)
- Handle cluster `Endpoint` being a string (hostname) vs dict (instance)

### CHANGELOG
- Entry added

## Design Decisions

- **Static path map vs botocore**: RDS Data API paths aren't derivable from action names (`ExecuteStatement` → `/Execute`), so a static map is needed. Using botocore at runtime was rejected upstream per #214 feedback.
- **Stub fallback scope**: Only stubs on missing endpoint or missing driver (`ImportError`). Generic `except Exception` still returns a real error — per code review feedback, masking real DB failures would hide bugs in workflows.

## Tests

- **Happy path**: SFN dispatching `aws-sdk:rdsdata:executeStatement` end-to-end (creates cluster + secret, executes via SFN, verifies stub response)
- **Unknown action**: `aws-sdk:rdsdata:notARealAction` fails deterministically
- **Path mapping**: Unit test verifying all 5 RDS Data API action→path mappings

All 1122 tests pass (only pre-existing `cbor2` skip).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,12 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ### Added
 - **SFN query-protocol acronym mapper** — Step Functions `aws-sdk:*` integrations now correctly convert SDK-style parameter names (e.g. `DbSubnetGroupName`) to wire-format names (`DBSubnetGroupName`) for query-protocol services (RDS, EC2, IAM, STS, etc.). Uses a static acronym mapping — no botocore dependency. Contributed by @jayjanssen.
+- **REST-JSON `aws-sdk:*` dispatcher** — Step Functions can now dispatch `aws-sdk:rdsdata:*` actions (e.g. `executeStatement`, `beginTransaction`) to the RDS Data API service via a new REST-JSON protocol handler. Uses a static action→path mapping — no botocore dependency. Contributed by @jayjanssen.
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


Nahuel990:
Thanks for this @jayjanssen!  I will merge this post migration tomorrow and release it in v1.2.0

Nahuel990 (approved):

PR #238, opened 2026-04-10
feat(lambda): support Node.js ESM (.mjs) handlers

bognari:
## Summary

Lambda handlers using ES modules (`.mjs` files with `export` syntax) could not be loaded because both the warm worker pool and the subprocess executor used `require()` exclusively, which cannot load ESM modules.

## Changes

**`ministack/core/lambda_runtime.py`** (`_NODEJS_WORKER_SCRIPT`):
- Handler loading now tries `require()` first, then falls back to dynamic `import()` when `MODULE_NOT_FOUND` or `ERR_REQUIRE_ESM` is encountered
- Resolves handler from named exports, `default[handlerName]`, or `default` — matching AWS Lambda's ESM resolution

**`ministack/services/lambda_svc.py`** (`_NODE_WRAPPER_SCRIPT`):
- Same `require()` → `import()` fallback
- `process.stdin.on('end')` callback made `async` to support `await import()`

**`tests/test_lambda.py`**:
- Added `test_lambda_nodejs_esm_mjs_handler` with two `.mjs` files:
  - `utils.mjs`: exports a helper function and constant using ESM `export` syntax
  - `index.mjs`: imports from `utils.mjs` via ESM `import` statement
- Verifies ESM import/export between modules works end-to-end

## Motivation

AWS Lambda natively supports `.mjs` handlers. Our Node.js backend uses ESM exclusively (all `.mjs` files). Without this fix, MiniStack's warm worker and subprocess executor fail with `MODULE_NOT_FOUND` for any `.mjs` handler.

## Test plan

- [x] `pytest tests/test_lambda.py -v -k "esm_mjs"` — new test passes
- [x] `pytest tests/test_lambda.py -v` — all 70 tests pass, 1 skipped (docker-only)
- [x] Verified with real-world ESM Lambda handlers using `import`/`export` between modules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ### Added
 - **SFN query-protocol acronym mapper** — Step Functions `aws-sdk:*` integrations now correctly convert SDK-style parameter names (e.g. `DbSubnetGroupName`) to wire-format names (`DBSubnetGroupName`) for query-protocol services (RDS, EC2, IAM, STS, etc.). Uses a static acronym mapping — no botocore dependency. Contributed by @jayjanssen.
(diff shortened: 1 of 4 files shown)


Nahuel990:
Thanks for this @bognari I will merge this post migration tomorrow and release it in v1.2.0

Nahuel990 (approved):

PR #239, opened 2026-04-10
fix(apigateway): populate pathParameters and fix queryStringParameters in v2 proxy event

bognari:
## Summary

API Gateway v2 proxy events were missing `pathParameters` entirely. Lambda handlers that destructured path parameters (e.g. `event.pathParameters.productId`) crashed with `TypeError` because the field was `undefined`. Additionally, multi-value query parameters were passed as Python lists instead of comma-joined strings, and `rawQueryString` stringified lists instead of expanding repeated keys.

## Changes

**`ministack/services/apigateway.py`:**

- Add `_extract_path_params(route_path, request_path)` — extracts `{param}` and `{proxy+}` values from route templates into a dict
- Simplify `_path_matches()` to delegate to `_extract_path_params()`
- Extract path params before `_invoke_lambda_proxy()` call and pass them through
- Add `pathParameters` to the Lambda proxy event (payload format 2.0)
- Fix `queryStringParameters`: comma-join multi-value params per AWS spec (`"tag": "a,b"`)
- Fix `rawQueryString`: expand repeated keys (`tag=a&tag=b`) instead of stringifying lists

**`tests/test_apigw.py`:**

- `test_apigw_path_parameters_in_event` — single `{itemId}` extraction
- `test_apigw_greedy_path_parameters_in_event` — `{proxy+}` extraction
- `test_apigw_query_params_and_headers_in_event` — multi-value query params and custom header forwarding
- `test_apigw_multiple_path_parameters` — two params in one route (`/projects/{projectKey}/items/{itemId}`)
- `test_apigw_no_path_parameters_returns_null` — route without path params returns `null`
- `test_apigw_url_encoded_path_parameter` — URL-encoded values (`%20`) are decoded correctly

## Motivation

Our Node.js backend uses API Gateway v2 routes with path parameters extensively (e.g. `GET /product/{productId}`, `GET /admin/{projectKey}/orders`). Without `pathParameters` in the event, every handler with path parameters fails immediately. The query parameter fixes ensure multi-value params (e.g. `?tag=a&tag=b`) match the AWS API Gateway v2 payload format 2.0 spec.

## Test plan

- [x] `pytest tests/test_apigw.py -v` — all 45 tests pass
- [x] Verified against [AWS API Gateway v2 payload format 2.0 documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)
- [x] Existing path matching tests (`test_apigw_path_param_route`, `test_apigw_greedy_path_param`) still pass
- [x] Edge cases: multiple params, no params, URL-encoded values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 - **API Gateway v1/v2 returns mock response for Node.js Lambdas** — `_invoke_lambda_proxy` in both `apigateway.py` (v2) and `apigateway_v1.py` (v1) only dispatched to the warm worker pool for Python runtimes. Node.js Lambdas received a hardcoded `"Mock response"` instead of being executed, even though the warm worker pool in `lambda_runtime.py` already supports Node.js. Now checks for both `python` and `nodejs` runtimes.
+- **API Gateway v2 missing `pathParameters` in Lambda event** — Routes with path parameters (e.g. `GET /items/{itemId}`) did not extract parameter values into the Lambda proxy event's `pathParameters` field. Handlers received `undefined` instead of `{ "itemId": "..." }`. Now extracts parameters from both `{param}` and `{proxy+}` route templates.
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this fix @bognari !

PR #252, opened 2026-04-11
fix: support list-stored SSM tags in ListTagsForResource

### LINKED ISSUE #249: SSM ListTagsForResource fails with AttributeError: 'list' object has no attribute 'items'
## Description

After upgrading to the latest image (v1.2.0+), `ListTagsForResource` for SSM parameters fails with a 500 error. This blocks Terraform/OpenTofu deployments because the AWS provider v5.x calls `ListTagsForResource` after every SSM `PutParameter`.

## Error

```
ERROR [ministack] Error handling ssm request: 'list' object has no attribute 'items'
  File "/opt/ministack/ministack/app.py", line 537, in app
    status, resp_headers, resp_body = await handler(method, path, headers, body, query_params)
  File "/opt/ministack/ministack/services/ssm.py", line 98, in handle_request
    return handler(data)
  File "/opt/ministack/ministack/services/ssm.py", line 487, in _list_tags_for_resource
    tag_list = [{"Key": k, "Value": v} for k, v in tag_dict.items()]
AttributeError: 'list' object has no attribute 'items'
```

## Root Cause

`_list_tags_for_resource` in `ssm.py:487` expects the stored tags to be a `dict` (`{key: value}`), but the v1.2.0 fix for "SSM `PutParameter` ignores `Tags`" now stores tags as a `list` of `{"Key": k, "Value": v}` objects (the AWS SDK format). The `.items()` call then fails on the list.

## Reproduction

```bash
docker run -d -p 4566:4566 --name ministack nahuelnucera/ministack:latest
# or: ministackorg/ministack:latest

# Create an SSM parameter with tags
aws --endpoint-url=http://localhost:4566 ssm put-parameter \
  --name "/test/param" --value "hello" --type String \
  --tags "Key=env,Value=test" --region us-east-1

# This call fails with 500:
aws --endpoint-url=http://localhost:4566 ssm list-tags-for-resource \
  --resource-type "Parameter" --resource-id "/test/param" --region us-east-1
```

## Expected Fix

In `_list_tags_for_resource`, handle both storage formats:
- If `tag_dict` is already a list of `{"Key": ..., "Value": ...}` objects, return it as-is
- If it's a dict, convert to list format

## Impact

Blocks all Terraform/OpenTofu deployments that create SSM parameters, because the AWS provider calls `ListTagsForResource` after every write operation.

## Environment

- Image: `nahuelnucera/ministack:latest` and `ministackorg/ministack:latest` (both affected)
- Terraform AWS Provider: v5.100.0
- Confirmed on macOS (Docker Desktop) with OpenTofu 1.9.x

saschabuehrle:
Fixes #249

PutParameter stores tags as a list of {Key, Value} entries, while ListTagsForResource was assuming a dict and calling .items(). This update handles both storage shapes without changing existing behavior for dict-backed tags. Added a regression test for the PutParameter(..., Tags=...) path to make sure list-tags keeps working.

Greetings, saschabuehrle

### DIFF
--- a/ministack/services/ssm.py
+++ b/ministack/services/ssm.py
@@ -483,8 +483,16 @@ def _list_tags_for_resource(data):
     else:
         arn = resource_id
 
-    tag_dict = _tags.get(arn, {})
-    tag_list = [{"Key": k, "Value": v} for k, v in tag_dict.items()]
+    tags = _tags.get(arn, {})
+
+    if isinstance(tags, list):
+        tag_list = [
+            {"Key": t.get("Key"), "Value": t.get("Value")}
+            for t in tags
+            if isinstance(t, dict) and "Key" in t and "Value" in t
+        ]
+    else:
+        tag_list = [{"Key": k, "Value": v} for k, v in tags.items()]
 
     return json_response({"TagList": tag_list})
 
--- a/tests/test_ssm.py
+++ b/tests/test_ssm.py
@@ -131,6 +131,21 @@ def test_ssm_tags_v2(ssm):
     assert "team" not in tag_map2
     assert tag_map2["env"] == "staging"
 
+
+def test_ssm_list_tags_handles_put_parameter_tag_format(ssm):
+    ssm.put_parameter(
+        Name="/ssm2/tag/from-put",
+        Value="v",
+        Type="String",
+        Tags=[{"Key": "env", "Value": "test"}, {"Key": "service", "Value": "api"}],
+    )
+
+    resp = ssm.list_tags_for_resource(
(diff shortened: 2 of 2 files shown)


Nahuel990:
Thanks @saschabuehrle! Already fixed in v1.2.2 the issue was PutParameter storing raw SDK tags as a list while AddTagsToResource normalized to dict. Fixed at write time so all tag operations behave the same. But it would be nice to have your test if you want to modify the PR it would be awesome to have it (I didnt add one for this case)

Nahuel990:
Thanks for this PR, I will close it since @bognari added the fix for this, but thank you very much for the contribution!

PR #259, opened 2026-04-11
feat(lambda): symlink layer node_modules for ESM import resolution

bognari:
## Summary

- Node.js ESM `import()` does not use `NODE_PATH` for package resolution — it only walks the ancestor directory tree for `node_modules/`
- When a Lambda Layer provides packages under `nodejs/node_modules/`, ESM `.mjs` handlers fail with `ERR_MODULE_NOT_FOUND` because the layer is extracted to a separate temp directory outside the handler's ancestor tree
- **The existing `NODE_PATH` setup (v1.2.0, PR #236) works for CJS `require()` but not for ESM `import()`** — `Module._initPaths()` also does not propagate into the ESM resolver

## Fix

After extracting layers, symlink each package from the layer's `nodejs/node_modules/` into the handler's `code/node_modules/` directory. This makes packages visible to both CJS `require()` and ESM `import()` without changing the existing `NODE_PATH` mechanism.

Applied in:
- **Warm worker pool** (`lambda_runtime.py` — `Worker._spawn()`)
- **One-shot subprocess** (`lambda_svc.py` — `_execute_function_local()`)

Both locations are gated to Node.js runtimes only.

## Reproduction

```python
# ESM handler that imports from a Layer — fails without this fix
handler_code = (
    "import helper from 'esmhelper';\n"
    "export const handler = async (event) => {\n"
    "  return { value: helper.LAYER_VALUE, esm: true };\n"
    "};\n"
)
```

Error before fix:
```
Cannot find package 'esmhelper' imported from /tmp/ministack-lambda-.../code/index.mjs
```

## Test plan

- [x] Added `test_lambda_warm_worker_nodejs_esm_uses_layer` — ESM handler + Layer with native `import`
- [x] Full Lambda test suite passes (80/80)
- [x] Existing CJS + Layer test (`test_lambda_warm_worker_nodejs_uses_layer`) still passes
- [x] Python + Layer test (`test_lambda_warm_worker_uses_layer`) still passes — symlinks correctly gated to Node.js only

### DIFF
--- a/ministack/core/lambda_runtime.py
+++ b/ministack/core/lambda_runtime.py
@@ -385,6 +385,22 @@ class Worker:
                 logger.error("Unexpected error resolving layer %s: %s", layer_arn, e)
                 raise RuntimeError(f"Failed to resolve layer {layer_arn}") from e
 
+        # Symlink layer node_modules packages into the code directory so that
+        # Node.js ESM import() can resolve them via ancestor-tree lookup.
+        # ESM does not use NODE_PATH, so packages must be physically reachable
+        # from the handler file's directory tree.
+        if layers_dirs and runtime.startswith("nodejs"):
+            code_nm = os.path.join(code_dir, "node_modules")
+            os.makedirs(code_nm, exist_ok=True)
+            for ld in layers_dirs:
+                layer_nm = os.path.join(ld, "nodejs", "node_modules")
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this @bognari ! I'll add it to next release

PR #292, opened 2026-04-13
fix(rds): parameter group persistence with ApplyMethod, Source filter, and engine version family fix

jayjanssen:
## Summary

Three related fixes to RDS parameter group handling:

### 1. Parameter persistence with ApplyMethod

`ModifyDBParameterGroup` and `ModifyDBClusterParameterGroup` now store `ApplyMethod` alongside parameter values (as `{ParameterValue, ApplyMethod}` dicts). Previously only the raw value string was stored, losing `ApplyMethod`.

### 2. Source filter support

`DescribeDBParameters` and `DescribeDBClusterParameters` now support the `Source` filter parameter (`user`, `engine-default`). Custom parameters are correctly identified as `Source=user`, and custom parameters not in the engine defaults are included in the response.

### 3. DescribeDBEngineVersions family fix

`DBParameterGroupFamily` was constructed as `{engine}{family}` but the `family` string already includes the engine prefix (e.g. `aurora-mysql8.0`), producing garbled values like `aurora-mysqlaurora-mysql8.0`. Fixed to use just `{family}`.

## Tests

- `test_rds_modify_and_describe_db_parameters` — modify with ApplyMethod, describe with Source=user filter
- `test_rds_modify_and_describe_cluster_parameters` — same for cluster parameter groups
- `test_rds_describe_engine_versions_family` — verify no double-prefix

## Checklist

- [x] Tests added and passing
- [x] CHANGELOG entry added
- [x] Linting passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Added
+- **RDS parameter group persistence** — `ModifyDBParameterGroup` and `ModifyDBClusterParameterGroup` now store `ApplyMethod` alongside parameter values. `DescribeDBParameters` and `DescribeDBClusterParameters` return stored parameters with `Source` filter support (`user`, `engine-default`). Custom parameters beyond engine defaults are included.
+
+### Fixed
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this @jayjanssen ! Great catch!

PR #296, opened 2026-04-13
fix(sfn): prefix aws-sdk error codes with service name for Catch matching

jayjanssen:
## Problem

Real AWS Step Functions prefixes SDK error codes with the service name when surfacing them to the ASL execution engine. For example, when `aws-sdk:secretsmanager:createSecret` fails with `ResourceExistsException`, the error is reported as `SecretsManager.ResourceExistsException`.

MiniStack was passing through raw error codes (e.g., `ResourceExistsException`), which meant ASL `Catch` blocks matching on the prefixed form silently missed:

```json
"Catch": [{
  "ErrorEquals": ["SecretsManager.ResourceExistsException"],
  "Next": "HandleExisting"
}]
```

This caused the execution to fail instead of routing to the catch handler.

## Solution

Added a static `_AWS_SDK_ERROR_PREFIX` map and `_prefix_sdk_error()` helper that converts lowercase service names from aws-sdk ARNs to PascalCase prefixes matching real AWS behavior. Applied in all three aws-sdk dispatchers:

- **JSON protocol** (SecretsManager, DynamoDB, etc.)
- **Query protocol** (RDS, EC2, SQS, etc.)
- **REST-JSON protocol** (RDS Data API)

If the error code already contains a dot (e.g., `States.Runtime`), it is returned as-is.

No new dependencies — the map is static with entries for all services in `_AWS_SDK_SERVICE_MAP`.

## Tests

- `test_sfn_aws_sdk_error_prefix_catch`: Pre-creates a secret, then runs an SFN that tries to create the same secret. Verifies the `Catch` block matching `SecretsManager.ResourceExistsException` routes to the fallback state.
- `test_sfn_aws_sdk_error_prefix_in_failed_execution`: Runs an SFN that describes a nonexistent RDS cluster without a Catch. Verifies the failure error starts with `Rds.`.

All 78 SFN tests pass.


Nahuel990 (approved):

Nahuel990:
Good fix @jayjanssen this is going in 1.2.10

PR #302, opened 2026-04-14
Run all lambdas in a container with RIE support if LAMBDA_EXECUTOR is set to docker

fzonneveld:
Currently a mix of docker runtime images is used, which has the following disadvantages.
* Using wrappers to start the code
* Does not mimic close enough to what AWS does
* Incompatible libraries, libs for slim are not the same as for AWS images
* Not a unified way of running AWS lambdas
Single advantage is speed, a slim image for python much smaller.

In this branch unified the lambda handling by always using the AWS docker images with RIE (Runtime Interface Emulator).
Named volumes (sub paths) are mounted for the code (/var/task) and for layers (/opt).
There should be not difference in running a python or a nodejs container.

### DIFF
--- a/ministack/services/lambda_svc.py
+++ b/ministack/services/lambda_svc.py
@@ -29,22 +29,34 @@ import base64
 import copy
 import hashlib
 import importlib
+import io
 import json
 import logging
 import os
+import pathlib
+import random
 import re
 import subprocess
 import tempfile
 import threading
 import time
+import typing
 import zipfile
 from datetime import datetime, timezone
-from typing import Any
 from urllib.parse import unquote
 
-from ministack.core.persistence import load_state, PERSIST_STATE
+from ministack.core.persistence import load_state
 from ministack.core.responses import AccountScopedDict, get_account_id, error_response_json, json_response, new_uuid
 from ministack.core.lambda_runtime import get_or_create_worker, invalidate_worker
+
+if typing.TYPE_CHECKING:
+    from mypy_boto3_lambda.type_defs import FunctionConfigurationResponseTypeDef, FunctionConfigurationTypeDef
+    from mypy_boto3_lambda.literals import RuntimeType
+    from docker.models.containers import Container
+
+    class LambdaConfigType(FunctionConfigurationResponseTypeDef):
+        ImageUri: typing.Optional[str]
+
 
 logger = logging.getLogger("lambda")
 
@@ -53,8 +65,11 @@ LAMBDA_EXECUTOR = os.environ.get("LAMBDA_EXECUTOR", "local").lower()
 LAMBDA_DOCKER_VOLUME_MOUNT = os.environ.get("LAMBDA_REMOTE_DOCKER_VOLUME_MOUNT", "")
 LAMBDA_DOCKER_NETWORK = os.environ.get("LAMBDA_DOCKER_NETWORK", "")
 
+if typing.TYPE_CHECKING:
+    from docker import DockerClient
+
 try:
-    docker_lib: Any = importlib.import_module("docker")
+    docker_lib: "DockerClient" = importlib.import_module("docker")
     _docker_available = True
 except ImportError:
     docker_lib = None
@@ -67,12 +82,32 @@ _function_urls = AccountScopedDict()  # function_name -> FunctionUrlConfig dict
 _poller_started = False
 _poller_lock = threading.Lock()
 
+_containers_lock = threading.Lock()
+
+
+class ContainerScopedDict(AccountScopedDict):
+    def clear(self):
+        """Stop running ccontainers when clear"""
+        with _containers_lock:
+            for containers in self._data.values():
+                for container in containers:
+                    try:
+                        container.stop()
+                        container.remove()
+                    except Exception:
+                        pass
+        super().clear()
+
+
+_containers = ContainerScopedDict()
 
 # ── Persistence ────────────────────────────────────────────
+
 
 def get_state():
     """Return JSON-serializable state. code_zip bytes are base64-encoded."""
     from ministack.core.responses import AccountScopedDict
+
     funcs = AccountScopedDict()
     # Iterate _data directly to capture ALL accounts, not just current request context
     for scoped_key, func in _functions._data.items():
@@ -94,6 +129,7 @@ def get_state():
 def restore_state(data):
     if data:
         from ministack.core.responses import AccountScopedDict
+
         funcs = data.get("functions", {})
         if isinstance(funcs, AccountScopedDict):
             for scoped_key, func in funcs._data.items():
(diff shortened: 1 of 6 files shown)


Nahuel990:
@fzonneveld there are several breaking changes from this PR, could you rebase it. I see some intentions are good but you are implementing a few things that could break Go/rust binaries

fzonneveld:
> @fzonneveld there are several breaking changes from this PR, could you rebase it. I see some intentions are good but you are implementing a few things that could break Go/rust binaries

Fast reply, thanks..
I will rebase.
I do not see where the breaking changes are that could break Go/rust binaries.  The binaries would just use the provided images right, interface for RIE should be the same I would hope.

fzonneveld:
Aha I do see how the code is running locally with a go/rust we extract the  "bootstrap" and run this in a server. Implementing the RIE. This is very complex. I would advise to run binaries always in dockers, this removes the problem of architecture and all other bugs in creating own API for Go/rust.
Go/rust is already supported in _execute_function_docker_rie.
Simple rule, if want to run lambda, need docker setup, always run in docker.

Even in the function _execute_function_docker_rie in this MR, there is a race condition. The _containers should be updated/accessed with a threading lock, because the function is called from a thread. 

I will rebase and include the Go/Rust binary execution locally.

jgrumboe:
Just a side question: would the use of `public.ecr.aws/lambda/python.*` mean that `boto3` would already be included in the lambda runtime (as it is in AWS)?
That would be a great DX improvement!

Yesterday I've read this article https://repost.aws/articles/AR-WRCanAJScabqnY2OQ5H7A/local-lambda-testing-with-ministack-sam-cli-and-finch, and the author mentions the absence of boto3 as one known limitation of ministack.

fzonneveld:
That is correct in the python lambda container boto3 is included, and it does not need to be included in the packages for the lambda, although in general people still advise to include the latest version of boto3 in the lambda package.

I rebased.

I also added the thead safety for the _containers variable and some shuffling of containers to pick for running, please have a look.

Adding some description howto to run  lambdas in a docker.

fzonneveld:
Added some small docs howto use in docker-compose, and rebased.

ministack/services/lambda_svc.py
```diff
@@ -1001,310 +859,256 @@ async def _invoke(name: str, event: dict, headers: dict, path_qualifier: str | N
...
+    "nodejs14.x": "public.ecr.aws/lambda/nodejs:14",
+    "nodejs16.x": "public.ecr.aws/lambda/nodejs:16",
+    "nodejs18.x": "public.ecr.aws/lambda/nodejs:18",
+    "nodejs20.x": "public.ecr.aws/lambda/nodejs:20",
+    "nodejs22.x": "public.ecr.aws/lambda/nodejs:22",
+    "nodejs24.x": "public.ecr.aws/lambda/nodejs:24",
     "provided.al2023": "public.ecr.aws/lambda/provided:al2023",
     "provided.al2": "public.ecr.aws/lambda/provided:al2",
-    "provided": "public.ecr.aws/lambda/provided:al2023",
+    "provided": "public.ecr.aws/lambda/provided:latest",
```
fzonneveld: Non intended change, should be al2023.

ministack/services/lambda_svc.py
```diff
@@ -1001,310 +859,256 @@ async def _invoke(name: str, event: dict, headers: dict, path_qualifier: str | N
...
+    config = func["config"]
+    func_name = config["FunctionName"]
+    version = config["Version"]
+    function_code = _get_function_code(func)
+    with _containers_lock:
+        _containers.setdefault(function_code, [])
+        # Gert a local copy
+        containers = _containers[function_code].copy()
+        # Shuffle order to divide the load
+        random.shuffle(containers)
```
fzonneveld: Shuffle does not need to be in the lock.

ministack/services/lambda_svc.py
```diff
@@ -1423,26 +1242,21 @@ def _execute_function_image(func: dict, event: dict) -> dict:
...
         import time as _time
         import urllib.request
+
         for _attempt in range(int(timeout * 2) + 10):
             _time.sleep(0.5)
             container.reload()
             if container.status != "running":
                 break
             try:
-                if LAMBDA_DOCKER_NETWORK:
```
fzonneveld, line 1447: This is not an intended change.

fzonneveld (commented):
Some small comments.

ministack/services/lambda_svc.py
fzonneveld: Outside the lock

ministack/services/lambda_svc.py
fzonneveld: Put back.

Nahuel990:
@fzonneveld Awesome work!

Could this PR instead of implementing from scratch extend the existing RIE logic to handle all (currently it has PackageType= Images). One shared code path for all package types. I tried with a Go binary and it goes through the same RIE pattern fine.

and I'm still having 2 things
After running a few times containers pile up with no way to clean them, not even with the reset endpoint (docker system prune was the only option which is not ideal in case we have persist state it will wipe out all) 
And the other thing in docs you mentioned privileged: true... I checked with just the socket mount and it works fine, I dont think there's need to give access the docker to the host 

Also could you move the wrapper scripts to core? We keep one file per service in service folder

And I'm getting into an error with multitenancy that I can't understand yet but I will post it here once I get the full trace

fzonneveld:
Will try to combine the logic for runtime, should work the same. Will try to use some typed dicts for the config, but this will result in including mypy boto3 stuff in the distribution.

For the reset interface will add the removal of running containers registered in _containers.

Privileged is needed for accessing /var/run/docker.sock in general I thought, but I think this can also be done with normal permission settings. Will remove it.

I will move the wrapper scripts to core.

Nahuel990:
sorry what are the uses you give to boto3? I can see the TypeCheckcing validation in the PR... is there another use?

fzonneveld:
I try try to use the typing from mypy boto3 to get a more clear view of
config in _functions a little bit more clear.

Should not be needed in release, only in dev group.

Thanks,
Frank

Op do 16 apr 2026, 08:43 schreef Nahuel Nucera ***@***.***>:

fzonneveld:
Merged package type Zip and Image into the same now.
Added a merged config.
Implemented the _reset part to remove created lambda containers.
Wrapper scripts moved to core.

fzonneveld:
Need to rebase again, lots of progress in the repo 👍

Nahuel990:
@fzonneveld Thanks for this work! I will merge this and release it as part of 1.2.20

Nahuel990 (approved):

fzonneveld:
I do not see the correct code in 1.2.20
My test code (copy file to s3 to invoke a long running lambda) is failing now with 500 response for docker running.
The invokde_docker_rte does not use (sub) volume mounting points, kills containers afterwards.
Merging of execute image into _execute_function_docekr is not there anymore.
"warm containers" is never a single warm container, must be a list.
code_sha is not always set, not for containers with package type image.

Actually I am missing all the code, very strange.

Nahuel990:
Hey @fzonneveld you are right. Post-merge fix PRs rebased against a prior #302 branch and removed part of your work: _execute_function_image got split back out, the pool went back to single-container-per-key, and CodeSha256 was dropped for Image type. We recommend adding tests to avoid this as we didn’t catch it at review.

Just tagged v1.3.3, which restores your idea, unified _execute_function_docker for Zip + Image, list-per-key warm pool, thread-safe, and reset() kills every pooled container. On top of that: LAMBDA_STRICT=1 for your “docker or nothing” rule, async retry + DLQ, Lambda → CloudWatch Logs, and State: Pending → Active transitions.

Your “copy file to S3 → long-running Lambda” case should work end-to-end on ministackorg/ministack:1.3.3. If you get a minute, please try it you’re the closest reviewer we have for this code path

PR #309, opened 2026-04-14
Fix/performance

Nahuel990:
This change is quite big. I discovered we don't need pip in the image and if we do lazy loading we can reduce 1.3 second start to 0.4, also memory idle could go down to 22Mb 

I'm also bundling scheduler service because I'm an idiot and did the change in the same branch

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -40,52 +40,49 @@ _S3_VHOST_EXCLUDE_RE = re.compile(r"\.(execute-api|alb|emr|efs|elasticache|s3-co
 from ministack.core.persistence import PERSIST_STATE, load_state, save_all
 from ministack.core.responses import set_request_account_id
 from ministack.core.router import detect_service, extract_access_key_id, extract_account_id, extract_region
-from ministack.services import (
-    acm,
-    alb,
-    apigateway,
-    autoscaling,
-    apigateway_v1,
-    appconfig,
-    appsync,
-    athena,
-    cloudformation,
-    cloudwatch,
-    cloudwatch_logs,
-    cognito,
-    dynamodb,
-    ec2,
-    ecr,
-    ecs,
-    efs,
-    elasticache,
-    emr,
-    eventbridge,
-    firehose,
-    glue,
-    kinesis,
-    kms,
-    lambda_svc,
-    rds,
-    rds_data,
-    route53,
-    s3,
-    secretsmanager,
-    ses,
-    ses_v2,
-    sns,
-    sqs,
-    ssm,
-    stepfunctions,
-    waf,
-    cloudfront,
-    servicediscovery,
-    s3files,
-    codebuild,
-    transfer,
-)
-from ministack.services import iam_sts
-from ministack.services.iam_sts import handle_iam_request, handle_sts_request
+
+# ---------------------------------------------------------------------------
+# Lazy service loader — modules are imported on first request, not at startup.
+# This saves ~20 MB of idle RAM and speeds up boot.
+# ---------------------------------------------------------------------------
+_loaded_modules: dict = {}
+
+
+class _ErrorModule:
+    """Stub returned when a service module fails to import."""
+    def __init__(self, name: str, error: str):
+        self._name = name
+        self._error = error
+
+    async def handle_request(self, method, path, headers, body, query_params):
+        return 500, {"Content-Type": "application/json"}, \
+            json.dumps({"__type": "ServiceUnavailable",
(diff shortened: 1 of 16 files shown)


jgrumboe:
Ok, cool.
I can do the review tomorrow morning, hope that's fine for you.

Nahuel990:
Yeah no worries! I'm mostly worried about the app.py file where it has the changes

ministack/app.py
```diff
@@ -95,49 +74,50 @@
...
-    "appconfig": appconfig.handle_request,
-    "appconfigdata": appconfig.handle_request,
+    "s3": _lazy_handler("s3"),
+    "sqs": _lazy_handler("sqs"),
+    "cloudformation": _lazy_handler("cloudformation"),
+    "sns": _lazy_handler("sns"),
+    "dynamodb": _lazy_handler("dynamodb"),
+    "lambda": _lazy_handler("lambda_svc"),
+    "iam": _lazy_handler("iam_sts", "handle_iam_request"),
+    "sts": _lazy_handler("iam_sts", "handle_sts_request"),
```
jgrumboe: Should we split `iam_sts.py` into two files, like every other service, too?
Both services have the standard handle_request() entry point afterward (consistent with all other services).
Simplifies _lazy_handler() input arguments as well.

ministack/app.py
```diff
@@ -620,47 +599,31 @@ async def _handle_lifespan(scope, receive, send):
...
-                    "ses_v2": ses_v2.get_state,
-                    "servicediscovery": servicediscovery.get_state,
-                    "s3files": s3files.get_state,
-                    "appconfig": appconfig.get_state,
-                })
+                # Only save state for modules that were actually loaded
+                _state_map = {
+                    "apigateway": "apigateway", "apigateway_v1": "apigateway_v1",
+                    "sqs": "sqs", "sns": "sns", "ssm": "ssm",
+                    "secretsmanager": "secretsmanager", "iam": "iam_sts",
```
jgrumboe: Splitting `iam_sts.py` would also simplify the _state_map from indirection to direct mappings.

ministack/app.py
```diff
@@ -40,51 +40,30 @@
...
+# ---------------------------------------------------------------------------
+_loaded_modules: dict = {}
+
+
+def _get_module(name: str):
+    """Import and cache a service module by short name (e.g. 's3', 'lambda_svc')."""
+    mod = _loaded_modules.get(name)
+    if mod is None:
+        mod = __import__(f"ministack.services.{name}", fromlist=["handle_request"])
+        _loaded_modules[name] = mod
```
jgrumboe:
```suggestion
        try:
            mod = __import__(f"ministack.services.{name}", fromlist=["handle_request"])
        except ModuleNotFoundError as e:
            logger.warning("Service module not found: %s - %s", name, e)
            mod = _ErrorModule(name)  # Return stub to prevent crashes
        _loaded_modules[name] = mod
```

I think we should add error handling here, just to be safe.

jgrumboe (commented):
Nice idea with the lazy loading! Small ideas about it inline.

Eventscheduler seems to be fine, I had just a quick look and trust the tests here.

Nahuel990:
Awesome! thanks for this @jgrumboe will look into this this afternoon

Nahuel990:
The review was spot on @jgrumboe I made the changes thanks!... Before merging could you take an extra look just in case?

ministack/app.py
```diff
@@ -622,48 +627,31 @@ async def _handle_lifespan(scope, receive, send):
...
-                    "transfer": transfer.get_state,
-                    "acm": acm.get_state,
-                    "firehose": firehose.get_state,
-                    "ses": ses.get_state,
-                    "ses_v2": ses_v2.get_state,
-                    "servicediscovery": servicediscovery.get_state,
-                    "s3files": s3files.get_state,
-                    "appconfig": appconfig.get_state,
-                })
+                # Only save state for modules that were actually loaded
```
jgrumboe, line 658: @Nahuel990  Just thinking: what happens with a state that's from a module that was loaded in a previous ministack run and stored? Do we leave the existing state of modules not loaded at this run untouched?

Nahuel990: there's a json per service... reset endpoint clears everyhing but if you dont run rjeset the module will have its s3.JSON or sqs.JSON from the previous run...now on load we create all the 40+ JSON

jgrumboe (approved):
LGTM 🚀 

Great improvement!
(nitpicking: next time, don't mix with other features 😉 )

Nahuel990:
I know 😞 thanks for the review, will merge at 11am UTC

PR #311, opened 2026-04-14
fix: suppress noisy Lambda BrokenPipeError and RDS Data API import warning spam

jayjanssen:
## Summary

Two log noise fixes for cleaner output during testing:

### Lambda Runtime API — BrokenPipeError suppression

When MiniStack runs a Go Lambda binary via the native `provided` runtime executor, it spins up a minimal HTTP server implementing the Lambda Runtime API. The Go binary calls `GET /invocation/next` to receive the event, processes it, then posts the response and disconnects. Python's `socketserver.TCPServer` sees the closed socket and prints a full `BrokenPipeError` traceback to stderr. This is benign — the Lambda executed successfully — but pollutes logs.

Fix: subclass `TCPServer` with a `handle_error` override that suppresses only connection-related errors (`BrokenPipeError`, `ConnectionResetError`, `ConnectionAbortedError`). Other unexpected errors still surface normally.

### RDS Data API — pymysql import warning

When `pymysql` is not installed (the default in the official Docker image), `_execute_statement` falls back to SQL-aware stubs. Each call logs `WARNING: DB driver not available, using stub: ...`. During an acceptance test run this produces 300+ identical warnings.

Fix: log the warning once per process using a function-attribute guard (`_execute_statement._import_warned`).

## Changes

- `lambda_svc.py`: `_QuietTCPServer` subclass with filtered `handle_error`
- `rds_data.py`: one-shot import warning guard
- `CHANGELOG.md`: entry added

No new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **Lambda Runtime API noise** — suppressed `BrokenPipeError` tracebacks from Lambda binaries disconnecting after reading the event. This is benign and expected behavior during native `provided` runtime execution.
+- **RDS Data API warning spam** — the `pymysql` import warning is now logged once per process instead of on every `ExecuteStatement` call.
+
+---
+
 ## [1.2.11] — 2026-04-14
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
Thanks @jayjanssen !

PR #315, opened 2026-04-14
fix(sfn): handle escape sequences in States.Format template strings

jayjanssen:
## Summary

`States.Format` now correctly processes `\'`, `\{`, `\}`, and `\\\\` escape sequences in template strings, matching AWS behavior.

### What was broken

The intrinsic argument parser used a naive `s.index("'", pos + 1)` to find the closing quote of string literals. When a `States.Format` template contained escaped quotes (e.g. `'it\\'s {}'`), the parser truncated the string at the first `\'`, producing garbled SQL and other template output.

Additionally, `States.Format` did no escape processing at all — backslash sequences like `\'` were passed through literally.

### What this fixes

1. **String literal scanning** — `_parse_intrinsic_args()` now handles `\'` escapes when scanning for the closing quote.
2. **Template-only escape processing** — `States.Format` processes escape sequences (`\x` → `x`) in template segments only. Interpolated argument values pass through verbatim, so backslashes in user data are preserved.

### Test coverage

New test `test_sfn_intrinsic_format_escapes` covers:
- Escaped single quotes: `it\\'s {}` → `it's fine`
- Escaped braces: `\\{literal\\}` → `{literal}`  
- Escaped backslash: `C:\\\\\tmp` → `C:\tmp`
- Interpolated values with backslashes are preserved verbatim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **SFN States.Format escape handling** — `States.Format` now correctly processes `\'`, `\{`, `\}`, and `\` escape sequences in template strings, matching AWS behavior. Escaped quotes no longer truncate the template during intrinsic argument parsing. Interpolated values are preserved verbatim (backslashes in arguments are not interpreted as escapes).
+
+---
+
 ## [1.2.12] — 2026-04-14
 
 ### Added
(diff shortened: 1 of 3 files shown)


Nahuel990:
@jayjanssen your test is doing the opposite of what you implemented here, also the Changelog entry is at line 800. Could you review those?

jayjanssen:
Thanks @Nahuel990 — both fixed:

1. **Test assertion**: The `preserved` case had the wrong expected value — Python string escaping meant it was comparing against a single-backslash string instead of the correct `C:\\tmp\\file`. Fixed.
2. **CHANGELOG**: Moved from line 874 (inside v1.1.13) to the top under `[Unreleased]`.

Force-pushed with the fixes.

Nahuel990 (approved):

PR #316, opened 2026-04-14
feat(rds): real database connectivity via Docker networking and secret-aware auth

jayjanssen:
## Summary

Enables end-to-end real MySQL/MariaDB execution through the RDS Data API when MiniStack runs in Docker. Previously, the Data API fell back to in-memory stubs because (a) `pymysql` wasn't in the image, and (b) the RDS endpoint used `localhost` which doesn't reach sibling Docker containers.

### Changes

**`pymysql` bundled in Docker image** (44 KB, pure Python)
- Already listed in `pyproject.toml[full]` but wasn't installed in the image.

**Docker network detection** (`rds.py`)
- `_get_ministack_network()` detects MiniStack's own Docker network — same pattern as `ecs.py` lines 796–806.
- RDS containers are attached to this network and assigned internal IPs.
- `_wait_for_port()` blocks until the DB container accepts TCP connections.
- The **public endpoint stays `localhost:host_port`** for host-mode access. The internal IP is stored separately (`_internal_address`/`_internal_port`) and used only by the Data API.

**Secret-aware authentication** (`rds_data.py`)
- `_get_secret_credentials()` extracts both username and password from Secrets Manager.
- `_connect()` maps the master user to MySQL `root` for admin operations (CREATE USER, GRANT, etc.). Non-master secrets use their actual username for user-level operations like password rotation.

**Cluster password management** (`rds.py`)
- `CreateDBCluster` stores `_MasterUserPassword`.
- `CreateDBInstance` inherits credentials from the parent cluster.
- `ModifyDBCluster` propagates password changes to the real MySQL container via `ALTER USER` (engine-gated to MySQL/MariaDB only).
- `MYSQL_ROOT_HOST=%` added to container environment to enable remote root access.

### Test coverage

- `test_rds_cluster_password_stored` — verifies cluster stores master password
- `test_rds_modify_cluster_password` — verifies ModifyDBCluster updates stored password
- `test_rds_data_secret_credentials_parsing` — verifies username+password extraction
- `test_rds_data_secret_credentials_no_username` — verifies password-only secret handling

Real MySQL integration tests require Docker-in-Docker and are validated locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Added
+- **RDS real MySQL/MariaDB connectivity** — `pymysql` (44 KB, pure Python) is now bundled in the Docker image. When MiniStack runs inside Docker, RDS containers are attached to MiniStack's Docker network with internal IP endpoints for sibling-container connectivity. The public `localhost` endpoint remains unchanged for host-mode access. The Data API authenticates using credentials from Secrets Manager, mapping the master user to MySQL `root` for admin operations. `CreateDBCluster` stores the master password; `CreateDBInstance` inherits credentials from parent clusters; `ModifyDBCluster` propagates password changes to the real MySQL container via `ALTER USER`.
+
+---
+
 ## [1.2.12] — 2026-04-14
 
 ### Added
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ RUN pip install --no-cache-dir --no-compile \
         "docker>=7.0.0" \
         "pyyaml>=6.0" \
         "cryptography>=41.0" \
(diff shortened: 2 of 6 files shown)


Nahuel990:
I finished pulling 313 building and I just saw you closed it in favor of this one, this has the missing fix for 313, I see the test failing because they are importing rds_svc._clusters directly the server runs a separate process to ensure we keep multitenancy feature, you need to set something like  set_request_account_id("000000000000") for them to pass

jayjanssen:
Thanks @Nahuel990 — good catch! Fixed both issues:

1. **Tests now call `set_request_account_id("test")`** before accessing `_clusters` and `_secrets` directly, matching the boto3 fixture's access key.
2. Also fixed the `rds_data` tests that access `secretsmanager._secrets` with the same pattern.

Force-pushed with the fixes.

Nahuel990 (approved):

PR #322, opened 2026-04-14
fix(cognito): default AnalyticsConfiguration to None instead of empty dict

### LINKED ISSUE #320: Cognito: AnalyticsConfiguration returned when not set on user pool client
### Affected Service

Cognito

### What happened?

In `_create_user_pool_client` (`ministack/services/cognito.py`, line 611), `AnalyticsConfiguration` defaults to `{}` (empty dict). The response filter `{k: v for k, v in client.items() if v is not None}` keeps it because `{}` is not `None`.

When Terraform reads back the user pool client, it sees an `analytics_configuration` block it never sent, causing:

```
Provider produced inconsistent result: analytics_configuration block count changed from 0 to 1
```

The same issue applies in `_update_user_pool_client` and `_describe_user_pool_client` since they share the same response filter and stored client dict.

### Reproduction steps

```hcl
resource "aws_cognito_user_pool" "test" {
  name = "test-pool"
}

resource "aws_cognito_user_pool_client" "test" {
  name         = "test-client"
  user_pool_id = aws_cognito_user_pool.test.id
  # Note: no analytics_configuration set
}
```

### Expected behavior

When `AnalyticsConfiguration` is not provided in the request, it should not appear in the response. The response should omit the key entirely.

### Fix

Default `AnalyticsConfiguration` to `None` instead of `{}` on line 611:

```python
"AnalyticsConfiguration": data.get("AnalyticsConfiguration"),
```

### MiniStack version

latest

### SDK / tool used

Terraform AWS Provider ~> 5.x

alexanderkrum-next:
## Summary

- Default `AnalyticsConfiguration` to `None` instead of `{}` in `_create_user_pool_client` so the existing `v is not None` response filter strips it when unset.
- Fixes Terraform error: `Provider produced inconsistent result: analytics_configuration block count changed from 0 to 1`.

Closes #320

## Test plan

- [x] Create a Cognito user pool client via Terraform **without** `analytics_configuration` — verify no inconsistent result error
- [x] Create a Cognito user pool client **with** `analytics_configuration` — verify it is returned correctly in the response

### DIFF
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -608,7 +608,7 @@ def _create_user_pool_client(data):
         "AllowedOAuthFlows": data.get("AllowedOAuthFlows", []),
         "AllowedOAuthScopes": data.get("AllowedOAuthScopes", []),
         "AllowedOAuthFlowsUserPoolClient": data.get("AllowedOAuthFlowsUserPoolClient", False),
-        "AnalyticsConfiguration": data.get("AnalyticsConfiguration", {}),
+        "AnalyticsConfiguration": data.get("AnalyticsConfiguration"),
         "PreventUserExistenceErrors": data.get("PreventUserExistenceErrors", "ENABLED"),
         "EnableTokenRevocation": data.get("EnableTokenRevocation", True),
         "EnablePropagateAdditionalUserContextData": data.get("EnablePropagateAdditionalUserContextData", False),


Nahuel990 (approved):

Nahuel990:
Awesome work @alexanderkrum-next thanks!

PR #336, opened 2026-04-15
fix(lambda): override AWS_ENDPOINT_URL after function env vars

jayjanssen:
## Problem

When a Lambda function has `AWS_ENDPOINT_URL` in its environment variables (e.g., set by a Terraform provider during `CreateFunction`), the function's env var overwrites MiniStack's internal endpoint URL. Lambda binaries run inside the MiniStack process/container and must call back to the internal port (4566), not the host-mapped port which is unreachable from inside the container.

This causes all Lambda SDK callbacks (RDS Data API, SecretsManager, etc.) to fail with "connection refused" whenever the host port differs from the container port (e.g., `docker compose` with `4568:4566`).

## Root Cause

In all four executor paths, `proc_env.update(env_vars)` was called **after** setting `AWS_ENDPOINT_URL`, allowing the function-level value to overwrite it:

```python
proc_env["AWS_ENDPOINT_URL"] = endpoint  # internal: http://localhost:4566
proc_env.update(env_vars)                 # OVERWRITES with http://localhost:4568
```

## Fix

Apply function env vars **first**, then restore `AWS_ENDPOINT_URL` from the process-level environment. Affects:
- `_execute_function_provided()` in `lambda_svc.py` (Go/Rust binaries)
- `_execute_function_docker()` in `lambda_svc.py`
- `_execute_function_image()` in `lambda_svc.py`
- `Worker.__init__()` in `lambda_runtime.py` (warm workers for Python/Node.js)

## Test

`test_lambda_endpoint_url_not_overridden_by_function_env` — creates a Lambda with a bogus `AWS_ENDPOINT_URL` env var, invokes it, and asserts the Lambda sees MiniStack's real endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **Lambda endpoint URL override** — function-level `AWS_ENDPOINT_URL` environment variables no longer override MiniStack's internal endpoint. When MiniStack runs in Docker with a host-port that differs from the container port (e.g., `4568:4566`), Lambda functions would receive the host-mapped URL which is unreachable from inside the container, causing SDK callbacks to fail with "connection refused". Fix applies to all executor paths: provided runtime, Docker mode, image mode, and warm workers. Contributed by @jayjanssen
+
+---
+
 ## [1.2.15] — 2026-04-15
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


Nahuel990:
@jayjanssen thanks for this I am merging this for now but we are about to change in favour of RIE to match AWS behaviour there is an openPR here https://github.com/ministackorg/ministack/pull/302

jayjanssen:
@Nahuel990 ah, good to hear.  Thanks for the news and for the merge.

Nahuel990 (approved):

PR #337, opened 2026-04-15
fix(sfn): don't scale activity/callback timeouts with SFN_WAIT_SCALE

jayjanssen:
## Problem

`SFN_WAIT_SCALE=0` causes `States.Timeout` on activity tasks and `waitForTaskToken` callbacks. Any callback or activity that takes longer than 10ms fails.

## Root Cause

The v1.2.12 release commit (#318) applied `_scaled_timeout()` to activity task and callback waits:

```python
timeout = _scaled_timeout(99999)        # with scale=0 → 0.01s
if not evt.wait(timeout=timeout): ...   # times out immediately
```

These are **functional timeouts** that must wait for real work to complete (Lambda execution, external callbacks). They are not artificial delays like Wait states or retry intervals.

## Fix

Revert to unscaled timeouts for the two affected call sites:
- `_invoke_activity()` — activity task wait
- `_invoke_with_callback()` — `waitForTaskToken` pattern

Remove the now-unused `_scaled_timeout()` function. `_scaled_sleep()` for Wait states and retry intervals remains unchanged.

## Test

`test_sfn_wait_scale_zero_does_not_timeout_lambda_tasks` — sets `SFN_WAIT_SCALE=0`, runs an SFN that invokes a Lambda with a 0.5s sleep, and asserts it succeeds without `States.Timeout`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **SFN callback/activity timeout not scaled** — `SFN_WAIT_SCALE=0` no longer causes `States.Timeout` on activity tasks and `waitForTaskToken` callbacks. The scale factor was incorrectly applied to functional timeouts (which must wait for real work to complete), not just Wait state sleeps and retry intervals. Contributed by @jayjanssen
+
+---
+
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Sorry for the correction @jayjanssen ! Merging it

Issue #341, opened 2026-04-15
Cognito access tokens missing cognito:groups claim


subrotosanyal:
### Affected Service

Router / General

### What happened?

When a user is added to a Cognito group via admin_add_user_to_group, the cognito:groups claim is not included in the access token returned by initiate_auth. Real AWS Cognito includes this claim automatically.

**Actual behavior**
The access token does not contain cognito:groups at all. The claim is completely absent.

### Reproduction steps

```python
import base64
import json
import time

import boto3
import httpx
from testcontainers.core.container import DockerContainer

MINISTACK_IMAGE = "ministackorg/ministack:latest"

container = (
    DockerContainer(MINISTACK_IMAGE)
    .with_env("SERVICES", "cognito-idp")
    .with_exposed_ports(4566)
)

with container:
    host = container.get_container_host_ip()
    port = int(container.get_exposed_port(4566))
    url = f"http://{host}:{port}"

    # Wait for MiniStack to be ready
    for _ in range(30):
        try:
            if httpx.get(f"{url}/_localstack/health", timeout=5).status_code == 200:
                break
        except Exception:
            pass
        time.sleep(2)

    client = boto3.client(
        "cognito-idp",
        region_name="us-east-1",
        endpoint_url=url,
        aws_access_key_id="test",
        aws_secret_access_key="test",
    )

    # Create pool, client, group, and user
    pool_id = client.create_user_pool(PoolName="test")["UserPool"]["Id"]
    client_id = client.create_user_pool_client(
        UserPoolId=pool_id,
        ClientName="frontend",
        ExplicitAuthFlows=["USER_PASSWORD_AUTH"],
        GenerateSecret=False,
    )["UserPoolClient"]["ClientId"]

    client.create_group(UserPoolId=pool_id, GroupName="admin")
    client.create_group(UserPoolId=pool_id, GroupName="superadmin")

    client.admin_create_user(
        UserPoolId=pool_id,
        Username="testuser",
        TemporaryPassword="Temp1234!",
        MessageAction="SUPPRESS",
    )
    client.admin_set_user_password(
        UserPoolId=pool_id,
        Username="testuser",
        Password="Test1234!",
        Permanent=True,
    )
    client.admin_add_user_to_group(
        UserPoolId=pool_id, Username="testuser", GroupName="admin"
    )
    client.admin_add_user_to_group(
        UserPoolId=pool_id, Username="testuser", GroupName="superadmin"
    )

    # Verify groups are assigned
[...]

PR #342, opened 2026-04-15
fix(cognito): include cognito:groups claim in access and ID tokens

subrotosanyal:
https://github.com/ministackorg/ministack/issues/341
## Summary

- `initiate_auth` / `admin_initiate_auth` now include the `cognito:groups` claim in both access and ID tokens when the user belongs to one or more groups
- Real AWS Cognito includes this claim automatically; MiniStack was not threading the already-populated `user["_groups"]` list through to the JWT builder
- Two-line fix: `_build_auth_result` reads `_groups` and passes it to `_fake_token`, which adds `cognito:groups` to both token types

## Problem

Applications that use `cognito:groups` for role-based access control (the standard AWS Cognito pattern) get tokens without group claims from MiniStack, even though `admin_add_user_to_group` correctly populates the internal `_groups` list.

## Changes

| File | Change |
|------|--------|
| `ministack/services/cognito.py` | Add `groups` param to `_fake_token`; set `cognito:groups` claim for access + id tokens; read `_groups` in `_build_auth_result` |
| `tests/test_cognito.py` | New `test_cognito_groups_in_auth_tokens` — creates user in 2 groups, authenticates, asserts claim present in both tokens |
| `CHANGELOG.md` | Added entry under `[Unreleased]` |

## Test plan

- [x] Confirmed bug on unpatched image: `cognito:groups` = `*** NOT PRESENT ***`
- [x] Rebuilt image with fix: `cognito:groups` = `['admin']` in both tokens
- [x] All 72 cognito tests pass (71 existing + 1 new)
- [x] Backward-compatible — `groups` defaults to `None`, existing callers unaffected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **Cognito `cognito:groups` missing from tokens** — `initiate_auth` and `admin_initiate_auth` now include the `cognito:groups` claim in both access and ID tokens when the user belongs to one or more groups. Real AWS Cognito includes this claim automatically; MiniStack was not threading the already-populated `_groups` list through to the JWT builder.
+
+---
+
 ## [1.2.16] — 2026-04-15
 
 ### Added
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Good fix @subrotosanyal ! I will add one small thing before release (the account_scope_dict so it is consistent with multitenancy) Thanks!

subrotosanyal:
Thank you @Nahuel990 .

PR #349, opened 2026-04-16
Fix stream when creating DynamoDB via CloudFormation 

davidtme:
This pull request improves support for DynamoDB table stream specifications in CloudFormation stacks, ensuring that tables with only `StreamViewType` specified (and not `StreamEnabled`) are handled correctly. It also adds a test to verify this behavior.

**DynamoDB StreamSpecification handling:**

* Updated `_ddb_create` in `provisioners.py` to automatically enable streams (`StreamEnabled: True`) when `StreamViewType` is present but `StreamEnabled` is not specified, aligning with CloudFormation expectations.

**Testing improvements:**

* Added `test_cfn_dynamodb_stream_spec` to `test_cfn.py` to verify that DynamoDB tables created via CloudFormation with only `StreamViewType` set have streams enabled, a valid stream ARN output, and correct table description fields.

### DIFF
--- a/ministack/services/cloudformation/provisioners.py
+++ b/ministack/services/cloudformation/provisioners.py
@@ -314,6 +314,8 @@ def _ddb_create(logical_id, props, stack_name):
     lsis = props.get("LocalSecondaryIndexes", [])
 
     stream_spec = props.get("StreamSpecification", {})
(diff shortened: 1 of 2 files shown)


Nahuel990:
Amazing fix @davidtme ! I'll tag it for next release, thank you!!

Nahuel990 (approved):

PR #350, opened 2026-04-16
Fix Windows compatibility for Lambda functions

davidtme:
This pull request introduces a more robust, windows compatible and non-blocking approach to capturing and handling `stderr` output from Lambda worker processes in the `ministack` runtime. It replaces the previous polling method with a dedicated background thread and queue, ensuring that `stderr` output is drained efficiently and does not interfere with repeated Lambda invocations. Additionally, a new test verifies the reliability of warm Lambda invocations even when the worker writes to `stderr`.

**Testing:**

* Added a new test, `test_lambda_warm_invoke_with_stderr_logging`, to verify that repeated (warm) Lambda invocations succeed even when the worker process writes to `stderr`. This helps ensure the stability of the new logging approach.

**Simple Lambda running on windows vis Python**
```JS
exports.handler = async () => { console.log('Hello, CDK!'); };
```

**Before:**

```json
{
  "errorMessage": "[WinError 10038] An operation was attempted on something that is not a socket",
  "errorType": "Runtime.HandlerError"
}
```

**After:**

```
Hello, CDK!
```

### DIFF
--- a/ministack/core/lambda_runtime.py
+++ b/ministack/core/lambda_runtime.py
@@ -16,6 +16,7 @@ import tempfile
 import threading
 import time
 import zipfile
+import queue
 
 logger = logging.getLogger("lambda_runtime")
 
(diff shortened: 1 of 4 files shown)


davidtme:
@Nahuel990 I can only test this on windows via python.

Nahuel990:
Thanks for the addition @davidtme I'll check it in a few hours

davidtme:
I've also started the pollers when using CF

Nahuel990 (approved):

Nahuel990:
Great addition @davidtme we are about to change the way lambda_svc behave but for the moment this is a great spot

PR #366, opened 2026-04-17
Add command arg to call to docker client run

s0rbus:
If a task definition has a command section within the containerDefinitions section which overrides that in the image being used, currently this command is not passed to the run function of the docker client in ecs.py so it is ignored and the default image's command is used. This PR adds the command 'field' to run()'s arguments.


Nahuel990 (approved):

Nahuel990:
Great addition @s0rbus ! Shipping it in v1.2.21

PR #373, opened 2026-04-17
Enhance log forwarding for lambda

Baptiste-Garcin:
Hello !

It's me, again. Log were not forwarded properly when invoking lambda (sync and async).
I found 2 issues

##### Missing log forwarding (18fdcd1)
For the `Event` invocation type, I wrapped the `_execute_function` in a function that logs the return
For the classic path, I simply added a call to the logger

##### Winston Logger (f2078b6)
For the node run time, the console.log() was redirected to stderr and then processed by the parent process. Classic production logger such as Winston use `process.stdout.write` so I put the redirection on this low level call. This way, I handle logger as well as `console.log()`

### DIFF
--- a/ministack/core/lambda_runtime.py
+++ b/ministack/core/lambda_runtime.py
@@ -90,13 +90,12 @@ const http = require("http");
 const https = require("https");
 const url = require("url");
 
-// Redirect all console methods to stderr so stdout stays clean for JSON-line protocol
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this addition @Baptiste-Garcin we are finishing the wiring to CloudWatch (where logs should really be, so there will be an upcoming change in the way logs are displayed to match real AWS behaviour)

PR #377, opened 2026-04-18
ci: Add Docker Publish workflow for pull requests

jgrumboe:
Implements docker images as preview builds as discussed in #288

### DIFF
--- a/.github/workflows/docker-publish-on-pr.yml
+++ b/.github/workflows/docker-publish-on-pr.yml
@@ -1,0 +1,71 @@
+name: Docker Publish on PR
+
+on:
+  pull_request:
+
+env:
+  DOCKERHUB_IMAGE_ORG: ministackorg/ministack-preview-build
+
+jobs:
+  publish-pr:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jgrumboe:
@Nahuel990 So far it works.
I'm wondering if we should also add the ${{github_sha}} to the suffix? Or does that get too many images in docker hub? (do we care? maybe not)

Nahuel990:
@jgrumboe Up to you... we dont have limit in Docker

jgrumboe:
@Nahuel990 ready for review.
I removed the git tag from the suffix as I couldn't find a way quickly to get the correct tag the branch/PR is based on.
I think it'd good enough to have the PR number and git short sha in the image tag to identify where the image is coming from.

Nahuel990:
Just a heads up that GITHUB_SHA will give us a different PR number, the "hidden" Gtihub sha is {{github.event.pull_request.head.sha }}

https://github.com/orgs/community/discussions/26325#discussioncomment-3251468

Nahuel990 (approved):

PR #384, opened 2026-04-19
Add Resource Groups Tagging API - Phase 3: TagResources and UntagResources (#382)

### LINKED ISSUE #382: Resource Groups Tagging API — Phase 3 (TagResources, UntagResources)
### AWS Service name

Resource Groups Tagging API (`resourcegroupstaggingapi`) — Phase 3 continuation of #379

### Operations needed

**Phase 3 (this issue) — write operations**
- `TagResources` — apply tags to resources by ARN across all 15 supported services
- `UntagResources` — remove tag keys from resources by ARN across all 15 supported services

### Use case

Phase 1 (#372) and Phase 2 (#380) landed read operations (`GetResources`, `GetTagKeys`,
`GetTagValues`) across 15 services. This completes the core API surface by adding write
operations, unblocking:

- CDK's `Tags.of(scope).add(...)` which calls `TagResources` to apply tags post-creation
- Terraform's `aws_resourcegroupstaggingapi_resources` workflows that tag resources after
  provisioning via `TagResources`
- Any tool that uses `UntagResources` to clean up or rotate tags on existing resources

Two services require special handling due to non-standard internal tag formats:
- **KMS** stores tags as `[{"TagKey": k, "TagValue": v}]` — write-back must preserve this
- **ECS** stores tags as `[{"key": k, "value": v}]` (lowercase) — write-back must preserve this
- **SQS** is indexed by queue URL, not ARN — requires a reverse scan to resolve ARN → URL

### Changes planned

| File | What changes |
|---|---|
| `ministack/services/tagging.py` | ARN routing helper, 16 tag writers + `_WRITERS` dict, 16 tag removers + `_REMOVERS` dict, `TagResources` + `UntagResources` handlers |
| `tests/test_tagging.py` | 17 integration tests (TagResources, UntagResources, cross-operation roundtrips) |

No changes to `router.py`, `app.py`, or `tests/conftest.py`.

- [x] Yes, I'd like to implement this and open a PR

AdigaAkhil:
Implements `TagResources` and `UntagResources` operations for the Resource Groups Tagging API, completing Phase 3 of #382.

### What's included

- 16 per-service tag writers covering all services supported by `GetResources` (S3, Lambda, SQS, SNS, DynamoDB, EventBridge, KMS, ECR, ECS, Glue, Cognito IdP, Cognito Identity, AppSync, Scheduler, CloudFront, EFS)
- 16 per-service tag removers with symmetric coverage
- ARN-based routing via `_service_key_from_arn` dispatching to `_WRITERS`/`_REMOVERS`
- `FailedResourcesMap` in response for unsupported or erroring ARNs (matching AWS behaviour)
- 14 integration tests covering tag/untag per-service, merge, overwrite, multi-ARN, unknown service, and full roundtrip scenarios
- Bug fix: deduplicate shared collectors to prevent duplicate Cognito resources in `GetResources` responses


ministack/services/tagging.py
```diff
@@ -205,6 +205,271 @@ def _matches_tag_filters(tags, tag_filters):
...
+def _service_key_from_arn(arn):
+    """Extract service segment from ARN (e.g. 'arn:aws:s3:::...' → 's3')."""
+    parts = arn.split(":")
+    return parts[2] if len(parts) >= 3 else ""
+
+
+# ── Per-service tag writers ───────────────────────────────────────────────────
+
+def _write_s3(arn, tags):
+    import ministack.services.s3 as svc
```
jgrumboe, line 217: I'm not a python dev at all but with the new lazy loading in app.py, shouldn't the _get_module() be used instead of straight import?

AdigaAkhil: @jgrumboe  ,`_get_module()` is in **app.py**, which already imports all services; pulling it into tagging.py would create a circular import. The function-scoped imports here achieve the same lazy loading through Python's sys. modules cache, so the behaviour is identical.

jgrumboe: But the direct import circumvents also the error handling from the lazy loader in app.py, I think. 

Somehow, I'm also not happy about calling the internal functions of the service modules.

Couldn't/shouldn't we route service-to- service requests like this tagging/untagging somehow centrally?

AdigaAkhil: @jgrumboe, fair points. On the error handling, the try/except wrappers at each call site cover import failures and surface them in FailedResourcesMap, so the behaviour is similar to _ErrorModule.                                                                                                                                                                                                                                                                                   

On accessing internals directly, provisioners.py does the same across ~30 services (_sns._topics, _sqs._queues, etc.), so I followed that precedent here. If the preference is to move towards per-service tag APIs, happy to refactor both together as a separate PR.

jgrumboe: @AdigaAkhil thanks for the explanation. 
I see that such direct calls exist also elsewhere.
You're right that a separate PR is probably better to "fix" that and introduce a better pattern.

I like to also wait for @Nahuel990 to comment what he thinks.

Thanks so far for your contribution!

Nahuel990: @AdigaAkhil overall looks good, a few things to consider. The error code is InternalServiceException instead of FailedResourcesMap, and we are using AccountScopedDict to handle multitenancy, also this is very very minor but get used to write a standard docstring on top of the file. I will merge this during the day .

@jgrumboe about your suggestion is totally valid, ideally we would have a directory per service with the service and a thin wrapper inside that would expose an API + a central registry at the app.py level. So far we choose simplicity for dev speed wihtout affecting performance, but it's a nice addition for a 1.4 version

AdigaAkhil: @Nahuel990  , should I make the error-code changes and the doc string changes in this PR?
OR  should I address this in a separate PR?

jgrumboe: @Nahuel990 I've got you covered with a "minimal service registry" for the start #391 

This was a change I had already prepared in anticipation of your lazy loading.

Nahuel990: @AdigaAkhil no worries, it was more a suggestion for future PRs I will merge this as is, no need for extra modifications

Nahuel990 (approved):

Nahuel990 (approved):

PR #403, opened 2026-04-20
Allow user pool client secrets to be created by Cloudformation

mgius-ae:
See
https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-cognito-userpoolclient.html#aws-resource-cognito-userpoolclient-properties for docs on this.

I did not see any existing testing for Cognito CloudFormation flows, which is why I did not add any new tests to this, but I did verify it works locally via a CDK based stack deployment.


Nahuel990:
Thanks for the addition @mgius-ae and don't worry I'll add the tests before releasing this, it will go in the next iteration.

Nahuel990 (approved):

PR #405, opened 2026-04-20
fix: Raise `ReceiptHandleIsInvalid` on singular `DeleteMessage` and `ChangeMessageVisibility`

nigel-campbell:
`DeleteMessage` and `ChangeMessageVisibility` silently succeeded when given an unrecognised receipt handle. This is inconsistent with:

- **Real AWS SQS**, which returns HTTP 400 `ReceiptHandleIsInvalid` — documented in the [DeleteMessage API reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_DeleteMessage.html) and [troubleshooting guide](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/troubleshooting-api-errors.html)
- **The existing batch equivalents** (`DeleteMessageBatch`, `ChangeMessageVisibilityBatch`) in the same file, which already return `ReceiptHandleIsInvalid` per-entry correctly

### DIFF
--- a/ministack/services/sqs.py
+++ b/ministack/services/sqs.py
@@ -398,9 +398,11 @@ def _act_delete_message(data: dict, qurl: str) -> dict:
     # Only remove messages whose receipt_handle is set and matches.
     # Messages that have never been received (receipt_handle is None) are never
     # accidentally removed by an empty or unrelated receipt handle.
+    found = False
     kept = []
     for m in q["messages"]:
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks for this addition @nigel-campbell We will be releasing it in the next iteration.

Nahuel990 (approved):

PR #416, opened 2026-04-21
fix: send s3:TestEvent to all destinations on PutBucketNotificationConfiguration

nigel-campbell:
From the [PutBucketNotificationConfiguration API reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketNotificationConfiguration.html):

> After Amazon S3 receives this request, it first verifies that any Amazon SNS or SQS destination exists, and that the bucket owner has permission to publish to it by sending a test notification.

Attempts to match the real AWS S3 behavior (and LocalStack) where configuring bucket notifications immediately delivers a flat s3:TestEvent payload (no Records wrapper) to every SQS, SNS, and Lambda destination in the config.

### DIFF
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -1096,6 +1096,7 @@ def _put_bucket_notification(name: str, body: bytes):
     if name not in _buckets:
         return _no_such_bucket(name)
     _bucket_notifications[name] = body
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Awesome! Great catch!

PR #444, opened 2026-04-24
Fix ApiGateway REST (v1): map REST UpdateStage patches into methodSettings

david-hay:
## Summary
Fixes **API Gateway REST (v1) `UpdateStage`** so patch paths used by Terraform’s **`aws_api_gateway_method_settings`** are applied to **`methodSettings`** correctly (e.g. **`*/*`**) instead of creating bogus nested keys on the stage object.
## Problem
The AWS API (and the Terraform AWS provider) send **`UpdateStage`** operations with paths like:
- `/*/*/metrics/enabled`
- `/*/*/logging/loglevel`
i.e. **`/{resourcePath}/{httpMethod}/…`**, with **`~1`** in path segments where the resource path contains **`/`** (JSON Pointer encoding).
MiniStack previously routed all stage patches through a generic JSON patch helper that split on **`/`**. For **`/*/*/metrics/enabled`**, that produced structure under **`stage["*"]["*"]…`** instead of **`stage["methodSettings"]["*/*"]…`**, so **`GetStage`** never returned a **`methodSettings["*/*"]`** entry. The provider then failed refresh with an **empty result** when reading method settings.
## Solution
- Parse **UpdateStage** method-setting paths per [API Gateway patch documentation](https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html) (**`UpdateStage`** / **`/{resourcePath}/{httpMethod}/…`**).
- Decode **`~0`** / **`~1`** in each path segment.
- Map leaf segments (e.g. **`metrics/enabled`**) to the same camelCase keys **`GetStage`** returns (**`metricsEnabled`**, etc.).
- Merge in sensible defaults when first creating a method-setting block.
- Leave **`/variables/…`**, **`/deploymentId`**, and other stage patches to the existing generic patch logic (avoid treating **`/variables/GET`**-style paths as method settings where inappropriate).
## Testing
- **`tests/test_apigatewayv1.py`**: **`test_apigwv1_update_stage_method_settings_wildcard`** — **`update_stage`** with **`/*/*/metrics/enabled`** and **`/*/*/logging/loglevel`**, then **`get_stage`** asserts **`methodSettings["*/*"]`**.
- Manual sanity: encoded resource paths such as **`/a~1b/GET/metrics/enabled`** → key **`a/b/GET`**.
## Scope
- **`ministack/services/apigateway_v1.py`**
- **`tests/test_apigatewayv1.py`**
---


Nahuel990 (approved):

Nahuel990:
Thanks @david-hay !

PR #453, opened 2026-04-24
feat: add DOCKER_NETWORK env var for RDS, EKS, ElastiCache container networking

bognari:
## Summary

- Add `DOCKER_NETWORK` env var so RDS, EKS, and ElastiCache containers join an explicit Docker network instead of relying on auto-detection via `$HOSTNAME` (which fails in docker-compose)
- When set, `Endpoint.Address` returns the container IP instead of `localhost`, making endpoints routable from other containers (e.g. Lambda)
- Lambda now reads `DOCKER_NETWORK` first and falls back to `LAMBDA_DOCKER_NETWORK` for backwards compatibility
- E2E tests for RDS and ElastiCache proving Lambda (Python + JS) can TCP-connect over the shared network

## Motivation

When MiniStack runs inside docker-compose, `_get_ministack_network()` looks up its own container via `$HOSTNAME`. Docker Compose sets `$HOSTNAME` to the `hostname:` directive value (e.g. `ministack`), not the container ID. The Docker API lookup fails, and RDS/ElastiCache containers land on the default bridge — isolated from MiniStack and Lambda workers.

This follows the existing `LAMBDA_DOCKER_NETWORK` pattern and unifies it into a single `DOCKER_NETWORK` knob for all container-backed services.

## Test plan

- [x] Existing RDS unit tests pass (28/29 — 1 pre-existing import issue)
- [x] Network E2E tests correctly skip when `DOCKER_NETWORK` is not set
- [x] CI/CD pipeline passes
- [x] Manual verification with docker-compose stack and `DOCKER_NETWORK` set

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,6 +635,7 @@ ecs.stop_task(cluster="dev", task=task_arn)
 | `RDS_BASE_PORT` | `15432` | Starting host port for RDS containers |
 | `RDS_TMPFS_SIZE` | `256m` | Tmpfs size for RDS database containers (when `RDS_PERSIST=0`). Set to `2g` or higher for large databases |
 | `RDS_PERSIST` | `0` | Set `1` to use Docker named volumes for RDS containers instead of tmpfs. Storage grows dynamically with no fixed cap |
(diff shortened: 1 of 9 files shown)


Nahuel990 (approved):

Nahuel990:
Hey @bognari very clinical implementation, really nice PR

PR #456, opened 2026-04-24
fix(secretsmanager): honor IncludePlannedDeletion in ListSecrets

### LINKED ISSUE #455: SecretsManager: ListSecrets ignores IncludePlannedDeletion — soft-deleted secrets invisible to list
## Summary

`ListSecrets` with `IncludePlannedDeletion=true` (`--include-planned-deletion` on the CLI) returns an empty list for secrets that were scheduled for deletion via `DeleteSecret` without `ForceDeleteWithoutRecovery`. In real AWS, those secrets appear in the list with `DeletedDate` set until the recovery window elapses.

This breaks any client that polls `ListSecrets` after a delete to confirm the operation landed — the soft-deleted secret simply never shows up, so the caller concludes "secret not found" even though `DescribeSecret` on that same name still works.

## Version

- ministack `1.3.13` (image `ministackorg/ministack:1.3.13`)
- Client: `aws-cli/2` against `AWS_ENDPOINT_URL=http://localhost:4566`

## Minimal reproduction

```bash
docker run -d --name ministack -p 4566:4566 ministackorg/ministack:1.3.13
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_REGION=us-east-1
alias m='aws --endpoint-url=http://localhost:4566'

# Create and soft-delete
m secretsmanager create-secret --name probe --secret-string x
m secretsmanager delete-secret --secret-id probe   # default: scheduled deletion

# DescribeSecret correctly reports DeletedDate (good):
m secretsmanager describe-secret --secret-id probe | grep DeletedDate
# => "DeletedDate": "..."

# But ListSecrets with the flag returns nothing (bug):
m secretsmanager list-secrets --include-planned-deletion
# => {"SecretList": []}

# Expected: one entry with Name=probe and DeletedDate set, matching real AWS.
```

## Expected

Per the [AWS Secrets Manager API reference](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_ListSecrets.html), `IncludePlannedDeletion=true` should include secrets in the `PendingDeletion` state in the response, each with `DeletedDate` populated. `GetSecretValue`/`DescribeSecret` on those secrets already behave correctly in ministack — only `ListSecrets` misses them.

## Actual

`ListSecrets` appears to hard-filter out any secret with a non-null `DeletedDate`, regardless of the `IncludePlannedDeletion` flag. Omitting the flag and passing `true` both return the same empty list.

## Other `SecretsManager` calls tested — all behave correctly

| Call after `delete-secret` (default) | ministack | AWS |
|---|---|---|
| `describe-secret` | returns with `DeletedDate` ✅ | same |
| `get-secret-value` | `InvalidRequestException: marked for deletion` ✅ | same |
| `list-secrets` (no flag) | empty ✅ | empty |
| `list-secrets --include-planned-deletion` | **empty ❌** | includes soft-deleted |

Looks isolated to `ListSecrets`.

weeco:
`ListSecrets` unconditionally filtered out any secret with a non-null `DeletedDate`, so soft-deleted secrets were invisible even when a client explicitly asked for them via `IncludePlannedDeletion=true`. The parameter itself was also never read from the request payload.

Two small changes in `_list_secrets`:

- Parse `IncludePlannedDeletion` (default `False`) and gate the `DeletedDate` filter on it, so soft-deleted secrets come through when requested.
- Include `DeletedDate` on each returned `SecretListEntry` when it's set, mirroring what `DescribeSecret` already does and matching the [AWS `SecretListEntry` spec](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_SecretListEntry.html). Without it clients can't distinguish scheduled-for-deletion entries from active ones.

Added `test_secretsmanager_list_include_planned_deletion` covering the default-hidden path, the flag-visible path, and the presence of `DeletedDate` on the returned entry. All existing secretsmanager tests still pass.

Ran into this while using ministack as the AWS endpoint for a local service that polls `list-secrets --include-planned-deletion` after a `delete-secret` to confirm the soft-delete landed — the empty list surfaced as a "not found" error even though `DescribeSecret`/`GetSecretValue` still resolved the secret correctly.

Fixes #455.

### DIFF
--- a/ministack/services/secretsmanager.py
+++ b/ministack/services/secretsmanager.py
@@ -323,8 +323,12 @@ def _list_secrets(data):
     max_results = min(data.get("MaxResults", 100), 100)
     next_token = data.get("NextToken")
     filters = data.get("Filters", [])
+    include_planned_deletion = bool(data.get("IncludePlannedDeletion", False))
 
-    names = sorted(n for n, s in _secrets.items() if not s.get("DeletedDate"))
+    names = sorted(
+        n for n, s in _secrets.items()
+        if include_planned_deletion or not s.get("DeletedDate")
+    )
 
     for f in filters:
         key = f.get("Key", "")
@@ -352,7 +356,7 @@ def _list_secrets(data):
     secret_list = []
     for n in page:
         s = _secrets[n]
-        secret_list.append({
+        entry = {
             "ARN": s["ARN"],
             "Name": s["Name"],
             "Description": s.get("Description", ""),
@@ -362,7 +366,10 @@ def _list_secrets(data):
             "Tags": s.get("Tags", []),
             "SecretVersionsToStages": _vid_to_stages(s),
             "RotationEnabled": s.get("RotationEnabled", False),
-        })
+        }
+        if s.get("DeletedDate"):
+            entry["DeletedDate"] = s["DeletedDate"]
+        secret_list.append(entry)
 
     resp: dict = {"SecretList": secret_list}
     end = start + max_results
--- a/tests/test_secretsmanager.py
+++ b/tests/test_secretsmanager.py
@@ -337,3 +337,26 @@ def test_secretsmanager_get_by_partial_arn(sm):
     assert partial_arn != full_arn
     assert sm.get_secret_value(SecretId=partial_arn)["SecretString"] == "partial-arn-value"
 
+
+def test_secretsmanager_list_include_planned_deletion(sm):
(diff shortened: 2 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Awesome! Thanks @weeco

PR #471, opened 2026-04-25
feat(ecr): Docker daemon integration for ECR service

KennyStryker:
## Summary

Integrate local Docker daemon image management into the ECR service. MiniStack ECR now synchronizes with the host's Docker daemon, allowing Docker images to be listed, managed, and cleaned up as part of ECR repository operations.

**Changes include:**
- Lazy-load Docker client connection on first use
- Query Docker daemon for images tagged with repository URIs
- Merge Docker daemon images with in-memory ECR image registry
- Automatically remove Docker images when repositories are deleted or force-deleted
- Clean up all Docker images during service reset

## Motivation

Previously, ECR only maintained an in-memory image registry. While this worked for simulated push/pull operations, it didn't account for users potentially having actual Docker images on their local system. This created an inconsistency where images visible in `docker images` weren't reflected in ECR API responses, and deleting ECR repositories didn't clean up corresponding Docker images on the host.

## Design Notes

- If Docker daemon is unavailable or the `docker` package isn't installed, the service continues to work with in-memory images only.
- Docker daemon images are merged with in-memory images and duplicate tags are skipped to avoid listing the same image twice.
- Docker client is only instantiated when first needed, minimizing overhead if not used.
- All Docker operations are wrapped in try/except, so failures are logged but don't crash the service.

## Test

Existing tests in `tests/test_ecr.py` all passed.

## Potential future considerations

`_get_docker` method  is now repeated three times: `ecr.py`, `elasticache.py`, `rds.py`, maybe in future it can be moved in a single utility file and referenced from there.

### DIFF
--- a/ministack/services/ecr.py
+++ b/ministack/services/ecr.py
@@ -726,6 +726,10 @@ def _repo_shape(repo):
 
 
 def reset():
+    docker_client = _get_docker()
+    if docker_client:
+        for name in list(_repositories.keys()):
+            _remove_docker_images_for_repo(name)
     _repositories.clear()
     _images.clear()
     _lifecycle_policies.clear()
--- a/ministack/services/ecr.py
+++ b/ministack/services/ecr.py
@@ -23,6 +23,8 @@ _images = AccountScopedDict()
 _lifecycle_policies = AccountScopedDict()
 _repo_policies = AccountScopedDict()
 
+_docker = None
+
 
 # ── Persistence ────────────────────────────────────────────
 
@@ -165,9 +167,13 @@ def _delete_repository(data):
         return error_response_json("RepositoryNotFoundException",
                                    f"The repository with name '{name}' does not exist", 400)
 
-    if not force and _images.get(name):
+    if not force and _all_images(name):
         return error_response_json("RepositoryNotEmptyException",
                                    f"The repository with name '{name}' is not empty", 400)
+
+    # Remove matching Docker images when force-deleting
+    if force:
+        _remove_docker_images_for_repo(name)
 
     repo = _repositories.pop(name)
     _images.pop(name, None)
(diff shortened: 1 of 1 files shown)


Nahuel990:
@KennyStryker Thanks for the contribution, but we cant merge this. ECR is a registry, AWS doesn’t read from your local Docker image cache, so reflecting the host daemon moves us away from AWS. It also breaks reset() would silently delete images from the host daemon, and persistence becomes non-deterministic because list-images results would depend on whatever happens to be in the local cache at the time

https://aws.amazon.com/ecr/faqs/
https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html

KennyStryker:
Thanks for the feedback, @Nahuel990. I understand your concerns about AWS parity and the potential for `reset()` to be destructive.

However, to clarify the safety side, the current logic already implements strict namespace filtering. In `_get_docker_images`, the service only interacts with images whose tags start with the ECR repositoryUri. This means the user's images are completely invisible to and untouched by MiniStack.

Regarding the AWS parity argument, while AWS doesn't read from a local daemon, a local simulator's storage is effectively the host's filesystem/daemon. This integration allows the ECR API to return real metadata (size, digests) for images already on the host, which I believe would be helpful in local integration testing.

In a typical local workflow, a user has an original image (e.g., my-local-image:v1) and a second tag specifically for the registry (e.g., 000000000000.dkr.ecr.us-east-1.amazonaws.com/my-local-image:v1). My implementation ensures that reset() or delete methods only untags the ECR-specific alias. The original image and its underlying layers remain perfectly intact on the host daemon.

Nahuel990:
@KennyStryker I understand your point and what you are trying to do, but we’re not taking shortcuts. In AWS, image metadata exists in the registry because something explicitly pushed it (you start form an empty registry). Your PR returns data for any local image whose tag matches the prefix, which also introduces a minor risk, anything a developer pulled from real ECR  could be listed via list-images to anyone who can reach the MiniStack endpoin

KennyStryker:
Oh, makes sense, fair enough. Closing this PR. Cheers.

PR #470, opened 2026-04-25
feat(stepfunctions): add state machine versioning API

mattwang44:
## Summary

Adds the Step Functions versioning family AWS shipped in 2023: `PublishStateMachineVersion`, `DeleteStateMachineVersion`, `ListStateMachineVersions`, plus `DescribeStateMachine` support for qualified version ARNs (`arn:…:stateMachine:<name>:<N>`) and the `publish=True` parameter on `CreateStateMachine`.

Without these, any AWS SDK call to `ListStateMachineVersions` returns `InvalidAction: Unknown action`. **terraform-provider-aws v6 unconditionally calls `ListStateMachineVersions` on every `aws_sfn_state_machine` refresh** — so the missing actions also broke `terraform plan` against MiniStack for any `aws_sfn_state_machine` resource regardless of whether versioning was opted into.

### Version identity and ARNs

- Version numbers are **monotonic per state machine** starting at 1 and **never reused after delete**. A new `lastVersionNumber` high-water mark on the base state machine enforces this — same contract as real AWS.
- Version ARN: `arn:aws:states:<region>:<acct>:stateMachine:<name>:<N>`.
- A new module-level `_state_machine_versions` `AccountScopedDict` in `stepfunctions.py` stores each snapshot's `definition` / `roleArn` / `type` / `loggingConfiguration` / `description` / `revisionId`, persisted alongside the existing state machine state via the standard `get_state` / `restore_state` hooks.

### `revisionId` as optimistic concurrency

Real AWS rotates `revisionId` on every `CreateStateMachine` / `UpdateStateMachine`. `PublishStateMachineVersion`'s optional `revisionId` parameter is a **precondition**: when supplied, it must match the state machine's current `revisionId`, otherwise the call raises `ConflictException`. Callers use this to refuse publishing a snapshot that shifted under them.

- `_create_state_machine` initialises `revisionId` (`new_uuid()`)
- `_update_state_machine` rotates `revisionId` on every edit
- `_publish_state_machine_version` checks the caller-supplied `revisionId` against `sm['revisionId']` and raises `ConflictException` on mismatch

### `CreateStateMachine` `publish=True`

`CreateStateMachine` accepts `publish=True` per AWS's contract: the call auto-publishes v1 in the same request and the response carries `stateMachineVersionArn`. `_create_state_machine` bumps the high-water mark, snapshots, and emits the version ARN.

### `DescribeStateMachine` accepts a version ARN

[...]

PR #472, opened 2026-04-25
feat(stepfunctions): add state machine alias API

mattwang44:
## Summary

Stacks on top of #470. Adds the Step Functions alias family AWS shipped in 2023:

- `CreateStateMachineAlias`
- `UpdateStateMachineAlias`
- `DeleteStateMachineAlias`
- `DescribeStateMachineAlias`
- `ListStateMachineAliases`
- `DescribeStateMachine` support for alias ARNs (`arn:…:stateMachine:<name>:<aliasName>`)
- The alias-awareness cross-check on `DeleteStateMachineVersion` that was deferred from #470

Aliases let callers address a state machine by a stable pointer whose `routingConfiguration` distributes invocations across one or two versions by weight.

**Same terraform-provider-aws v6 motivation as #470**: the provider unconditionally calls `ListStateMachineAliases` on every `aws_sfn_state_machine` refresh, so without this any `terraform plan` returned `InvalidAction`.

## Validation (matched to AWS)

- **Alias name** matches AWS regex `^[0-9A-Za-z_-]+$`, length 1–80.
- **`routingConfiguration`** is a list of 1–2 entries, each with int `weight` in `[0, 100]`, weights summing to 100. The list type and each entry's dict shape are checked **before any indexing**, so a malformed payload (a dict, a scalar, …) returns `ValidationException` instead of crashing on `routing[0]`.
- **No duplicate `stateMachineVersionArn`** across entries.
- Every referenced version ARN must belong to the **same** state machine and must already exist.

## `DescribeStateMachine` accepts alias ARN

Returns state-machine-shaped fields resolved from the base state machine plus the alias's description. `routingConfiguration` is NOT part of `DescribeStateMachine`'s response shape — callers use `DescribeStateMachineAlias` for routing.

## `DeleteStateMachineVersion` alias guard

`_delete_state_machine_version` now scans `_state_machine_aliases` and raises `ConflictException` when an alias still routes to the version, matching AWS's behaviour. This guard lives here rather than in #470 because `_state_machine_aliases` is declared in this PR.

## Tests

Seven integration tests in `tests/test_sfn.py`:

- Single-version alias create/update/list/delete round-trip
- Weights sum ≠ 100 → rejected
- Alias name regex rejects space / length > 80
- Duplicate version entries in `routingConfiguration` rejected
- `DescribeStateMachine` on alias ARN returns base SM fields with `routingConfiguration` explicitly absent
- `DeleteStateMachineVersion` blocked while alias references version; succeeds after alias removed
- Malformed `routingConfiguration` (missing `stateMachineVersionArn` on an entry) surfaces as `ValidationException`, not server crash

`pytest tests/test_sfn.py` passes (existing + new).

## Out of scope

- `StartExecution` / `StartSyncExecution` traffic-splitting based on alias `routingConfiguration` weights. The state lookup correctly resolves an alias ARN to a version, but actual percentage-based routing across two versions per request is not implemented yet — calls to an alias execute against the alias's first routing entry. A reasonable follow-up if anyone needs canary-style execution split.

## Checklist

- [x] Tests added and passing (`pytest tests/test_sfn.py -v`)
- [x] CHANGELOG entry under `[Unreleased]` → `Added`
- [N/A] New service file — extends existing `stepfunctions.py`
- [N/A] SERVICE_REGISTRY / router patterns — already wired for `states`


Nahuel990:
@mattwang44 we already merged this branch. If you need to add changes please do it from main. I'm closing this one

mattwang44:
@Nahuel990 
Thanks for merging #470! Though this PR is actually for a different API family. #470 added the versioning actions while this PR adds the alias actions ([AWS docs](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-state-machine-alias.html#manage-aliases-with-api)) and the alias-awareness cross-check on `DeleteStateMachineVersion` that was deferred from #470. Could you please re-open this PR?

mattwang44:
Failed test cases seem to be unrelated to this PR, and they're passed in my local. Should I try to cope with this issue with this PR?

Nahuel990:
@mattwang44 Ideally when the branch is already merged you should pull from main and create a separate branch, if the branch is not ready to merge you can mark it as a draft. I'll review those failing tests (related to API gateway) and rebase the branch

mattwang44:
Thanks! Just to clarify — the branch has been separate from the start: `feat/stepfunctions-alias-api`, not the same as `feat/stepfunctions-version-api` that became #470. I implemented both in parallel but held this one back until #470 landed to avoid sending two at once. This PR is ready at the time it's opened. Though I should have rebased the branch the latest main branch again before opening the PR (will do next time).

Nahuel990:
Sorry my bad, I confused branch names and saw you were the author in both so I assume you were pushing extra content in the same, thanks for this. I'll merge it later

Nahuel990 (approved):

PR #477, opened 2026-04-26
feat(cli): add --host / --port options to `ministack` command

mattwang44:
## Summary

`ministack`'s bind interface and port were configurable only through env vars (`GATEWAY_PORT` / `EDGE_PORT`), and the bind interface had no env knob at all (it was hard-coded to `0.0.0.0`). Neither was discoverable from `ministack --help`, so anyone reading the help output had no way to know they could change them.

This PR adds two CLI flags:

- `-p` / `--port PORT` — overrides `GATEWAY_PORT` / `EDGE_PORT`. Default: `4566`.
- `-H` / `--host HOST` — interface to bind on. Default: `0.0.0.0`. Use `127.0.0.1` to restrict to loopback (e.g. on shared dev machines), or a specific NIC address.

**Precedence:** CLI flag > env var > default. Existing setups keep working unchanged because env vars are still consulted when no CLI flag is given.

## Why `--host` is a separate concept from `MINISTACK_HOST`

These are two different things and the help text spells the difference out so they don't get confused:

| | What it controls | Default |
|---|---|---|
| `--host` (this PR) | The TCP **bind interface** the server listens on | `0.0.0.0` |
| `MINISTACK_HOST` (existing env var) | The **virtual hostname** used for S3 virtual-host-style URLs (`<bucket>.s3.<MINISTACK_HOST>`) and execute-api routing (`<api-id>.execute-api.<MINISTACK_HOST>`) | `localhost` |

A user who wants MiniStack to listen only on loopback should pass `--host 127.0.0.1`. They should NOT be touching `MINISTACK_HOST` for that purpose.

## What else changed

- The in-use socket probe now targets the actual bind interface (probes `127.0.0.1` when binding `0.0.0.0`; probes the explicit host otherwise).
- Detached and foreground hypercorn config use `{bind_host}:{port}` rather than the old hard-coded `0.0.0.0:{port}`.
- The "port already in use" error message now suggests `ministack --port 4567` rather than the old `GATEWAY_PORT=4567 ministack` form.
- Foreground startup logs `MiniStack listening on <host>:<port> (PID …)` so users can confirm what they're bound to.

## Verified

| Scenario | Result |
|---|---|
| `ministack --help` shows both new options | ✅ |
| `ministack -d` (no flags, no env) | ✅ binds `0.0.0.0:4566` (existing behavior) |
| `ministack -d --port 4599` | ✅ binds `0.0.0.0:4599`; `4566` stays free |
| `ministack -d --host 127.0.0.1 --port 4598` | ✅ reachable via `127.0.0.1`, **not** reachable via the LAN IP — confirms loopback-only |
| `GATEWAY_PORT=4597 ministack -d` | ✅ env var still honoured (backward compat) |
| `GATEWAY_PORT=4500 ministack -d --port 4596` | ✅ CLI wins; nothing binds on `4500` |
| `ministack --stop --port <p>` | ✅ finds the right PID file by port |

## Out of scope

- **Dockerfile `ENTRYPOINT`**: the published image's entrypoint is `python -m hypercorn ministack.app:app --bind 0.0.0.0:4566`, which bypasses the `ministack` CLI entirely. Plumbing these flags through to docker users would require a separate PR (entrypoint script + arg forwarding semantics). This PR only touches the `ministack` CLI itself.
- **`BIND_HOST` env var**: deliberately not added. Users who want env-var-only configuration already have `GATEWAY_PORT`; adding a new env knob for the bind interface would create three ways to configure the same thing. The CLI flag should be enough; if a maintainer wants to add an env var fallback for parity, that's a one-liner (`default=os.environ.get("BIND_HOST", "0.0.0.0")`).

## Checklist

- [x] Verified end-to-end on macOS host (Python 3.12) — full test matrix above
- [x] CHANGELOG entry under `[Unreleased]` → `Added`
- [x] Defaults unchanged → no impact on existing setups
- [N/A] New service file — extends `ministack/app.py` `main()`
- [N/A] SERVICE_REGISTRY / router patterns — unrelated to this change


Nahuel990:
@mattwang44 I get what you're trying to do here, but this is "dev-convenience" and we lean more on the "AWS alignment" side. It's not even something other AWS emulatorssupport

For the two scenarios:
Changing the port: the env var works today (GATEWAY_PORT=4567 ministack), and Docker users
` docker run -p 4567:4566 ministackorg/ministack` Both established patterns.

Binding to loopback only (I imagine for security case): if you're on Docker, the standard pattern already covers this without any MiniStack change:

`docker run -p 127.0.0.1:4566:4566 ministackorg/ministack`

For the pip-install case, I'd rather add a BIND_HOST env var and keep the naming clear (no confusion with MINISTACK_HOST)it stays consistent with our current env-var convention. One line in app.py.

mattwang44:
Thanks for comment! It makes sense.
Closing this in favor of #488

PR #478, opened 2026-04-26
[FEATURE] : implement EventBridge archive event storage and replay dispatch (#475)

### LINKED ISSUE #475: Add EventBridge archive event storage and replay dispatch
### Affected Service

EventBridge (events)

### What happened?

The EventBridge Archive and Replay control plane is fully implemented in
ministack — CreateArchive, DescribeArchive, ListArchives, StartReplay,
DescribeReplay, and ListReplays all return correct responses. However the
data plane is missing:

- PutEvents does not store events into matching archives, so EventCount
  on every archive stays permanently at 0
- StartReplay sets State RUNNING but never actually re-dispatches the
  archived events to the destination bus

Teams that use EventBridge replays for integration testing and debugging
cannot validate their replay workflows locally.

### Root cause (located)

- ministack/services/eventbridge.py:71 — _archives and _events_log
  exist as AccountScopedDict state but are never written to during PutEvents
- eventbridge.py:600 (_put_events) — dispatches events to rules/targets
  but never checks or appends to archives
- eventbridge.py:1052 (_start_replay) — creates replay record with
  State RUNNING and exits; no event iteration or dispatch
- eventbridge.py:104 (get_state) — _archives is not included in the
  persistence snapshot, so archives are lost on restart

### Operations needed

- Wire _put_events to check active archives and append matching events
  including an EventCount increment
- Wire _start_replay to iterate the source archive events, filter by
  EventStartTime and EventEndTime, and dispatch each via the existing
  _dispatch_event path in a background thread
- Set replay State to COMPLETED and ReplayEndTime after dispatch finishes
- Add _archives to get_state and restore_state for persistence

### Expected behavior

After this fix, creating an archive, putting events, and starting a replay
should result in EventCount reflecting stored events and replay State
reaching COMPLETED with events re-dispatched to the destination bus.

### Use case

Step Functions workflows and integration test suites that verify
event-driven flows use StartReplay to re-process historical events
against updated rule targets. Without dispatch, the replay API surface
exists but does nothing, making local testing of replay-dependent
workflows impossible.

### Scope

No new AWS operations required. All changes are internal to the existing
_put_events and _start_replay handlers plus persistence.

AdigaAkhil:
Fixes #475

## What

Wires the EventBridge Archive and Replay data plane. The control plane
was already fully implemented. CreateArchive, StartReplay, DescribeReplay
all returned correct responses. Two gaps remained:

- PutEvents did not write events to matching archives, so EventCount
  stayed permanently at 0
- StartReplay set State RUNNING and exited without dispatching anything

## Changes

ministack/services/eventbridge.py

Added _archive_event() called from _put_events after each event is
dispatched. Scans all active archives, matches on EventSourceArn and
EventPattern, appends the event and increments EventCount.

Replaced the _start_replay stub with a background threading.Thread that
iterates the source archive events, filters by EventStartTime and
EventEndTime, and dispatches each qualifying event to the destination bus
via the existing _dispatch_event path. Sets State COMPLETED and
ReplayEndTime after the thread finishes.

Added _archives to get_state and restore_state so archive data
survives restarts when persistence is enabled.

No new operations, no changes to _dispatch_event, no new dependencies.

## Tests

12 new tests across two commits covering:

Core data plane: archive stores events, pattern filtering, replay
reaches COMPLETED, nonexistent archive returns ResourceNotFoundException

Edge cases: EventCount accumulation across multiple events, empty
EventPattern stores all events from bus, multiple archives on same bus,
time range filtering excludes out-of-window events, empty archive
completes immediately, end-to-end replay delivery via SQS target,
EventCount unchanged after replay, duplicate replay name rejected

### DIFF
--- a/ministack/services/eventbridge.py
+++ b/ministack/services/eventbridge.py
@@ -106,6 +106,7 @@ def get_state():
         "rules": copy.deepcopy(_rules),
         "targets": copy.deepcopy(_targets),
         "tags": copy.deepcopy(_tags),
+        "archives": copy.deepcopy(_archives),
         "replays": copy.deepcopy(_replays),
         "endpoints": copy.deepcopy(_endpoints),
         "partner_event_sources": copy.deepcopy(_partner_event_sources),
@@ -118,6 +119,7 @@ def restore_state(data):
         _rules.update(data.get("rules", {}))
         _targets.update(data.get("targets", {}))
         _tags.update(data.get("tags", {}))
+        _archives.update(data.get("archives", {}))
         _replays.update(data.get("replays", {}))
         _endpoints.update(data.get("endpoints", {}))
         pe = data.get("partner_event_sources")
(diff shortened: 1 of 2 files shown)


Nahuel990:
Awesome! Thanks @AdigaAkhil

PR #481, opened 2026-04-26
fix: CloudFormation Lambda provisioner fetches code from S3

hiddengearz:
### Problem

  When a Lambda function is deployed via CloudFormation with `Code.S3Bucket` / `Code.S3Key` (the standard path for CDK `NodejsFunction`, `PythonFunction`, and any SAM/CFN template that uploads a zip artifact), the
  CFN provisioner `_lambda_create` in `provisioners.py` only handles inline `Code.ZipFile` via `_zip_inline()`. When `ZipFile` is absent — which is always the case for S3-based deployments — `code_zip` is set to
  `None`, and all subsequent invocations hit the early return in `_execute_function_warm`:

  ```python
  if not code_zip:
      return {"body": {"statusCode": 200, "body": "Mock response - no code deployed"}}
  ```

  This means every CDK-deployed Lambda (`cdklocal deploy`) silently returns a mock response instead of executing the real handler. The direct `CreateFunction` API path (`lambda_svc.py:805`) already calls
  `_fetch_code_from_s3()` correctly — only the CloudFormation provisioner was missing the fallback.

  ### Root cause

  `_lambda_create` (line 425) sets:
  ```python
  "code_zip": _zip_inline(code.get("ZipFile"), handler, runtime),
  ```

  `_zip_inline` returns `None` when `source` is `None` (no inline code). The `code_s3_bucket` and `code_s3_key` fields are stored but never used to actually fetch the zip from S3. Meanwhile, CDK's asset pipeline
  uploads the bundled zip to S3 via the bootstrap bucket and references it in the CloudFormation template's `Code.S3Bucket` / `Code.S3Key` properties — so `ZipFile` is always absent.

  ### Fix

  One-line change: fall back to `_fetch_code_from_s3()` when `_zip_inline()` returns `None`, matching the direct API path:

  ```python
  "code_zip": _zip_inline(code.get("ZipFile"), handler, runtime)
      or _lambda_svc._fetch_code_from_s3(code.get("S3Bucket", ""), code.get("S3Key", "")),
  ```

  `_fetch_code_from_s3` is already battle-tested (used by `CreateFunction` and `UpdateFunctionCode`). It reads from the in-memory S3 service and returns `None` gracefully if the object doesn't exist, so the
  existing "Mock response" fallback still applies for functions with no code at all.

  ### Test

  `test_cfn_lambda_s3_code` — uploads a Node.js ESM handler zip to S3, deploys it via a CloudFormation stack with `Code.S3Bucket` / `Code.S3Key`, invokes the function, and verifies:
  - The real handler executes (returns `{ ok: true }`, not the mock response)
  - `AWS_REGION` is available in the Lambda's process environment
  - Stack cleanup via `delete_stack` succeeds

  ### Files changed

  | File | Change |
  |------|--------|
  | `ministack/services/cloudformation/provisioners.py` | Add `_fetch_code_from_s3` fallback in `_lambda_create` (line 425) |
  | `tests/test_cfn.py` | Add `test_cfn_lambda_s3_code` |


Nahuel990:
@hiddengearz Thanks for catching this. CFN-deployed S3-backed Lambdas returning a mock response is a real bug. The fix is on the right track, but a few things need adjusting before merge:

1. The test checks body["region"] == "us-east-1", which relies on AWS_REGION being set in the Lambda. That is not true on main today (covered by #482), so this test will fail unless that PR lands too. Please remove this assertion. The rest of the test is sufficient.

2. _fetch_code_from_s3 is called whenever ZipFile is missing, including for PackageType: Image where S3Bucket andS3Key are empty... This leads to an unnecessary lookup and possible warnings on deploy. Please guard it:

code_zip = _zip_inline(code.get("ZipFile"), handler, runtime)
if code_zip is None and code.get("S3Bucket") and code.get("S3Key"):
    code_zip = _lambda_svc._fetch_code_from_s3(
        code["S3Bucket"],
        code["S3Key"],
        version_id=code.get("S3ObjectVersion"),
    )

3. Code.S3ObjectVersion is ignored. CDK and Terraform both support pinning object versions. Without passing this through, users get the latest object instead of the pinned one. Please thread version_id into _get_object_data(...) this likely needs a small change to the helper signature.

After these changes, this looks good to merge.

hiddengearz:
@Nahuel990 Thanks for the review — pushed 23f07df addressing all three points.
  

1. Dropped the body["region"] assertion; the test now only checks ok: true so it doesn't depend on #482.
2. Added the guard so _fetch_code_from_s3 is only called when S3Bucket and S3Key are both present (Image-package Lambdas now skip the lookup).
3. Threaded Code.S3ObjectVersion through as version_id on _fetch_code_from_s3 and s3._get_object_data. The S3 service's _object_versions registry already stores per-version bytes when bucket versioning is enabled, so pinned versions return the historical object instead of the latest.

Nahuel990 (approved):

PR #480, opened 2026-04-26
feat: CloudFormation provisioners for ApiGatewayV2 Integration + Route

hiddengearz:
### AWS Service name

  API Gateway V2 (CloudFormation)

  ### What was added

  CloudFormation provisioners for `AWS::ApiGatewayV2::Integration` and `AWS::ApiGatewayV2::Route`. This completes the API Gateway V2 CloudFormation surface alongside the existing `Api` and `Stage` provisioners.

  ### Motivation

  CDK `HttpApi.addRoutes()` generates `AWS::ApiGatewayV2::Integration` and `AWS::ApiGatewayV2::Route` resources. Without provisioners for these types, `cdklocal deploy` fails with:

  CREATE_FAILED | AWS::ApiGatewayV2::Integration | Unsupported resource type: AWS::ApiGatewayV2::Integration

  The underlying API Gateway V2 service already supports these operations via SDK (`CreateIntegration`, `CreateRoute`, etc.) — only the CloudFormation provisioners were missing.

  ### Changes

  **`ministack/services/cloudformation/provisioners.py`**
  - `_apigw_v2_integration_create` / `_apigw_v2_integration_delete` — creates/deletes integrations in the `_apigw_v2._integrations` store. Supports `IntegrationType`, `IntegrationUri`, `PayloadFormatVersion`,
  `ConnectionType`, `ContentHandlingStrategy`, and other standard properties. Returns `IntegrationId` via `Fn::GetAtt`.
  - `_apigw_v2_route_create` / `_apigw_v2_route_delete` — creates/deletes routes in the `_apigw_v2._routes` store. Supports `RouteKey`, `Target`, `AuthorizationType`, `ApiKeyRequired`. Returns `RouteId` via
  `Fn::GetAtt`.
  - Both registered in `_RESOURCE_HANDLERS`.

  **`tests/test_cfn.py`** — 6 new tests:
  - `test_cfn_apigwv2_integration_basic` — create + delete with service API verification
  - `test_cfn_apigwv2_route_basic` — create + delete with cross-resource `Ref` (Integration → Route Target)
  - `test_cfn_apigwv2_integration_getatt` — `Fn::GetAtt` resolves `IntegrationId`
  - `test_cfn_apigwv2_route_getatt` — `Fn::GetAtt` resolves `RouteId`
  - `test_cfn_apigwv2_integration_idempotent_delete` — double delete does not crash
[...]

PR #482, opened 2026-04-26
fix: inject standard Lambda env vars in warm-worker execution path

hiddengearz:
### AWS Service name

  Lambda (warm-worker execution path)

  ### What was fixed

  The warm-worker execution path in `lambda_runtime.py` was missing standard Lambda runtime env vars that real AWS always injects — and that the Docker execution path in `lambda_svc.py` already injects. Lambda
  functions running in warm-worker mode (the default for Python and Node.js) ran with an inconsistent environment compared to functions in Docker mode.

  ### Motivation

  Per [AWS Lambda runtime env var docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html):

  > AWS_REGION – The default AWS Region where the Lambda function is executed.
  > AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN – The access keys obtained from the function's execution role.
  > AWS_LAMBDA_FUNCTION_NAME – The name of the function.
  > AWS_LAMBDA_FUNCTION_MEMORY_SIZE – The amount of memory available to the function in MB.
  > AWS_LAMBDA_FUNCTION_VERSION – The version of the function being executed.
  > AWS_LAMBDA_LOG_STREAM_NAME – The name of the Amazon CloudWatch Logs stream for the function.

  These vars are reserved by Lambda and **always set** in the runtime environment. Functions calling `boto3.client(...)`, X-Ray tracers, CloudWatch metric emitters, and any code branching on `$LATEST` vs published
  versions all rely on them. Their absence in warm-worker mode caused silent behavioral divergence from real AWS.

  ### Gap analysis

  Vars the Docker path (`lambda_svc.py:_execute_function_docker`) injects but warm-worker did NOT:

  | Var | Docker | Warm-worker (before) | Warm-worker (after this PR) |
  |---|---|---|---|
  | `AWS_REGION` | ✅ | ❌ | ✅ |
  | `AWS_DEFAULT_REGION` | ✅ | ❌ | ✅ |
  | `AWS_ACCESS_KEY_ID` | ✅ | ❌ | ✅ |
  | `AWS_SECRET_ACCESS_KEY` | ✅ | ❌ | ✅ |
  | `AWS_SESSION_TOKEN` | ✅ | ❌ | ✅ |
  | `AWS_LAMBDA_FUNCTION_NAME` | ✅ | ✅ (pre-existing) | ✅ |
  | `AWS_LAMBDA_FUNCTION_MEMORY_SIZE` | ✅ | ✅ (pre-existing) | ✅ |
  | `AWS_LAMBDA_FUNCTION_VERSION` | ✅ | ❌ | ✅ |
  | `AWS_LAMBDA_LOG_STREAM_NAME` | ✅ | ❌ | ✅ |
  | `_LAMBDA_FUNCTION_ARN` | ✅ | ✅ (pre-existing) | ✅ |
  | `_LAMBDA_TIMEOUT` | ✅ | ❌ | ✅ |

  Plus `AWS_ENDPOINT_URL` (Ministack-specific) — required so SDK calls from the Lambda reach this MiniStack instance instead of a host-mapped port that's unreachable from inside the worker subprocess.

  ### Changes

  **`ministack/core/lambda_runtime.py`** — `_start_process`:
  - Added `setdefault()` for all 9 standard runtime env vars + `_LAMBDA_TIMEOUT`. User-configured `Environment.Variables` still override (matches real AWS).
  - `AWS_ENDPOINT_URL` is **protected**: applied unconditionally with precedence `host process value > internal default`. **Stripped from the init message** sent to the worker subprocess via stdin, so the worker's
  `os.environ.update(init["env"])` cannot re-introduce a function-level override.

  **Why the strip matters**: subprocess.Popen uses `spawn_env` (correctly merged), but the worker script does `os.environ.update(init["env"])` *after startup*. Without filtering, function-level `AWS_ENDPOINT_URL`
  would be re-applied on top of our value, redirecting SDK calls to an unreachable host. This was a latent two-phase application bug — `test_lambda_endpoint_url_not_overridden_by_function_env` was previously
  skipping itself because `AWS_ENDPOINT_URL` wasn't in the host env, hiding the bug.

  **`tests/test_lambda.py`** — `test_lambda_standard_runtime_env_vars_injected`:
  - Strengthened to assert ALL 9 Docker-path vars are present and non-empty (the full set the Docker path injects).
  - Asserts `AWS_SESSION_TOKEN` is *present* (may be empty — boto3 credential chain checks the key explicitly).
  - Asserts function-identity vars (`FUNCTION_NAME`, `MEMORY_SIZE`, `VERSION`) match the configured function.

  ### Test results

  pytest tests/test_lambda.py
  126 pass, 1 skip, 0 fail

  **Negative test verified**: with the 4 missing `setdefault` calls reverted (simulating the original incomplete fix), the strengthened test fails on `AWS_LAMBDA_FUNCTION_VERSION` first — proving the test catches
  the regression. Same negative-test pattern as #480's followup PR.

  ### Why the original PR's tests didn't catch this

  The original test asserted only the 5 vars it added (`AWS_REGION`, `AWS_DEFAULT_REGION`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_ENDPOINT_URL`) — a tautology test that mirrored the implementation
  rather than asserting against the AWS spec or the Docker path it claimed to align with. The strengthened test now asserts against the **full Docker-path set**, so any future divergence between paths fails the
  test.

  ### Checklist

  - [x] AWS spec verified via documentation link above
  - [x] Aligned with Docker execution path (verified by enumerating both)
  - [x] Test asserts ALL 9 standard runtime vars (not just the ones added)
  - [x] Negative test confirmed: test fails without the full fix
  - [x] Full `tests/test_lambda.py` passes (126/126), no regressions
  - [x] Latent bug `test_lambda_endpoint_url_not_overridden_by_function_env` (was skipping itself) now passes correctly

  Refs: previous incomplete fix in this branch


Nahuel990:
Thanks for this. The warm-worker vs Docker-path difference for standard runtime env vars is a real gap and worth fixing. Most of the PR looks good but the issue is AWS_ENDPOINT_URL.... AWS Lambda does not inject this. It is an SDK/testing convention. In real AWS, only Environment.Variables can set it, and the function value takes precedence.

The current implementation overrides that:

spawn_env["AWS_ENDPOINT_URL"] = os.environ.get("AWS_ENDPOINT_URL", f"http://127.0.0.1:{port}")

Combined with the _PROTECTED filter removing it from the init payload, this always overwrites any function-level value and prevents it from being restored later. That breaks existing patterns like function-scoped endpoint routing and diverges from AWS behavior.

Please change it to:

spawn_env.setdefault("AWS_ENDPOINT_URL", f"http://127.0.0.1:{port}")

Since spawn_env = {**os.environ, **env_vars} is already applied earlier, this gives the correct precedence:
function Environment.Variables → host AWS_ENDPOINT_URL → internal default.

The Docker networking concern you mentioned is valid, but the default fallback already handles the unset case. It should not override an explicit user value.

Also:

- Remove the _PROTECTED filter and pass env_vars through init["env"] unchanged. The worker calling os.environ.update(init["env"]) after startup is correct behavior.
- Update or remove test_lambda_endpoint_url_not_overridden_by_function_env. Its current expectation is opposite to AWS behavior. The correct expectation is that the function-level value wins.

With those changes, this is good to merge.

hiddengearz:
Thanks — pushed 717d667. You're right that real AWS doesn't inject `AWS_ENDPOINT_URL` and the function-scoped value should win.

- Switched to `spawn_env.setdefault("AWS_ENDPOINT_URL", f"http://127.0.0.1:{port}")`. Since `spawn_env = {**os.environ, **env_vars}` runs first, precedence is now function `Environment.Variables` → host `AWS_ENDPOINT_URL` → internal default.
- Removed the `_PROTECTED` filter; `init["env"]` is now `env_vars` unchanged so the worker's post-startup `os.environ.update(init["env"])` no longer strips function-level overrides.
- Replaced `test_lambda_endpoint_url_not_overridden_by_function_env` with `test_lambda_function_env_overrides_endpoint_url`, which asserts the function value wins.

Nahuel990 (approved):

PR #490, opened 2026-04-27
fix(sns): Forward SNS attributes for raw SQS subscriptions

arischow:
## Summary

Fix raw SNS-to-SQS fanout so SNS publish `MessageAttributes` are preserved as top-level SQS `MessageAttributes`.

When an SQS subscription has `RawMessageDelivery` set to `"true"`, MiniStack now forwards SNS publish message attributes into the queued SQS message. This matches the behavior expected by SQS consumers that call `receive_message` with `MessageAttributeNames`, and prevents consumers from missing the `MessageAttributes` field on raw SNS-to-SQS messages.

Non-raw SNS-to-SQS delivery remains unchanged.

### DIFF
--- a/ministack/services/sns.py
+++ b/ministack/services/sns.py
@@ -813,7 +813,8 @@ def _fanout(topic_arn: str, msg_id: str, message: str, subject: str,
 
         if protocol == "sqs":
             _deliver_to_sqs(endpoint, envelope, raw, effective_message,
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
Thanks for this @arischow !

PR #491, opened 2026-04-27
fix: restore persisted state on startup for 5 services (autoscaling/backup/eks/scheduler/pipes)

bognari:
## Summary

Fixes a persistence-symmetry gap: 5 services participate in the shutdown save path (`save_all` via `_state_map`) but had no corresponding restore on startup. With `PERSIST_STATE=1`, every autoscaling group / backup vault / EKS cluster / EventBridge schedule / pipe was silently lost on every restart.

`pipes` was additionally missing from `_state_map` entirely, so its state was never even written to disk.

## What changed

| File | Change |
|---|---|
| `ministack/app.py` | `_state_map` lifted to module scope (testable). `pipes` added to it. |
| `ministack/services/autoscaling.py` | Added module-level `try: load_state(...) / restore_state(...)` block. |
| `ministack/services/backup.py` | Same restore block. `get_state()` switched from `dict(_x)` to `copy.deepcopy(_x)` — the `dict()` cast silently dropped every tenant except the current request's. |
| `ministack/services/eks.py` | Same restore block. |
| `ministack/services/pipes.py` | Added `restore_state()` (didn't exist) + restore block. |
| `ministack/services/scheduler.py` | Restore block + same `dict() → deepcopy()` fix. |
| `tests/test_persistence_symmetry.py` *(new, 56 tests)* | Regression coverage. |

The added restore pattern matches the canonical one already used by `services/sqs.py`, `services/dynamodb.py`, and ~36 other service modules.

Restore is gated by `PERSIST_STATE` — `load_state()` returns `None` immediately when the env var is unset. New parametrised test `test_load_state_is_noop_when_persist_state_disabled` enforces this.

## Test plan

- [x] `pytest tests/test_persistence_symmetry.py` — all 56 pass.
- [x] `pytest tests/test_autoscaling.py tests/test_eks.py tests/test_scheduler.py` — all 70 pass (no regressions in the touched services).
- [x] `pytest tests/test_health.py tests/test_multitenancy.py` — all 15 pass.
- [x] Verified pre-existing test failures on `main` (test_backup, test_logs delivery API, test_secretsmanager IncludePlannedDeletion, test_s3 zero-byte chunked, test_ec2 default subnets) are unaffected by this PR.
- [ ] Reviewer check: warm-boot a real ministack with `PERSIST_STATE=1`, create resources in each of the 5 services, restart, verify they survive.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -193,6 +193,33 @@ SERVICE_HANDLERS = {
     for service_name, service_config in SERVICE_REGISTRY.items()
 }
 
+# Maps the on-disk persistence key to the service module name. `save_all`
+# (lifespan.shutdown) consumes this. Restore happens at module import time
+# in each service via its own `load_state()` call (see e.g. services/sqs.py);
+# a small allow-list is also restored centrally by `_load_persisted_state`
+# below. Symmetry between save and restore is enforced by
+# tests/test_persistence_symmetry.py.
+_state_map = {
+    "apigateway": "apigateway", "apigateway_v1": "apigateway_v1",
+    "sqs": "sqs", "sns": "sns", "ssm": "ssm",
+    "secretsmanager": "secretsmanager", "iam": "iam",
+    "dynamodb": "dynamodb", "kms": "kms", "eventbridge": "eventbridge",
+    "cloudwatch_logs": "cloudwatch_logs", "kinesis": "kinesis",
+    "ec2": "ec2", "route53": "route53", "cognito": "cognito",
+    "ecr": "ecr", "cloudwatch": "cloudwatch", "s3": "s3",
+    "lambda": "lambda_svc", "rds": "rds", "ecs": "ecs",
+    "elasticache": "elasticache", "appsync": "appsync",
+    "stepfunctions": "stepfunctions", "alb": "alb",
+    "glue": "glue", "efs": "efs", "waf": "waf",
+    "athena": "athena", "emr": "emr", "cloudfront": "cloudfront",
+    "codebuild": "codebuild", "acm": "acm", "firehose": "firehose",
+    "ses": "ses", "ses_v2": "ses_v2",
+    "servicediscovery": "servicediscovery", "s3files": "s3files",
+    "appconfig": "appconfig", "transfer": "transfer",
+    "scheduler": "scheduler", "autoscaling": "autoscaling",
+    "eks": "eks", "backup": "backup", "pipes": "pipes",
+}
+
 SERVICE_NAME_ALIASES = {
     alias: service_name
     for service_name, service_config in SERVICE_REGISTRY.items()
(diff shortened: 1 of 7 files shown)


bognari:
Pushed review fixes (commit 9216a2c):

1. **`pipes.restore_state()` now starts the poller for RUNNING pipes** — this was a real semantic bug the reviewer caught. `register_pipe()` is the only other place that calls `_ensure_poller()`, and it's not invoked on warm-boot, so persisted pipes silently stopped forwarding events after a restart. Added a regression test (`test_pipes_restore_starts_poller_for_running_pipes`).

2. **`test_pipes_round_trip` now populates a complete pipe record** matching `register_pipe()`'s shape (incl. `Arn`, `CurrentState`, `Tags`, `_positions` entry). Source/Target are intentionally non-DDB/non-SNS so the poller's `_poll_once` skips them quickly even if a sibling test has started the background thread.

3. **File handle leak** in `test_service_has_restore_path` fixed by switching to `pathlib.Path(...).read_text()`.

---

Re the failing CI run: both `test_apigwv2_integration_uri_with_alias_qualifier_resolves_to_alias_target` and `test_apigwv2_integration_wrapped_alias_arn` time out under parallel CI load. This is the same `TimeoutError` pattern that already caused a sibling-set of `test_apigatewayv2.py::test_apigwv2_*` tests to be moved to `_SERIAL_TESTS` in `tests/conftest.py:53-58`. Identical failures appear in main-branch CI runs (e.g. run 24961078426) on tests untouched by this PR. Happy to add these two to the serial allow-list as a follow-up if useful, but it feels orthogonal to the persistence fix.

Nahuel990 (approved):

bognari:
Pushed eager-import fix (commit 825723d) addressing the second-round review.

**Real bug — confirmed and broader than just `pipes`.** Services in `_state_map` that are not in `SERVICE_REGISTRY` are never imported by the lazy router on startup, so their module-level `load_state()` block never fires. `_load_persisted_state()` now eagerly imports them.

Two such services exist today:
- `pipes` — CloudFormation-internal, imported only when a CFN `AWS::Pipes::Pipe` resource is provisioned.
- `ses_v2` — reached only via the `/v2/email/*` path-prefix shortcut in `app.py`. **Caught by the new regression test below**, which is exactly why I wrote it.

For pipes this combines with the `restore_state()` poller-start (the prior round's review fix) so persisted event forwarding actually resumes after warm-boot — not just the in-memory definition.

**Regression test.** New `test_state_map_services_without_endpoint_are_eagerly_imported` derives the set `_state_map - SERVICE_REGISTRY` and asserts each entry is referenced by name in `_load_persisted_state()`'s source. The next time someone adds a non-routable service to `_state_map`, this test fails until they wire it up.

Re the second sub-threshold comment about the brittle source-text scan in `test_service_has_restore_path` — accepted as sub-threshold (Copilot self-flagged it that way too). The functional `_round_trip` tests in this same file are the actual behavior-based check; the structural test is a fast smoke-test on top. Open to revisiting if the suggestion becomes a blocker.

All 58 persistence tests pass.

Issue #438, opened 2026-04-24
aws_iam_policy Description field lost on warm-boot (PERSIST_STATE=1) — forces Terraform replacement


whittin3:
### Summary

`CreatePolicy` accepts a `Description` parameter, but MiniStack never stores it on the policy object and `GetPolicy` never returns it. With `PERSIST_STATE=1`, Terraform plans a destroy-and-recreate of every `aws_iam_policy` with a description on every warm boot, because `description` is a `ForceNew` attribute in the AWS Terraform provider. Any `aws_iam_role_policy_attachment` referencing the policy is force-replaced in the same plan.

### MiniStack version

`ministackorg/ministack:1.3.11`

### Source references (against `main`)

- [`ministack/services/iam.py:304-323`](https://github.com/ministackorg/ministack/blob/main/ministack/services/iam.py#L304-L323) — `CreatePolicy` builds `_policies[arn]` without reading `Description`.
- [`ministack/services/iam.py:1455-1465`](https://github.com/ministackorg/ministack/blob/main/ministack/services/iam.py#L1455-L1465) — `_managed_policy_xml()` never emits a `<Description>` element.

### Buggy code

```python
# iam.py:304-323
_policies[arn] = {
    "PolicyName": name,
    "Arn": arn,
    "PolicyId": policy_id,
    "CreateDate": _now(),
    "UpdateDate": _now(),
    "DefaultVersionId": version_id,
    "AttachmentCount": 0,
    "IsAttachable": True,
    "Path": path,
    "Tags": [],
    "Versions": { ... },
}
```

### Reproducer

**Start MiniStack with persistence:**

```bash
docker run -d --name ministack-repro \
  -p 4566:4566 \
  -e PERSIST_STATE=1 \
  -v ministack_state_repro:/tmp/ministack-state \
  ministackorg/ministack:1.3.11
```

**`provider.tf`:**

```hcl
terraform {
  required_providers {
    aws = { source = "hashicorp/aws", version = "~> 6.0" }
  }
}

provider "aws" {
  region                      = "us-east-1"
  access_key                  = "test"
  secret_key                  = "test"
  skip_credentials_validation = true
  skip_metadata_api_check     = true
  skip_requesting_account_id  = true
  endpoints { iam = "http://localhost:4566" }
}
```

**`main.tf`:**

```hcl
resource "aws_iam_policy" "demo" {
  name        = "ministack-repro-demo-policy"
  description = "Repro description that should survive a warm boot"
[...]

PR #492, opened 2026-04-27
fix: wire 5 dropped state dicts into get_state/restore_state

### LINKED ISSUE #424: S3 bucket logging config not persisted — _bucket_logging_config missing from get_state()/restore_state()
## Summary

With `PERSIST_STATE=1`, S3 bucket logging configuration is **not** persisted across restarts, even though every other per-bucket dict (tags, versioning, CORS, lifecycle, policy, PAB, encryption, ACL, notifications, object lock, replication, websites) is. The root cause is localized: `ministack/services/s3.py` defines `_bucket_logging_config` at module scope and populates it via `_put_bucket_logging`, but `get_state()` / `restore_state()` never reference it — so the config is serialized to nothing on shutdown, and on warm boot `GetBucketLogging` returns an empty `<BucketLoggingStatus/>` as if the bucket had never been configured.

This survived the #422 fix because `_bucket_logging_config` is not in the list of `_bucket_*` dicts the fix enumerated.

## Version

- MiniStack: 1.3.8
- Image: `ministackorg/ministack:1.3.8`
- Host: Docker Desktop on macOS (Darwin 25.3.0)

## Reproduction

Fully self-contained — two files in an empty directory.

`docker-compose.yaml`:

```yaml
services:
  ministack:
    image: ministackorg/ministack:1.3.8
    ports:
      - "4566:4566"
    environment:
      - PERSIST_STATE=1
      - S3_PERSIST=1
    volumes:
      - ministack_state:/tmp/ministack-state
      - ministack_s3:/tmp/ministack-data
volumes:
  ministack_state:
  ministack_s3:
```

Steps:

```bash
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test \
       AWS_DEFAULT_REGION=us-east-1 AWS_ENDPOINT_URL=http://localhost:4566

# Cold start.
docker compose up -d && sleep 4

# Seed two buckets — one as the access-log target, one with logging enabled.
aws s3api create-bucket --bucket target-logs
aws s3api create-bucket --bucket src
aws s3api put-bucket-logging --bucket src \
  --bucket-logging-status '{"LoggingEnabled":{"TargetBucket":"target-logs","TargetPrefix":"log/"}}'

# Sanity — logging config is readable while the container is live.
aws s3api get-bucket-logging --bucket src
# { "LoggingEnabled": { "TargetBucket": "target-logs", "TargetPrefix": "log/" } }

# Graceful shutdown — triggers lifespan.shutdown and writes s3.json.
docker compose down --timeout 30

# Show the serialized state — bucket_logging / bucket_logging_config is absent.
docker run --rm -v "$(basename $PWD)_ministack_state:/d" python:3.12-alpine \
  python3 -c "import json; d=json.load(open('/d/s3.json')); \
              print('top-level keys:', list(d.keys()))"
# top-level keys: ['buckets_meta', 'bucket_versioning', 'bucket_notifications',
#                  'bucket_tags', 'bucket_policies', 'bucket_encryption',
#                  'bucket_lifecycle', 'bucket_cors', 'bucket_acl',
#                  'bucket_websites', 'bucket_object_lock', 'bucket_replication']
# (no 'bucket_logging' / 'bucket_logging_config')

# Warm boot.
docker compose up -d && sleep 4

# Logging config is gone.
aws s3api get-bucket-logging --bucket src
# {}   ← empty; LoggingEnabled no longer present
```

## Expected

After warm boot, `GetBucketLogging src` returns the same `LoggingEnabled` block that was set pre-shutdown.

## Actual

After warm boot, `GetBucketLogging src` returns `{}`. The access-log target bucket (`target-logs`) itself does restore — only the `src → target-logs/log/` linkage is lost.

## Root cause

`ministack/services/s3.py` holds logging state in `_bucket_logging_config` (declared ~line 70), populated by `_put_bucket_logging` (~line 1127) and read by `_get_bucket_logging` (~line 1114). But the persistence functions skip it:

```python
# ministack/services/s3.py

def get_state():
    return {
        "buckets_meta": copy.deepcopy(buckets_meta),
        "bucket_versioning": copy.deepcopy(_bucket_versioning),
        "bucket_notifications": copy.deepcopy(_bucket_notifications),
        "bucket_tags": copy.deepcopy(_bucket_tags),
        "bucket_policies": copy.deepcopy(_bucket_policies),
        "bucket_encryption": copy.deepcopy(_bucket_encryption),
        "bucket_lifecycle": copy.deepcopy(_bucket_lifecycle),
        "bucket_cors": copy.deepcopy(_bucket_cors),
        "bucket_acl": copy.deepcopy(_bucket_acl),
        "bucket_websites": copy.deepcopy(_bucket_websites),
        "bucket_object_lock": copy.deepcopy(_bucket_object_lock),
        "bucket_replication": copy.deepcopy(_bucket_replication),
        # ← _bucket_logging_config is missing here
    }


def restore_state(data):
    if data:
        # …
        _bucket_replication.update(data.get("bucket_replication", {}))
        # ← and here
```

## Suggested fix

Add `_bucket_logging_config` to both `get_state()` and `restore_state()`:

```python
def get_state():
    return {
        # …existing entries…
        "bucket_logging_config": copy.deepcopy(_bucket_logging_config),
    }

def restore_state(data):
    if data:
        # …existing entries…
        _bucket_logging_config.update(data.get("bucket_logging_config", {}))
```

No changes to `_put_bucket_logging` / `_get_bucket_logging` — those already work correctly in-session; only the persistence wiring is missing.

## Related

- #422 (S3 bucket metadata never persisted to `s3.json` on shutdown) — fixed in 1.3.8, enumerated the `_bucket_*` dicts that needed to be added to `get_state()`. This issue is the same class of miss: `_bucket_logging_config` was overlooked in that enumeration.

bognari:
## Summary

Four `AccountScopedDict` instances were mutated by API handlers but missing from `get_state()` / `restore_state()`. With `PERSIST_STATE=1`, every record stored via these APIs silently disappeared on the next restart — same bug family as recently closed #424 (S3 logging config) and #438 (IAM policy Description).

| Service | Dict | API |
|---|---|---|
| `secretsmanager` | `_resource_policies` | `PutResourcePolicy` / `DeleteResourcePolicy` |
| `kinesis` | `_consumers` | `RegisterStreamConsumer` (enhanced fan-out) |
| `ecs` | `_attributes` | `PutAttributes` / `DeleteAttributes` / `ListAttributes` |
| `sns` | `_platform_applications` | `CreatePlatformApplication` |
| `sns` | `_platform_endpoints` | `CreatePlatformEndpoint` |

## User-visible impact (Terraform / IaC)

- `aws_secretsmanager_secret_policy` lost on restart.
- `aws_kinesis_stream_consumer` breaks after restart with `ResourceNotFoundException`.
- ECS capacity-provider scheduling annotations lost.
- Mobile push (GCM / APNS) topology lost.

## What changed

| File | Change |
|---|---|
| `ministack/services/secretsmanager.py` | Adds `_resource_policies` to `get_state()` / `restore_state()`. |
| `ministack/services/kinesis.py` | Adds `_consumers` to both. |
| `ministack/services/ecs.py` | Adds `_attributes` to both. |
| `ministack/services/sns.py` | Adds `_platform_applications` and `_platform_endpoints` to both. |
| `tests/test_state_dict_persistence.py` *(new, 5 tests)* | Regression coverage. |

Each `get_state()` now copies the dict via `copy.deepcopy()` (preserving multi-tenant scoping) and `restore_state()` calls `.update()` on it. The `reset()` functions already cleared these dicts, confirming they hold real state.

Restore is gated by `PERSIST_STATE` — `load_state()` returns `None` immediately when the env var is unset.

## Test plan

- [x] `pytest tests/test_state_dict_persistence.py` — all 5 pass (all 5 fail without the fix).
- [x] `pytest tests/test_secretsmanager.py tests/test_kinesis.py tests/test_ecs.py tests/test_sns.py` — 135 pass; the 3 failures (`test_secretsmanager_list_include_planned_deletion`, `test_sns_http_subscription_*`) are pre-existing on `main` and unrelated to this PR.
- [ ] Reviewer check: warm-boot a real ministack with `PERSIST_STATE=1`, register a Kinesis consumer / put a secret resource policy / put an ECS attribute / create an SNS platform application, restart, verify they survive.

## Notes

- Independent of #491 (the architectural persistence fix). They can land in either order.
- All 4 dicts are pre-existing in the modules; this PR only adds them to the persistence path.

### DIFF
--- a/ministack/services/ecs.py
+++ b/ministack/services/ecs.py
@@ -108,6 +108,7 @@ def get_state():
         "tags": copy.deepcopy(_tags),
         "account_settings": copy.deepcopy(_account_settings),
         "capacity_providers": copy.deepcopy(_capacity_providers),
+        "attributes": copy.deepcopy(_attributes),
     }
     # Save tasks but strip Docker container IDs.
     # Iterate _data directly to capture ALL accounts.
@@ -131,6 +132,7 @@ def restore_state(data):
     _tags.update(data.get("tags", {}))
     _account_settings.update(data.get("account_settings", {}))
     _capacity_providers.update(data.get("capacity_providers", {}))
+    _attributes.update(data.get("attributes", {}))
     from ministack.core.responses import AccountScopedDict
     tasks_data = data.get("tasks", {})
     if isinstance(tasks_data, AccountScopedDict):
--- a/ministack/services/ecs.py
+++ b/ministack/services/ecs.py
@@ -53,6 +53,12 @@ _tasks = AccountScopedDict()
 _tags = AccountScopedDict()
 _account_settings = AccountScopedDict()
 _capacity_providers = AccountScopedDict()
+# `_attributes` was originally declared next to its handler block much
+# further down the file. Moved up here so the import-time `load_state`
+# block (which calls `restore_state` and references `_attributes`) sees
+# it defined; otherwise warm-boot fires NameError, the surrounding
+# try/except swallows it, and ALL ECS state silently fails to restore.
+_attributes = AccountScopedDict()
 
 _docker = None
 
@@ -1496,9 +1502,9 @@ def _delete_account_setting(data):
 
 # ---------------------------------------------------------------------------
 # Attributes (PutAttributes / DeleteAttributes / ListAttributes)
+# (state dict `_attributes` is declared at module top with the other
+# state — see the comment there for why.)
 # ---------------------------------------------------------------------------
-
-_attributes = AccountScopedDict()
 
 def _put_attributes(data):
     attrs = data.get("attributes", [])
--- a/ministack/services/kinesis.py
+++ b/ministack/services/kinesis.py
@@ -41,12 +41,16 @@ _sequence_lock = threading.Lock()
 # ── Persistence ────────────────────────────────────────────
 
 def get_state():
-    return {"streams": copy.deepcopy(_streams)}
+    return {
+        "streams": copy.deepcopy(_streams),
+        "consumers": copy.deepcopy(_consumers),
+    }
 
 
 def restore_state(data):
     if data:
         _streams.update(data.get("streams", {}))
+        _consumers.update(data.get("consumers", {}))
 
 
 try:
--- a/ministack/services/secretsmanager.py
+++ b/ministack/services/secretsmanager.py
@@ -42,12 +42,16 @@ _resource_policies = AccountScopedDict()
 # ── Persistence ────────────────────────────────────────────
 
 def get_state():
-    return {"secrets": copy.deepcopy(_secrets)}
+    return {
+        "secrets": copy.deepcopy(_secrets),
+        "resource_policies": copy.deepcopy(_resource_policies),
+    }
 
 
 def restore_state(data):
     if data:
         _secrets.update(data.get("secrets", {}))
+        _resource_policies.update(data.get("resource_policies", {}))
 
 
 try:
--- a/ministack/services/sns.py
+++ b/ministack/services/sns.py
@@ -58,13 +58,20 @@ _platform_endpoints = AccountScopedDict()
 # ── Persistence ────────────────────────────────────────────
 
 def get_state():
-    return {"topics": copy.deepcopy(_topics), "sub_arn_to_topic": copy.deepcopy(_sub_arn_to_topic)}
+    return {
+        "topics": copy.deepcopy(_topics),
+        "sub_arn_to_topic": copy.deepcopy(_sub_arn_to_topic),
+        "platform_applications": copy.deepcopy(_platform_applications),
+        "platform_endpoints": copy.deepcopy(_platform_endpoints),
+    }
 
 
 def restore_state(data):
     if data:
         _topics.update(data.get("topics", {}))
         _sub_arn_to_topic.update(data.get("sub_arn_to_topic", {}))
+        _platform_applications.update(data.get("platform_applications", {}))
+        _platform_endpoints.update(data.get("platform_endpoints", {}))
 
 
 try:
--- a/tests/test_state_dict_persistence.py
+++ b/tests/test_state_dict_persistence.py
@@ -23,16 +23,39 @@ import importlib
 
 import pytest
 
+from ministack.core import persistence
+
 
 def _module(mod_name):
     return importlib.import_module(f"ministack.services.{mod_name}")
 
 
-def _round_trip(mod):
-    """Snapshot → reset → restore via the module's public hooks."""
-    snapshot = mod.get_state()
+@pytest.fixture(autouse=True)
+def _enable_persistence(monkeypatch, tmp_path):
+    """Force PERSIST_STATE on and point STATE_DIR at a tmp dir for the
+    duration of each test so save_state / load_state actually write and
+    read JSON files instead of short-circuiting."""
+    monkeypatch.setattr(persistence, "PERSIST_STATE", True)
+    monkeypatch.setattr(persistence, "STATE_DIR", str(tmp_path))
+
+
+def _round_trip(mod, svc_key):
+    """Simulate a full warm-boot through the on-disk JSON path.
+
+    Going through `save_state` / `load_state` (rather than calling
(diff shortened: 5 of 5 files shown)


bognari:
The docstring fix landed via the GitHub UI suggestion (commit f566075) — thanks!

Re the CI failure: both `test_apigwv2_integration_uri_with_alias_qualifier_resolves_to_alias_target` and `test_apigwv2_integration_wrapped_alias_arn` time out under parallel load. This is the same `TimeoutError` pattern that already caused a sibling-set of `test_apigatewayv2.py::test_apigwv2_*` tests to be moved to `_SERIAL_TESTS` in `tests/conftest.py:53-58`. Identical failures appear in main-branch CI runs (e.g. run 24961078426) on tests that this PR doesn't touch — same observation as on #491. Treating it as an orthogonal pre-existing flakiness issue; happy to add these two to the serial allow-list as a follow-up if useful, but it feels out of scope for the per-service-state-dict-drops fix.

bognari:
Force-pushed (commit 198ca9d) addressing both review points and rebasing onto current main (which now has #491 merged).

**1. Real critical bug — confirmed and fixed.** `ecs._attributes` was declared at line 1499 but `restore_state()` at line 124 referenced it. The import-time `try: load_state("ecs")` block at line 148 fired BEFORE Python evaluated line 1499, so with `PERSIST_STATE=1` and a populated `ecs.json`, restore raised `NameError` → caught by the surrounding try/except → ALL ECS state silently lost on warm-boot.

Fix: hoisted `_attributes = AccountScopedDict()` up next to the other module-level state dicts. Removed the duplicate later declaration. Comment near both sites explains why.

**2. Tests now exercise the full disk path.** New `_round_trip(mod, svc_key)` writes via `persistence.save_state` to a tmp `STATE_DIR`, resets, then loads via `persistence.load_state`. Autouse `_enable_persistence` fixture forces `PERSIST_STATE=True` and points `STATE_DIR` at pytest's `tmp_path`. Same pattern used in #493.

Plus a dedicated **import-order regression** `test_ecs_module_reload_with_persisted_attributes_does_not_namerror` that calls `importlib.reload()` after writing `ecs.json` — exactly the failure mode the reviewer caught. Without the `_attributes` hoist, this test fails.

All 6 tests pass + 57 inherited from #491's test file (63 total in the persistence layer).

Branch is now rebased onto main (post-#491-merge); no conflicts.

Nahuel990 (approved):

bognari:
Pushed a generic NameError-at-import regression test (commit 121f399) — same bug pattern surfaced in 2 PRs in this series (ECS `_attributes` here, ACM `_synthetic_pem` in #494), so worth a structural guard that runs against EVERY persisted service.

`test_module_cold_import_with_typical_snapshot_does_not_log_restore_failure` parametrizes over every entry in `_state_map` (45 services today). For each:

1. Capture the module's current `get_state()` snapshot.
2. Persist via `save_state` to a tmp `STATE_DIR`.
3. **Cold-import** by removing the module from `sys.modules` and re-importing — `importlib.reload()` would NOT catch the bug because it merges new defs into the existing namespace, leaving any late-declared symbol bound from the prior import.
4. Capture log records during the cold import; assert no WARNING+ message about restore-failure was emitted.

Verified the test catches the bug by temporarily reverting the ECS hoist — the test fails with the actual NameError traceback in the captured log:

```
ERROR ministack.services.ecs:ecs.py:156 Failed to restore persisted state
Traceback ... NameError: name '_attributes' is not defined
```

**What it catches:** unconditional symbol references in `restore_state` (ECS-style — `_xxx.update(data.get(...))` where `_xxx` is declared too late).

**What it doesn't catch:** conditional references inside loops over restored data when the data is empty (ACM-style — `for cert in _certificates._data.values(): if "_pem_body" not in cert: cert["_pem_body"] = _synthetic_pem(...)`). For those, the per-service `test_restore_state_backfills_pem_body_for_pre_upgrade_snapshots` in #494 covers it explicitly. Future per-service additions can extend that pattern.

All 51 tests pass.

PR #493, opened 2026-04-27
fix: persist CloudWatch Logs destinations / metric filters / queries

bognari:
## Summary

Three module-level `AccountScopedDict`s in `cloudwatch_logs.py` were mutated by public APIs but missing from `get_state()` / `restore_state()`. With `PERSIST_STATE=1`, every record stored via these APIs disappeared on restart — same bug family as #491 / #492 / #424 / #438.

| Dict | API that writes it | Real-world symptom |
|---|---|---|
| `_destinations` | `PutDestination` | Terraform `aws_cloudwatch_log_destination` vanishes |
| `_metric_filters` | `PutMetricFilter` | Terraform `aws_cloudwatch_log_metric_filter` vanishes |
| `_queries` | `StartQuery` | CloudWatch Logs Insights queries lose history |

## Plus a follow-on consistency bug

Subscription filters live inside `_log_groups[group]["subscriptionFilters"]` (already persisted) and reference `destinationArn` values that resolve through `_destinations` (was *not* persisted). After warm-boot you got split-brain: the subscription filter still pointed at a destination that no longer existed in `_destinations`. Persisting `_destinations` resolves this.

## What changed

| File | Change |
|---|---|
| `ministack/services/cloudwatch_logs.py` | +3 lines in `get_state()`, +3 in `restore_state()`. |
| `tests/test_cloudwatch_logs_persistence.py` *(new, 4 tests)* | One round-trip per dict + the subscription-filter ↔ destination consistency test. |

`reset()` already cleared all three dicts, confirming they hold real state — the omission was an oversight in the original wiring.

## Tuple-key note (`_metric_filters`)

`_metric_filters` is keyed by `(log_group_name, filter_name)` tuples. JSON cannot natively serialise non-string keys, but `AccountScopedDict`'s encoder hook in `core/persistence.py` repr-encodes the original key on save and `ast.literal_eval`-decodes on restore — so the tuple round-trips correctly. The new `test_metric_filters_survive_warm_boot` covers this path.

Restore is gated by `PERSIST_STATE` — `load_state()` returns `None` immediately when the env var is unset.

## Test plan

- [x] `pytest tests/test_cloudwatch_logs_persistence.py` — all 4 pass (all 4 fail without the fix).
- [x] `pytest tests/test_logs.py` — 33 pass; the 9 failures (`test_logs_*delivery*`) are pre-existing on `main` (`InvalidOperationException: Unknown action: PutDeliverySource`) and unrelated to this PR.
- [ ] Reviewer check: warm-boot a real ministack with `PERSIST_STATE=1`, create a destination / metric filter / query, restart, verify they survive.

## Notes

Independent of #491 (persistence architectural gap) and #492 (per-service state-dict drops). Can land in any order.

### DIFF
--- a/ministack/services/cloudwatch_logs.py
+++ b/ministack/services/cloudwatch_logs.py
@@ -70,6 +70,9 @@ _deliveries = AccountScopedDict()
 def get_state():
     return {
         "log_groups": copy.deepcopy(_log_groups),
+        "destinations": copy.deepcopy(_destinations),
+        "metric_filters": copy.deepcopy(_metric_filters),
+        "queries": copy.deepcopy(_queries),
         "delivery_sources": copy.deepcopy(_delivery_sources),
         "delivery_destinations": copy.deepcopy(_delivery_destinations),
         "deliveries": copy.deepcopy(_deliveries),
@@ -79,6 +82,9 @@ def get_state():
 def restore_state(data):
     if data:
         _log_groups.update(data.get("log_groups", {}))
+        _destinations.update(data.get("destinations", {}))
+        _metric_filters.update(data.get("metric_filters", {}))
+        _queries.update(data.get("queries", {}))
         _delivery_sources.update(data.get("delivery_sources", {}))
         _delivery_destinations.update(data.get("delivery_destinations", {}))
         _deliveries.update(data.get("deliveries", {}))
--- a/tests/test_cloudwatch_logs_persistence.py
+++ b/tests/test_cloudwatch_logs_persistence.py
@@ -16,20 +16,52 @@ Plus a follow-on consistency bug:
        still references a destination that no longer exists in
        _destinations.
 
-Each test exercises the public get_state / restore_state contract
-without going through the live HTTP server.
+Each test exercises the FULL warm-boot path:
+  1. populate the in-memory dict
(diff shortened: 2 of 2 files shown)


bognari:
Pushed test improvement (commit 062ede1) addressing the review point.

`_round_trip()` now goes through the full warm-boot disk path:

  1. populate the in-memory dict
  2. `get_state()` snapshot
  3. `persistence.save_state()` → JSON-encode to disk
  4. `mod.reset()` (simulate process restart)
  5. `persistence.load_state()` → JSON-decode from disk
  6. `restore_state(loaded)`

A new autouse fixture monkeypatches `persistence.PERSIST_STATE=True` and points `STATE_DIR` at pytest's `tmp_path` so the save/load calls actually write and read JSON files instead of short-circuiting.

This means the tuple-key path for `_metric_filters` (repr → JSON string → `ast.literal_eval` round-trip via `core/persistence.py::_json_default` / `_json_object_hook`) is now genuinely exercised by the test suite. A regression in those encoder/decoder hooks would now fail `test_metric_filters_survive_warm_boot`.

All 4 tests still pass. Switching back to PR 4 to continue the chain.

Nahuel990 (approved):

PR #494, opened 2026-04-27
fix: ACM cert body / chain fidelity + private-key disk-leak scrub

bognari:
## Summary

Two fidelity bugs in `services/acm.py`, plus a persistence side-channel surfaced by the fix.

| Bug | Symptom |
|---|---|
| `GetCertificate` returned hard-coded literal `MIIFakeCertificateDataHere` regardless of stored data | Any consumer that parses the PEM (mTLS validators, ALB attachment, X.509 validators) fails |
| `ImportCertificate` discarded the request's `Certificate` / `CertificateChain` / `PrivateKey` and hard-coded `DomainName='imported.example.com'` | Imported certs were unrecoverable; advertised a fabricated domain unrelated to the actual cert |
| **Side-channel** the fix uncovered: with `PERSIST_STATE=1`, `_private_key` would be JSON-written plaintext to `${STATE_DIR}/acm.json` | Wire path refuses to leak the key, but persistence silently did |

## What changed

- `_import_certificate` now stores `Certificate` / `CertificateChain` / `PrivateKey` from the request body. Re-import (when `CertificateArn` is supplied) preserves `CreatedAt` / `Tags` / `Options` / `SubjectAlternativeNames` / `DomainName` (matches real AWS cert-renewal semantics).
- `_get_certificate` returns the stored cert + chain. `PrivateKey` is intentionally never returned — real AWS only exposes it via passphrase-protected `ExportCertificate`, which this emulator does not implement.
- `_request_certificate` now seeds a clearly-synthetic placeholder PEM via new `_synthetic_pem()` so the unified `GetCertificate` path works for `AMAZON_ISSUED` certs too.
- New helper `_decode_pem_field` handles both base64-over-wire (boto3 / AWS-JSON shape) and raw-PEM strings.
- `DomainName` for `IMPORTED` certs is now `imported-cert-{8hex}.invalid` (RFC 2606 reserved TLD) — clearly synthetic, doesn't claim coverage of any real domain. ASN.1 / X.509 parsing remains out of scope.
- `get_state()` now scrubs `_private_key` before returning the snapshot. Without this, `save_state()` would write plaintext key material to `${STATE_DIR}/acm.json`. Cert body and chain still round-trip; only `PrivateKey` is lost — re-import is required after restart for `IMPORTED` certs that need the key.

- `RequestCertificate` flow still returns a `BEGIN CERTIFICATE`-prefixed body (existing `test_acm_get_certificate` substring check still passes).
- `DescribeCertificate` / `ListCertificates` schemas unchanged (PEM bodies live in private `_pem_*` keys, not exposed via `_cert_shape`).
- `PERSIST_STATE` gating is unchanged — `load_state()` returns `None` when the env var is unset.

## Test plan

- [x] `pytest tests/test_acm_cert_body.py tests/test_acm.py` — all 16 pass (6 new + 10 existing).
- [x] Without the fix: 4 of 6 new tests fail (round-trip + domain).
- [x] Without the persistence-scrub fix: `test_get_state_strips_private_key_from_persisted_snapshot` fails — verified by direct `get_state()` invocation in a Python REPL.
- [ ] Reviewer check: end-to-end with `PERSIST_STATE=1` set: `import_certificate` → restart → `get_certificate` returns the imported PEM (no PrivateKey).

## Notes

Independent of #491 / #492 / #493. Can land in any order.

### DIFF
--- a/ministack/services/acm.py
+++ b/ministack/services/acm.py
@@ -25,7 +25,20 @@ _certificates = AccountScopedDict()  # arn -> certificate dict
 
 
 def get_state():
-    return copy.deepcopy({"_certificates": _certificates})
+    # Strip _private_key before persisting — real AWS only exposes the
+    # private key via passphrase-protected ExportCertificate, and the
+    # GetCertificate path already honours that. Writing it plaintext to
+    # ${STATE_DIR}/acm.json would turn warm-boot persistence into a
+    # side-channel for material that the wire protocol refuses to leak.
+    # The cert body and chain still round-trip; only PrivateKey is lost,
+    # which means a re-import is required after restart for IMPORTED
+    # certs that need the key.
+    scrubbed = AccountScopedDict()
+    for arn, cert in _certificates.items():
+        c = copy.deepcopy(cert)
+        c.pop("_private_key", None)
+        scrubbed[arn] = c
+    return {"_certificates": scrubbed}
 
 
 def restore_state(data):
@@ -124,6 +137,19 @@ async def handle_request(method, path, headers, body, query_params):
     return handler(data)
 
 
+def _synthetic_pem(domain):
+    """A clearly-synthetic but well-formed-looking PEM placeholder for
+    RequestCertificate-issued certs. The emulator does not generate real
+    X.509, so anything that actually parses the cert will fail — but
+    consumers that only do a 'BEGIN CERTIFICATE' substring check (the
+    common Terraform / CDK pattern) work."""
+    return (
+        f"-----BEGIN CERTIFICATE-----\n"
+        f"MIIBsynth-emulator-cert-for-{domain[:40]}\n"
+        f"-----END CERTIFICATE-----\n"
+    )
+
+
 def _request_certificate(data):
     domain = data.get("DomainName", "")
     if not domain:
@@ -148,6 +174,9 @@ def _request_certificate(data):
         "ValidationMethod": method,
         "Tags": data.get("Tags", []),
         "Options": {},
+        "_pem_body": _synthetic_pem(domain),
+        "_pem_chain": "",
+        "_private_key": "",
     }
     logger.info("RequestCertificate: %s -> %s", domain, arn)
     return json_response({"CertificateArn": arn})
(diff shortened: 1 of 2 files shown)


bognari:
Pushed review fixes (commit 45dfe56) addressing all three findings.

**1. Multi-tenancy break in get_state() — real bug, now fixed.** Was iterating `_certificates.items()` which is request-scoped via AccountScopedDict's contextvar; would silently drop every other tenant's certificates from the snapshot. Now iterates `_certificates._data` directly to capture every `(account_id, key)` pair. Same bug class as the `dict()` cast issue in scheduler/backup that #491 caught.

New regression test: `test_get_state_preserves_certs_across_all_tenants` sets up two accounts via `_request_account_id`, takes a snapshot from one, and asserts both tenants' certs survive.

**2. `_synthetic_pem()` body had invalid base64 — fixed.** The placeholder embedded the domain string (containing `-` / `.`) directly into the PEM body line, breaking any consumer that pre-decodes (PyOpenSSL, cryptography) before ASN.1 parsing. Now uses a fixed valid base64 payload (`AQIDBAUGBwgJCgsMDQ4PEA==`); the domain remains in `DomainName` / `SubjectAlternativeNames` metadata where it belongs.

New regression test: `test_synthetic_pem_body_is_valid_base64` extracts the body between BEGIN/END markers and asserts `base64.b64decode()` succeeds.

**3. Test docstrings were no-op string literals — fixed.** All 5 affected tests had `acm = acm_client` before the triple-quoted string, making it a meaningless expression statement. Moved each docstring to be the first statement of its function.

All 18 ACM tests pass (8 new + 10 existing).

bognari:
Force-pushed (commit 749adc3) addressing both review points and rebasing onto main (which now has #491 merged).

**1. Real edge case — backfill `_pem_body` on warm-boot.** Pre-fix `acm.json` snapshots have no `_pem_body` / `_pem_chain` keys. Without backfill, those certs would return an empty Certificate field after warm-boot — strictly worse than the old behaviour. `restore_state()` now backfills with `_synthetic_pem(cert["DomainName"])`, so consumers that substring-check 'BEGIN CERTIFICATE' (Terraform / CDK) keep working.

Side-fix uncovered while doing this: `_synthetic_pem` was defined further down the file than `restore_state` — same NameError-on-import-time pattern as the ECS bug caught in #492. Hoisted it up; removed the duplicate definition.

New regression test: `test_restore_state_backfills_pem_body_for_pre_upgrade_snapshots`.

**2. Test correctness — all-tenant private-key scrub verification.** My `json.dumps(snapshot, default=str)` defensive check used `AccountScopedDict.__repr__` which is request-scoped — so it only saw one tenant's data, and a secret string from another tenant could have leaked unnoticed.

Now uses `core/persistence._json_default` (the actual production encoder) and sets up two tenants via `_request_account_id.set(...)`. Asserts both `_private_key` keys are stripped AND that neither secret marker appears in the JSON blob.

Branch is now rebased onto main; no conflicts. All 19 ACM tests pass (9 in this file + 10 existing).

Nahuel990 (approved):

Issue #439, opened 2026-04-24
aws_apigatewayv2_integration.ContentHandlingStrategy not persisted across warm-boot


whittin3:
### Summary

`CreateIntegration` for an HTTP/WebSocket API accepts `contentHandlingStrategy`, but MiniStack never stores it on the integration and `GetIntegration` never returns it. Terraform plans an in-place update adding the field back on every `apply`, and the field is silently missing at runtime (so requests that rely on `CONVERT_TO_TEXT` / `CONVERT_TO_BINARY` payload translation don't get it).

### MiniStack version

`ministackorg/ministack:1.3.11`

### Source references (against `main`)

- [`ministack/services/apigateway.py:717-735`](https://github.com/ministackorg/ministack/blob/main/ministack/services/apigateway.py#L717-L735) — integration dict construction, `contentHandlingStrategy` absent.
- [`ministack/services/apigateway.py:753-757`](https://github.com/ministackorg/ministack/blob/main/ministack/services/apigateway.py#L753-L757) — `UpdateIntegration` allowlist, `contentHandlingStrategy` absent.

### Buggy code

```python
# apigateway.py:717-735
integration = {
    "integrationId": int_id,
    "integrationType": data.get("integrationType", "AWS_PROXY"),
    "integrationUri": data.get("integrationUri", ""),
    "integrationMethod": data.get("integrationMethod", "POST"),
    "payloadFormatVersion": data.get("payloadFormatVersion", "2.0"),
    "timeoutInMillis": data.get("timeoutInMillis", 30000),
    "connectionType": data.get("connectionType", "INTERNET"),
    "description": data.get("description", ""),
    "requestParameters": data.get("requestParameters", {}),
    "requestTemplates": data.get("requestTemplates", {}),
    "responseParameters": data.get("responseParameters", {}),
}
```

### Reproducer

**Start MiniStack with persistence:**

```bash
docker run -d --name ministack-repro \
  -p 4566:4566 \
  -e PERSIST_STATE=1 \
  -v ministack_state_repro:/tmp/ministack-state \
  ministackorg/ministack:1.3.11
```

**`provider.tf`:**

```hcl
terraform {
  required_providers {
    aws     = { source = "hashicorp/aws",     version = "~> 6.0" }
    archive = { source = "hashicorp/archive", version = "~> 2.7" }
  }
}

provider "aws" {
  region                      = "us-east-1"
  access_key                  = "test"
  secret_key                  = "test"
[...]

PR #495, opened 2026-04-27
fix: persist contentHandling on API Gateway v1 PutIntegration (#439 backport)

bognari:
## Summary

Backports the API Gateway v2 `contentHandling` fidelity fix from #439 to API Gateway v1 (REST API). The field was accepted by `PutIntegration` but never stored, so `GetIntegration` always returned `None` and Terraform's `aws_api_gateway_integration.content_handling` planned a perpetual in-place replace on every apply.

| Bug | Status |
|---|---|
| `PutIntegration` silently dropped `contentHandling` | **fixed** by this PR |
| `PutIntegrationResponse` historically dropped `contentHandling` | **already correct** (was added in commit `0ef45048`); regression-locked by new tests |

## What changed

| File | Change |
|---|---|
| `ministack/services/apigateway_v1.py` | +1 functional line in `_put_integration` (`"contentHandling": data.get("contentHandling")`) plus a 4-line comment explaining the AWS-spec context and the v2 cross-reference. |
| `tests/test_apigatewayv1_content_handling.py` *(new, 6 tests)* | Parametrised round-trip for both `CONVERT_TO_TEXT` / `CONVERT_TO_BINARY`, omit-default behaviour, JSON-Patch `replace /contentHandling` via UpdateIntegration, plus regression-lock for `PutIntegrationResponse`. |

`_apply_patch` is path-agnostic (no per-field whitelist), so the single-site fix is structurally complete for v1 — the v2 fix needed two sites because v2 has an explicit allow-list in `_update_integration`.

When `contentHandling` is omitted, the value is stored as `None` (matching the long-standing convention for adjacent fields like `credentials`). boto3 strips null-valued fields client-side per the service model, so wire-format consumers see no change.

## Test plan

- [x] `pytest tests/test_apigatewayv1_content_handling.py` — all 6 pass (2 of them fail without the fix).
- [x] `pytest tests/test_apigatewayv1.py` — all 54 existing v1 tests still pass.
- [ ] Reviewer check: a Terraform `aws_api_gateway_integration` with `content_handling = "CONVERT_TO_TEXT"` should now plan as no-op on the second `terraform apply`, instead of perpetually replacing.

## Notes

Independent of #491 / #492 / #493 / #494. Can land in any order.

### DIFF
--- a/ministack/services/apigateway_v1.py
+++ b/ministack/services/apigateway_v1.py
@@ -1242,6 +1242,11 @@ def _put_integration(api_id, resource_id, http_method, data):
         "timeoutInMillis": data.get("timeoutInMillis", 29000),
         "cacheNamespace": resource_id,
         "cacheKeyParameters": data.get("cacheKeyParameters", []),
+        # contentHandling (CONVERT_TO_TEXT | CONVERT_TO_BINARY) is the v1
+        # equivalent of v2's contentHandlingStrategy (#439). Without
+        # storing it Terraform's aws_api_gateway_integration plans a
+        # perpetual replace on every apply.
+        "contentHandling": data.get("contentHandling"),
         "integrationResponses": {},
     }
     method_obj["methodIntegration"] = integration
--- a/tests/test_apigatewayv1_content_handling.py
+++ b/tests/test_apigatewayv1_content_handling.py
@@ -24,9 +24,12 @@ Uses the session-scoped `apigw_v1` fixture from tests/conftest.py.
 import pytest
 
 
-def _create_method_setup(apigw_v1):
+@pytest.fixture
+def method_setup(apigw_v1):
     """Create a fresh REST API + resource + method as a foundation for
-    integration tests. Returns (api_id, resource_id, http_method)."""
+    integration tests. Yields (api_id, resource_id, http_method) and
+    deletes the REST API in teardown so the session-scoped client
+    doesn't leak state across tests."""
     api = apigw_v1.create_rest_api(name="ch-test-api")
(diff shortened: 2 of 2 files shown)


bognari:
Pushed test cleanup (commit 83d5a8b) addressing the review point.

Converted `_create_method_setup` from a plain helper into a `method_setup` pytest fixture that `yield`s the `(api_id, resource_id, http_method)` tuple and calls `delete_rest_api` in teardown (wrapped in try/except so a failed test doesn't poison the cleanup path). All 5 tests now take `method_setup` as a parameter.

Since the session-scoped `apigw_v1` client only resets ministack state once per session, the previous helper would leak a REST API per test invocation — would have slowed down the suite and risked cross-test flakiness over time.

Also corrected an off-by-5 line reference in one test docstring (1306 → 1311) that the prior review noted as cosmetic.

All 60 v1 tests pass (6 in this file + 54 existing).

bognari:
Pushed doc fixes (commit 174150c) addressing all three review points + rebased onto main.

1. **Module docstring line-number** (`apigateway_v1.py:1306`) → replaced with function name `_put_integration_response`.
2. **`test_update_integration_can_patch_content_handling` docstring** said "the field must exist on the integration so the patch applies cleanly" — misleading, because `_apply_patch` actually creates missing keys for `replace`/`add` ops. Reworded to focus on the round-trip behaviour.
3. **`test_put_integration_response_persists_content_handling` docstring** — same line-number issue (1311), same fix.

No functional changes. All 6 tests still pass.

Nahuel990 (approved):

PR #497, opened 2026-04-27
fix: persist Cognito hosted-UI / federation _auth_codes across warm-boot

bognari:
## Summary

`_auth_codes` (hosted-UI / SAML-OIDC federation relay flow) was declared with the comment "ephemeral, not persisted" but in fact has a 5-minute TTL. Any in-flight Cognito hosted-UI sign-in that straddled a warm-boot was silently invalidated even though the code itself was still well within its TTL — forcing a needless re-auth.

| Symbol | Pre-fix | Post-fix |
|---|---|---|
| `_auth_codes` (hosted-UI / federation) | Plain dict, NOT in `get_state` / `restore_state` | Plain dict, IN both |
| `_authorization_codes` (managed-login PKCE) | Plain dict, already IN both | Unchanged (regression-locked by new test) |

## Why these dicts are intentionally PLAIN dicts (not AccountScopedDict)

The OAuth2 token endpoint (`cognito.py:_oauth2_token`) is a public callback authenticated via HTTP Basic (`client_id` / `client_secret`), with **no SigV4 access key**. Wrapping these dicts in `AccountScopedDict` would make every callback lookup happen under the default account (`MINISTACK_ACCOUNT_ID` / `"000000000000"`), invisible to codes minted under any other tenant — breaking the entire OAuth2 flow.

Effective tenant isolation is provided by the `secrets.token_urlsafe(32)` (256-bit unguessable) token namespace. The new code adds a multi-line section comment explaining this and a `test_auth_codes_dict_types_not_account_scoped` regression test pins the choice so a future "well-meaning" `AccountScopedDict` refactor can't silently break OAuth2.

## What changed

| File | Change |
|---|---|
| `ministack/services/cognito.py` | +1 line in `get_state()`, +1 in `restore_state()`, plus a 9-line section-header comment block replacing the misleading "ephemeral, not persisted" one-liner. |
| `tests/test_cognito_auth_codes_persistence.py` *(new, 2 tests)* | Warm-boot persistence round-trip (fails before the fix) + a type-pin regression guard. |

Restore is gated by `PERSIST_STATE` — `load_state()` returns `None` when the env var is unset.

## Stale-code safety

`_cleanup_expired_relay_codes()` (line 493) runs from the OAuth2 token endpoint before any code lookup. Past-TTL codes restored from disk are evicted on the next token-exchange attempt; they cannot be successfully redeemed. No replay-risk surface.

## Test plan

- [x] `pytest tests/test_cognito_auth_codes_persistence.py` — both pass after the fix (the persistence test fails without it).
- [x] `pytest tests/test_cognito.py` — all 94 existing cognito tests still pass.
- [ ] Reviewer check: end-to-end with `PERSIST_STATE=1`: initiate hosted-UI flow → restart ministack within 5 min → completing the flow with the same code should succeed.

## Notes

Independent of #491 / #492 / #493 / #494 / #495.

### DIFF
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -200,8 +200,17 @@ _identity_pools = AccountScopedDict()
 _identity_tags = AccountScopedDict()   # identity_pool_id -> {key: value}
 
 # ---------------------------------------------------------------------------
-# In-memory state — OAuth2 authorization codes (ephemeral, not persisted)
+# In-memory state — OAuth2 hosted-UI / federation codes
 # ---------------------------------------------------------------------------
+# `_auth_codes` and `_authorization_codes` are intentionally plain dicts
+# (NOT AccountScopedDict). The OAuth2 token endpoint is a public callback
+# with no AWS authentication context — it identifies the caller via the
+# random code + client_id + client_secret, not via a SigV4 access key.
+# Wrapping these in AccountScopedDict would make the callback lookup
+# happen under a default account, invisible to codes issued under any
+# other tenant. Effective tenant isolation is provided by the random
+# unguessable token namespace. See
+# tests/test_cognito_auth_codes_persistence.py for a regression pin.
 
 _auth_codes = {}   # code -> {pool_id, client_id, username, redirect_uri, scopes, state, created_at}
 _AUTH_CODE_TTL = 300  # 5 minutes
@@ -217,6 +226,7 @@ def get_state():
         "identity_tags": copy.deepcopy(_identity_tags),
         "authorization_codes": copy.deepcopy(_authorization_codes),
         "refresh_tokens": copy.deepcopy(_refresh_tokens),
+        "auth_codes": copy.deepcopy(_auth_codes),
     }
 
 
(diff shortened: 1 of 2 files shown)


bognari:
Pushed both fixes (commit d69cab2) + rebased onto main.

**1. Schema mismatch in test payload — fixed.** The test was writing `scopes` as a list and adding a `state` field, but production `type: "code"` entries in `_auth_codes` use `scopes` as a space-separated string (copied from `relay["scope"]`) and have no `state` field — that lives on the sibling `type: "relay"` entry. Aligned the test payload with the actual production schema so the regression test won't drift from real-world behaviour.

**2. Tighter dict-type pin — fixed.** Switched `not isinstance(x, AccountScopedDict)` to `type(x) is dict`. Strict identity catches not just AccountScopedDict but any other Mapping subclass that might slip in.

**Re the CI failure**: same pre-existing apigwv2 TimeoutError pattern (this run hit `test_apigwv2_cors_preflight_denies_non_allowlisted_origin` instead of the alias-resolution tests, but it's the same root cause — sibling-set is already in `_SERIAL_TESTS`). Not related to anything in this PR; same observation noted on #491 / #492 / #494.

ministack/services/cognito.py
```diff
@@ -200,8 +200,17 @@ def well_known_openid_configuration(pool_id: str, region: str | None = None):
 _identity_tags = AccountScopedDict()   # identity_pool_id -> {key: value}
 
 # ---------------------------------------------------------------------------
-# In-memory state — OAuth2 authorization codes (ephemeral, not persisted)
```
Nahuel990, line 203: Please be careful when adding comments, as they may mislead contributors.

Nahuel990 (approved):

PR #498, opened 2026-04-27
fix: eagerly import non-routable persisted services on startup

bognari:
## Summary

Follow-up to #491 addressing a second-round Copilot review point that landed after the merge.

Services that participate in `_state_map` but are NOT in `SERVICE_REGISTRY` have no AWS endpoint, so the lazy router never imports them on process startup. Their module-level `load_state()` block therefore never fires until something else triggers an import — silently dropping the persisted snapshot until the first service-specific request arrives.

Two such services exist today:

| Module | How it's normally reached | Restore impact pre-fix |
|---|---|---|
| `pipes` | CloudFormation-internal, imported only when a CFN `AWS::Pipes::Pipe` resource is provisioned | RUNNING pipes silently stop forwarding events after warm-boot until a new pipe is registered |
| `ses_v2` | `/v2/email/*` path-prefix shortcut in `app.py` | v2 SES identities / templates / config-sets evaporate after warm-boot until the first v2 request arrives |

## What changed

| File | Change |
|---|---|
| `ministack/app.py` | `_load_persisted_state()` now eagerly imports both modules. Each entry is a tiny cold-start cost; comment explains the criterion (in `_state_map`, not in `SERVICE_REGISTRY`). |
| `tests/test_persistence_symmetry.py` | New `test_state_map_services_without_endpoint_are_eagerly_imported` derives the set `_state_map - SERVICE_REGISTRY` and asserts each entry is referenced by name in `_load_persisted_state()`'s source. The next time someone adds a non-routable service to `_state_map`, this test fails until they wire it up. |

For pipes this combines with the `restore_state()` poller-start that landed in #491, so persisted event forwarding actually resumes after warm-boot — not just the in-memory definition.

## Test plan

- [x] `pytest tests/test_persistence_symmetry.py` — all 58 pass.
- [x] The new structural test caught `ses_v2` as a second instance of the same bug class while writing the fix for `pipes` — exactly the regression behavior the test is designed to provide.
- [ ] Reviewer check: with `PERSIST_STATE=1`, persist a SES v2 identity, restart ministack, immediately call `list_email_identities` (without any prior `/v2/email/*` request) — should return the identity.

## Notes

This is a follow-up to merged #491. Original review thread on `pipes` was addressed before the merge; the broader generalization to all non-routable persisted services arrived afterwards.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -1203,6 +1203,20 @@ def _load_persisted_state():
             _get_module(svc_key).load_persisted_state(data)
             logger.info("Loaded persisted state for %s", svc_key)
 
+    # Eagerly import services that participate in `_state_map` but are
+    # not directly in `SERVICE_REGISTRY` (or are reached only by a
+    # path-prefix shortcut, like `ses_v2` via `/v2/email/*` and `pipes`
+    # via CloudFormation). The lazy router won't import these on
+    # startup, so their module-level `load_state()` block never fires
+    # until something else triggers an import — silently dropping the
+    # persisted snapshot. Importing here triggers the restore (and, for
+    # pipes, also restarts the background poller for any RUNNING pipe).
+    # Keep this list narrow — every entry costs a cold-start import.
+    # Enforced by `tests/test_persistence_symmetry.py::test_state_map_
+    # services_without_endpoint_are_eagerly_imported`.
+    for svc_key in ("pipes", "ses_v2"):
+        _get_module(svc_key)
+
 
 async def _wait_for_port(port, timeout=30):
     """Wait until the server is accepting TCP connections."""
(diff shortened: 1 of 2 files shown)


bognari:
Pushed comment-only fix (commit fe14876) addressing the review point. Reworded the eager-import block's comment to spell out the actual criterion (persisted services whose restore depends on import-time `load_state()` AND that aren't otherwise imported during startup) instead of the looser "_state_map - SERVICE_REGISTRY" wording. Cites the two current examples (ses_v2 via path-prefix shortcut, pipes via CloudFormation) so the rationale is concrete.

Nahuel990 (approved):

Nahuel990 (approved):

PR #500, opened 2026-04-27
fix: 3 verified MEDIUM/LOW correctness bugs (M-1, M-2, L-2)

bognari:
## Summary

Three independent correctness fixes from §7.4 / §7.5 of the gap audit, each verified against real AWS behaviour first (per the lesson from the rejected KMS audit claims).

| ID | File | Bug | Fix |
|---|---|---|---|
| **M-1** | `services/apigateway.py` + `apigateway_v1.py` | `get_state()` returned LIVE `AccountScopedDict` references rather than deep copies. A concurrent write during shutdown serialisation could corrupt the persisted JSON. The apigateway pair was the only outlier — all other persisted services already deep-copy. | Wrap each dict in `copy.deepcopy(...)`. |
| **M-2** | `services/secretsmanager.py` | `_delete_secret(force=True)` removed the secret but left the matching entry in `_resource_policies[arn]`. Invisible to APIs but accumulating in memory and surviving warm-boot via the persistence path. | Add `_resource_policies.pop(arn, None)` after `del _secrets[key]`. |
| **L-2** | `services/acm.py` | `_list_certificates` returned `{"NextToken": null}` unconditionally. boto3 strips null fields client-side, but other SDKs (Java, Go, raw HTTP) and pagination loops checking `if "NextToken" in response` see the literal null and loop forever. | Drop the `NextToken` key from the response (single-page emulator). |

## What changed

| File | Change |
|---|---|
| `ministack/services/apigateway.py` | `get_state()`: wrap 9 dicts in `copy.deepcopy(...)`. |
| `ministack/services/apigateway_v1.py` | Same shape: wrap 13 dicts. |
| `ministack/services/secretsmanager.py` | +1 `pop` line in force-delete path. |
| `ministack/services/acm.py` | Drop `NextToken` from response when no next page. |
| `tests/test_misc_medium_low_fixes.py` *(new, 4 tests)* | M-1 parametrised over both apigateway modules; M-2 in-process via the module's `handle_request`; L-2 wire-level via raw HTTP to bypass boto3's null-stripping. |

### Test design notes

- **L-2 test bypasses boto3** because boto3 strips null fields client-side, hiding the bug. Uses `urllib.request.urlopen` to assert the wire-level JSON response.
- **M-2 test bypasses boto3 + HTTP** because orphan resource policies are by definition invisible from the API surface. Uses an in-process helper that drives `mod.handle_request(...)` synchronously via `asyncio.run`, then inspects both `_secrets` and `_resource_policies` directly.

## Skipped from §7.4 / §7.5 with rationale

- **M-3** SNS `Signature: "FAKE"` — needs real RSA signing infrastructure; cosmetic for test usage. Track separately.
- **M-5** KMS `DescribeKey` shape — verified non-bug in the PR-6 follow-up: real AWS `DescribeKey` doesn't return `Tags` or `KeyRotationEnabled` either.
- **L-1** secretsmanager filter substring vs prefix — AWS docs don't clearly specify the matching semantics; defer until verifiable.
- **M-8 / M-11** Lambda alias migration — needs separate per-service backfill PR.

## Test plan

- [x] `pytest tests/test_misc_medium_low_fixes.py` — all 4 pass (all 4 fail without the fix).
- [x] `pytest tests/test_acm.py tests/test_secretsmanager.py tests/test_apigatewayv1.py tests/test_apigatewayv2.py` — 174 existing tests still pass.
- [ ] Reviewer check: SDK pagination test (e.g. Java SDK against ministack ACM `ListCertificates`) — should no longer loop forever.

### DIFF
--- a/ministack/services/acm.py
+++ b/ministack/services/acm.py
@@ -172,7 +172,13 @@ def _list_certificates(data):
             "DomainName": cert["DomainName"],
             "Status": cert["Status"],
         })
-    return json_response({"CertificateSummaryList": summaries, "NextToken": None})
+    # Real AWS omits NextToken when there's no next page. boto3 strips
+    # null fields client-side so it tolerates `"NextToken": null`, but
+    # other SDKs (Java, Go, raw HTTP) and pagination loops checking
+    # `if "NextToken" in response` see the literal null and loop
+    # forever. ACM emulator currently emits a single page, so always
+    # omit the key.
+    return json_response({"CertificateSummaryList": summaries})
 
 
 def _delete_certificate(data):
--- a/ministack/services/apigateway.py
+++ b/ministack/services/apigateway.py
@@ -119,17 +119,24 @@ def _extract_lambda_ref_from_integration_uri(uri: str) -> str:
 # ---- Persistence hooks ----
 
 def get_state() -> dict:
-    """Return full module state for persistence."""
+    """Return full module state for persistence.
+
+    Deep-copies each dict so a concurrent write during shutdown
+    serialisation can't corrupt the persisted JSON. Every other
+    persisted service in this codebase already does the same; the
+    apigateway pair was an outlier.
+    """
+    import copy
     return {
-        "apis": _apis,
-        "routes": _routes,
-        "integrations": _integrations,
-        "stages": _stages,
-        "deployments": _deployments,
-        "authorizers": _authorizers,
-        "api_tags": _api_tags,
-        "route_responses": _route_responses,
-        "integration_responses": _integration_responses,
+        "apis": copy.deepcopy(_apis),
+        "routes": copy.deepcopy(_routes),
+        "integrations": copy.deepcopy(_integrations),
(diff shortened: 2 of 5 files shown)


bognari:
Pushed test portability fix (commit 4f8593f) addressing the review point.

The L-2 wire-level test now reads `MINISTACK_ENDPOINT` (default `http://localhost:4566`) the same way `tests/conftest.py:11` does, so tests run unchanged against non-default host/port (Docker networking, alternate CI bind, etc.). The URL is built as `ENDPOINT.rstrip("/") + "/"` to tolerate either form.

The other two tests (M-1, M-2) don't go over HTTP so they don't need the change.

No behaviour change in the default config; all 4 tests still pass.

Nahuel990:
Thanks for these PRs @bognari before releasing the new version I will clean up all those doc strings mentioning GAP_PLAN and H-1 H-2 M-3 and so on convetions we do not handle per project

Nahuel990 (approved):

PR #503, opened 2026-04-27
feat: ministack full Docker variant for DuckDB-backed Athena

### LINKED ISSUE #411: Consider a Debian-based image variant for DuckDB-backed Athena
We are evaluating MiniStack for local development workflows that need Athena support backed by DuckDB. While testing `ministackorg/ministack:1.3.6`, we noticed the image is Alpine-based.

MiniStack's Athena implementation appears to enable the DuckDB engine by checking whether the Python module can be imported:

```python
try:
    import duckdb
    _duckdb_available = True
except ImportError:
    _duckdb_available = False
```

The current image does not include the Python duckdb module, so `ATHENA_ENGINE=auto` falls back to the mock engine.

We tried adding DuckDB in a derived image. The issue is that DuckDB publishes Linux wheels as manylinux wheels, but does not appear to publish `musllinux` wheels for Alpine. As a result, installing duckdb on Alpine falls back to building from source, which requires a native C++ toolchain and can be slow.

On a Debian/glibc-based image, pip can use DuckDB’s published manylinux wheel directly, which would make DuckDB-backed Athena much easier to enable in downstream images.

Would you be open to publishing a Debian-based MiniStack image variant? For example:

* ministackorg/ministack:1.3.6-debian

A separate Debian/glibc variant may be less disruptive than changing the default Alpine image, while still giving users a practical path for DuckDB-backed Athena support.

I'd be happy to help with a patch or PR if this is something you'd be comfortable supporting.

Nahuel990:
Debian/glibc-based superset of the Alpine image. Adds duckdb, psycopg2-binary, and pymysql, which rely on manylinux wheels and do not work reliably on musl. Fixes Athena support in Alpine-based setups.

- Dockerfile.full: strict superset of Dockerfile (keeps asyncssh, SFTP port, MINISTACK_VERSION arg, healthcheck, env vars).
- docker-publish.yml: builds and pushes regular and full variants in parallel, with separate GHA cache scopes. Full image tagged as :full, :{version}-full, and :{major}.{minor}-full.

### DIFF
--- a/.github/workflows/docker-publish-on-pr.yml
+++ b/.github/workflows/docker-publish-on-pr.yml
@@ -2,6 +2,7 @@ name: Docker Publish on PR
 
 on:
   pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
 
 env:
   DOCKERHUB_IMAGE_ORG: ministackorg/ministack-preview-build
@@ -37,7 +38,8 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-      - name: Extract metadata
+      # ── Regular (Alpine) image ──
+      - name: Extract metadata (regular)
         id: meta
         uses: docker/metadata-action@v6
         with:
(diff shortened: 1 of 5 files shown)


Nahuel990:
@jgrumboe do we release this to temporary support it or do you have a better idea?

jgrumboe:
> @jgrumboe do we release this to temporary support it or do you have a better idea?

We can also add it to the https://github.com/ministackorg/ministack/blob/feat/docker-full-variant/.github/workflows/docker-publish-on-pr.yml workflow to get a test build for PRs.
Let's do that first.

PR #504, opened 2026-04-28
fix: {bucket}.{ENDPOINT_HOST}/{object} s3 patterns

mgius-ae:
I'm using @aws-sdk/client-s3@npm:3.1010.0 (via npm:~3) node client.  When configured with `AWS_ENDPOINT=http://ministack:4566` it's atttempting to connect to buckets at `http://<bucket>.ministack:4566`.  Note the lack of an s3 or region in the url.

I added a test that explicitly configures the boto client for path and vhost style, with a dns override patch in there.  This test passes even with the old code, because boto seems to always put s3 in the host header.

Includes tests against the vhost extraction method in app.py which covers some common bucket constructions.

Also includes a couple of skiped tests for dotted buckets which are valid s3 bucket names but aren't working correctly in ministack at the moment.

Tested with node SDK in vhost mode and passes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,10 @@
 """
 Pytest fixtures for MiniStack integration tests.
 """
+import contextlib
 import os
+import socket
+from urllib.parse import urlparse
 import urllib.request
 
 import boto3
@@ -9,24 +12,42 @@ import pytest
 from botocore.config import Config
 
 ENDPOINT = os.environ.get("MINISTACK_ENDPOINT", "http://localhost:4566")
+ENDPOINT_HOST = urlparse(ENDPOINT).hostname
 REGION = "us-east-1"
 
-_kwargs = dict(
+_default_kwargs = dict(
     endpoint_url=ENDPOINT,
     aws_access_key_id="test",
     aws_secret_access_key="test",
     region_name=REGION,
-    # Hardcoded retry and pool settings to reduce transient connection flakes
-    config=Config(
+)
+  # Hardcoded retry and pool settings to reduce transient connection flakes
+_default_config_kwargs = dict(
         region_name=REGION,
         retries={"mode": "standard"},
         max_pool_connections=50,
-    ),
 )
 
 
-def make_client(service):
-    return boto3.client(service, **_kwargs)
+@contextlib.contextmanager
+def patch_endpoint_dns():
+    """Make *.MINISTACK_ENDPOINT subdomains resolve to 127.0.0.1 for virtual-hosted S3 testing."""
+    _real_getaddrinfo = socket.getaddrinfo
+
+    def _patched(host, port, *args, **kwargs):
+        if isinstance(host, str) and host.endswith(f".{ENDPOINT_HOST}"):
+            host = ENDPOINT_HOST
+        return _real_getaddrinfo(host, port, *args, **kwargs)
+
+    socket.getaddrinfo = _patched
+    yield
+    socket.getaddrinfo = _real_getaddrinfo
+
+
+def make_client(service, additional_config_kwargs=None):
+    if additional_config_kwargs is None:
+        additional_config_kwargs = {}
+    return boto3.client(service, **_default_kwargs, config=Config(**_default_config_kwargs, **additional_config_kwargs))
 
 
 _SERIAL_TESTS = {
--- a/tests/test_ministack.py
+++ b/tests/test_ministack.py
@@ -633,3 +633,80 @@ def test_persistence_s3_writes_state_after_ownership_and_public_access_block(tmp
     pers.save_state("roundtrip-bytes", state_with_bytes)
     loaded = pers.load_state("roundtrip-bytes")
     assert loaded["blob"] == b"\x00\x01\xff\xfe"
+
+
+# ---------------------------------------------------------------------------
+# Unit tests for _S3_VHOST_RE / _S3_VHOST_EXCLUDE_RE / _NON_S3_VHOST_NAMES
+# ---------------------------------------------------------------------------
+
+from ministack.app import _S3_VHOST_EXCLUDE_RE, _NON_S3_VHOST_NAMES
+
+class TestS3VhostExcludeRe:
+    """_S3_VHOST_EXCLUDE_RE gates out sub-service hostnames that look like vhosts."""
+
+    def test_excludes_execute_api(self):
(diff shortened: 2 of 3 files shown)


tests/test_ministack.py
mgius-ae, line 1: I had this made because my first assumption was that the regexes had some subtle bug in em.  They don't, but more tests can't hurt?  I can drop em if you like.

tests/conftest.py
```diff
@@ -1,32 +1,56 @@
...
+    _real_getaddrinfo = socket.getaddrinfo
+
+    def _patched(host, port, *args, **kwargs):
+        if isinstance(host, str) and host.endswith(f".{ENDPOINT_HOST}"):
+            host = ENDPOINT_HOST
+        return _real_getaddrinfo(host, port, *args, **kwargs)
+
+    socket.getaddrinfo = _patched
+    yield
+    socket.getaddrinfo = _real_getaddrinfo
```
mgius-ae, line 44: I don't love this, but it was a decent way of testing the boto client's vhost and path code path end to end.

ministack/services/s3.py
```diff
@@ -167,6 +168,7 @@ def restore_state(data):
 
 _BUCKET_NAME_RE = re.compile(r"^[a-z0-9][a-z0-9.\-]{1,61}[a-z0-9]$")
 _IP_RE = re.compile(r"^\d{1,3}(\.\d{1,3}){3}$")
+_MINISTACK_HOST = urllib.parse.urlsplit(os.environ.get("MINISTACK_HOST")).hostname or "localhost"
```
mgius-ae: There are at least 10 different places in the service and test code that load this value, and they're all handling it differently.  There should be a single entrypoint where service code and tests can grab this thing.

MINISTACK_ENDPOINT is similarly strewn around.

Nahuel990:
Thanks for reporting this. The issue is in app.py:_S3_VHOST_RE. It is hard-coded to MINISTACK_HOST (default localhost), so bucket.ministack:4566 never matches. One-line fix in app.py:

_S3_VHOST_RE = re.compile(r"^([a-z0-9][a-z0-9\-]{1,61}[a-z0-9])\..+$")


Could you drop the s3.py changes, keep the tests and conftest helpers, and apply this regex update in app.py? Happy to merge after that.

mgius-ae:
I refactored to split out the test code as a separate commit then applied a commit with only your suggested vhost_re change applied: https://github.com/mgius-ae/ministack/commits/s3_vhost_tests_vhost_re_no_key_extract_changes/

nearly every s3 test fails with a NoSuchBucket.  I'm pretty sure you need the vhost matching regex to take into account host or endpoint, otherwise it's going to treat "localhost" as a bucket name.  Or if the host is "ministack.corp.com" you need to know where to start looking for the vhost part of the host header and not treat ministack.corp as a bucket.

mgius-ae:
Rebased and re-pushed with split test-only and fix commits.

Nahuel990:
@mgius-ae I still see changes in s3.py, there shouldn't be any change in that file. The change should be made in line 44 in app.py

mgius-ae:
I don't think you read my response.  But here's the pushed version with exactly the regex you specified and no changes to s3.py.  It fails all s3 tests locally for me, maybe github actions will differ.

Edit: oh dang, it fails way more than just s3 tests, that RE looks like it's redirecting all kinds of stuff to the s3 subsystem that it shouldn't be.

Nahuel990:
Hey @mgius-ae I read your answer but this problem shouldn't be tackled from s3.py file as the entry point for bucket name is in the app.py file

In the docs, vhost could support these:
  - Bucket: 3-63 chars, lowercase a-z/0-9/./-, start+end alphanum, no .., not IP-format       
  - Reject IPv4 host (with/without port), IPv6 bracketed host                                 
  - Lowercase normalization                                  
  - Strip whitespace                                                                          
  - Tail must be non-empty   
  
  
  So the changes would be 
  
   _IPV4_RE = re.compile(r"^(?:\d{1,3}\.){3}\d{1,3}$")                                         
  _BUCKET_LABEL_RE = re.compile(r"^[a-z0-9](?:[a-z0-9.\-]{1,61}[a-z0-9])$")                   
         
         
         
   def _extract_s3_vhost_bucket(host: str):                                                    
      """Return the bucket if Host is virtual-hosted-style S3, else None."""                  
      if not host:                                          
          return None
      host = host.strip()
      if not host or host.startswith("["):
          return None                                                                         
      host = host.lower()
      if ":" in host:                                                                         
          host = host.rsplit(":", 1)[0]                     
      if not host or _IPV4_RE.match(host) or "." not in host:
          return None                                                                         
      candidate, tail = host.split(".", 1)
      if not tail or tail.startswith("."):                                                    
          return None                                       
      if not _BUCKET_LABEL_RE.match(candidate):
          return None                                                                         
      if ".." in candidate or _IPV4_RE.match(candidate):
          return None                                                                         
      return candidate  
      
      
      
      and around line ~900
      
      instead of the vhost  block 
      
      
        bucket = _extract_s3_vhost_bucket(host)                                                 
      if not bucket or _S3_VHOST_EXCLUDE_RE.search(host) or bucket in _NON_S3_VHOST_NAMES:
          return None

Nahuel990:
Hey @mgius-ae I just merged a follow-up that fixes the virtual-hosted S3 regex in `app.py` (`_extract_s3_vhost_bucket`) and removes the dead vhost branch in `s3.py:_parse_bucket_key`.

When you rebase on `main`, please remove:

1. In `tests/test_ministack.py`: the `from ministack.app import _S3_VHOST_RE, ...` import and the related classes (`TestS3VhostRe`, `TestS3VhostExcludeRe`, `TestNonS3VhostNames`). `_S3_VHOST_RE` is gone.
2. In `tests/test_s3.py`: `test_parse_bucket_key_vhost_plain` and `test_parse_bucket_key_vhost_s3_with_non_endpint_host` — they tested the removed branch.

Keep `test_s3_vhost_get_put_object` — that’s the integration test covering the real behavior.

 Vhost detection is a routing decision, not an S3 decision so that's why it is at app.py level, that regex was sitting there as an extra validation that actually never hit (all the routing decision was coming from the app.py 

After that, the PR should be ready to merge.

mgius-ae:
I should be able to get that rebased and up for merge Wednesday AM PST.  I have another couple tests that aren't pushed as well that I'll include.

mgius-ae:
Cleaned up the tests, added some for that vhost extraction function, and validated against the node sdk.

Nahuel990 (approved):

PR #507, opened 2026-04-28
Fix (sts): GetCallerIdentity retunrs assumed-role ARN instead of hard-coded root

hectormauer:
I suggest the following change because `GetCallerIdentity` always returns `arn:aws:iam::000000000000:root` regardless of the credentials used to make the request. On real AWS, when you call `GetCallerIdentity` using temporary credentials obtained via `AssumeRole`, the response reflects the assumed role, for instance:

```
arn:aws:sts::123456789012:assumed-role/MyRole/my-session
```

My main motivation is that I am using ministack (thank you so much for the fork and the tool!) for testing purposes and one of the tests I want to apply is: obtaining temporary credentials from STS and verifying that those credentials actually represent the intended role before using them to access other services (e.g. S3).

## Fix / Change

I've added a module-level `_sessions` dict in `ministack/services/sts.py` that maps `AccessKeyId → {Arn, UserId}` at `AssumeRole` (and `AssumeRoleWithWebIdentity`) time. `GetCallerIdentity` now extracts the `AccessKeyId` from the AWS Signature V4 `Authorization` header and looks it up in `_sessions`, falling back to the root ARN for non-assumed credentials. 

This gets reset by the `reset()` function clears the dict between test runs via the existing `/_ministack/reset` mechanism. And also already wired in the `_reset_all_state()` in `app.py`

## Tests

I've added the following tests and validated against the updated docker image running locally with the changes.

- `test_get_caller_identity_reflects_assumed_role` — verifies the ARN returned contains the role name and session name after `AssumeRole`
- `test_get_caller_identity_without_assume_role_returns_root` — verifies root credentials still return the root ARN
- `test_get_caller_identity_different_roles_return_different_arns` — verifies two distinct assumed roles produce distinct identities

## Notes

In the [CONTRIBUTING.md](https://github.com/ministackorg/ministack/blob/main/CONTRIBUTING.md) it is mentioned to run `ruff check ministack/` to validate that linting passes. Locally this raises hundreds of errors unrelated to my changes. Maybe some config I am missing?

### DIFF
--- a/ministack/services/sts.py
+++ b/ministack/services/sts.py
@@ -17,6 +17,13 @@ from ministack.services.iam import _p, _xml, _error, _future, \
     _gen_session_access_key, _gen_secret, _gen_session_token
 
 
+_sessions: dict[str, dict] = {}
+
+
+def reset():
+    _sessions.clear()
+
+
 async def handle_request(method, path, headers, body, query_params):
     params = dict(query_params)
     content_type = headers.get("content-type", "")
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks @hectormauer I'll release this in the next iteration

hectormauer:
> Thanks @hectormauer I'll release this in the next iteration

Thank you!

PR #514, opened 2026-04-29
fix(glue): return null LocationUri when database has none

dcrn:
Glue's `CreateDatabase` was defaulting `LocationUri` to `""` when not provided. AWS and Localstack both return `null` in this case, and Iceberg relies on this — it checks `database.getLocationUri() == null` to detect a missing location ([GlueCatalog.java:299](https://github.com/apache/iceberg/blob/8ac703067a2adfae1928748712dc1d47dbc3c22b/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java#L299)).

Fixes the default to `None` so it serialises as JSON `null`, and adds a test covering both the absent and present cases.

### DIFF
--- a/ministack/services/glue.py
+++ b/ministack/services/glue.py
@@ -192,7 +192,7 @@ def _create_database(data):
     _databases[name] = {
         "Name": name,
         "Description": db_input.get("Description", ""),
-        "LocationUri": db_input.get("LocationUri", ""),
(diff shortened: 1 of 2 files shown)


dcrn:
Confirmed this fixes the issue. The preview builds are excellent 👌

Nahuel990 (approved):

PR #517, opened 2026-04-29
fix(s3): tighten virtual-hosted-bucket detection (post-#508 regression)

marcin-nowak-scl:
### Summary

After #508, `_extract_s3_vhost_bucket` treated too many `Host` values as S3 virtual-hosted buckets (`<label>.<anything>`). Requests that are **not** S3 — for example **STS** `GetCallerIdentity` when the client sends **`Host: host.docker.internal:<port>`** (common for Docker / Aspire hitting the gateway on the host) — were routed to the S3 handler and returned **`405 MethodNotAllowed`**.

This branch **requires the host tail to look like an S3 endpoint**: it ends with `MINISTACK_HOST`, or the first segment of the tail is `s3` / `s3-*` (covering regional, dualstack, FIPS, accelerate, website-style hostnames and `bucket.localhost` compatibility). Non-S3 hosts like `host.docker.internal`, `*.local`, and arbitrary domains are no longer misclassified.

### What changed

- `ministack/app.py` — `_extract_s3_vhost_bucket` logic
- `tests/test_ministack.py` — `TestExtractS3VhostBucket` (unit coverage for positive and negative cases)
- **CHANGELOG.md** — fixed / unreleased entry

### How to test

```powershell
if ($py) { ruff check @py }

pytest tests/test_ministack.py::TestExtractS3VhostBucket -v --tb=short
# Optional integration: MiniStack must be listening (e.g. GATEWAY_PORT=4566). POST to loopback but set Host like Docker clients do:
# curl.exe ... "http://127.0.0.1:4566/" -H "Host: host.docker.internal:4566" ...  → expect 200, not 405
```

### Optional: reproduction for reviewers (`<details>`)

Place this block at the **bottom** of the PR body so the summary stays short.

<details>
<summary>Repro: STS via <code>host.docker.internal</code> returned 405 before fix</summary>

**Before:** With MiniStack on the gateway port, a client that used **`Host: host.docker.internal:4566`** (typical for Docker → host, or Aspire/terraform containers) could get S3’s **`405 MethodNotAllowed`** because that hostname was mis-parsed as a virtual-hosted bucket named `host`.

**After:** The same `Host` is **not** treated as S3 vhost; STS handles **`GetCallerIdentity`** and returns **`200`**.

**Why the old one-liner failed from PowerShell on the host:** `curl.exe http://host.docker.internal:4566/` tries to open a TCP connection *to* `host.docker.internal`. On the host that name may not reach your MiniStack process (nothing listening there, or name not wired like inside a container). The bug is about the **`Host` HTTP header**, not about using that name as the connect URL.

**Minimal check** (MiniStack running, default port `4566` — change both URL port and `Host:` if you use `GATEWAY_PORT` elsewhere):

```powershell
curl.exe -sS -w "`nHTTP %{http_code}`n" -X POST "http://127.0.0.1:4566/" `
  -H "Host: host.docker.internal:4566" `
  -H "Content-Type: application/x-www-form-urlencoded" `
  -H "X-Amz-Date: 20260428T120000Z" `
  -d "Action=GetCallerIdentity&Version=2011-06-15"
```

Expect **HTTP 200** and STS XML — not an S3 `MethodNotAllowed` error body.

**From inside a Linux container** (`--network=host` or similar), connecting to `http://host.docker.internal:4566/` can still be valid because there the hostname resolves to the host running MiniStack.

</details>

### PR checklist

- [ ] No new service file
- [x] Gateway / routing behaviour corrected in `ministack/app.py`
- [x] Unit tests added (`tests/test_ministack.py`)
- [ ] **README.md** — N/A
- [x] **CHANGELOG.md** updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **`_extract_s3_vhost_bucket` mis-classified non-S3 hostnames as S3 vhosts (regression from #508)** — the v1.3.18 rewrite accepted any `<word>.<rest>` host as a virtual-hosted bucket, which broke multi-service gateway setups: requests through sibling-Docker DNS names (`host.docker.internal`), mDNS (`service.local`), or any private CNAME got hijacked by the S3 handler and answered with 405 MethodNotAllowed. The canonical reproducer was an Aspire/Docker `terraform-apply` container hitting `STS:GetCallerIdentity` via `http://host.docker.internal:4566` (request fell into the S3 vhost path, S3 returned 405 for POST `/`). The function now requires the tail to look like an S3 endpoint: ends with `MINISTACK_HOST`, or first tail segment is `s3` / `s3-...` (covers AWS dot-region, dash-region, dualstack, FIPS, accelerate, website endpoints, plus the LocalStack-compat `bucket.localhost`). CNAME alias mode (AWS docs rule 3) and S3 Express One Zone remain unsupported — deliberate, the gateway shares port 4566 with all other services and the bug from #508 was that "any tail" hijacked them. 14 unit tests added covering both regression hosts and all positive S3 vhost shapes.
+
 ---
 
 ## [1.3.18] — 2026-04-28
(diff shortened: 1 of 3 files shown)


Nahuel990:
This is fixed already

PR #518, opened 2026-04-29
feat(apigatewayv2): JWT authorizer, HTTP proxy mappings, non-blocking JWKS

marcin-nowak-scl:
### Summary

Extends **API Gateway HTTP API (v2)** support with **JWT authorizers** (including JWKS fetch), **HTTP proxy** integration behaviour with **parameter mapping**, and **non-blocking** HTTP for JWKS and upstream proxy calls so the ASGI gateway stays responsive under slow or failing upstreams.

### What changed

- `ministack/core/nonblocking_http.py` — async-friendly HTTP helper(s)
- Major updates in `ministack/services/apigateway.py`, small related changes in `ministack/services/apigateway_v1.py` and `ministack/services/cognito.py` where JWT / identity overlaps
- Tests: `tests/test_apigatewayv2.py`, `tests/test_apigatewayv1.py`
- **README** and **CHANGELOG** updated

### How to test

```powershell
if ($py) { ruff check @py }

pytest tests/test_apigatewayv2.py tests/test_apigatewayv1.py -v --tb=short
```

### PR checklist

- [x] Router / service updates
- [x] Tests added
- [x] **README.md** / **CHANGELOG.md** updated


Nahuel990:
@marcin-nowak-scl  I'm taking this in, but I'll apply some changes before release.

**Main ones:**
- The 23-line `ministack/core/nonblocking_http.py` will get inlined into the two call sites`core/` shouldn't create a helper module per feature. For a shared async-IO helper to live there, it'd need its own scoped discussion.
- The Cognito RSA-key disk persistence is a real behavioral change (multi-process signing-key invariant) and doesn't belong silently inside an "API Gateway v2" PR, I'll split it into its own commit with its own CHANGELOG entry.
- A few AWS-correctness items I'll tighten: the JWT verifier is RS256-only (will document as a limitation), the reserved-header list needs to be cited against the AWS docs rather than asserted inline, and the JWKS-URL rewrite for Cognito needs to be called out explicitly.
- The `importlib.reload(apigateway_v1)` test will be rewritten reloading a service module mid-suite wipes module-level state and is fragile against test order.

**General ask for next time:** before opening a PR of this size against ministack, please take a pass through the repo to see how existing services are structured (multi-tenancy via `AccountScopedDict`, where helpers live, the persistence/reset pattern), and read the contributing guide. If something isn't obvious, open an issue first so we can align on shape before you spend the time, it's faster for both of us than reworking a 900-line PR at review.

Taking it in because the JWT + parameter-mapping work has real value and I'd rather land it than round-trip.

Nahuel990 (approved):

PR #519, opened 2026-04-29
feat(appsync): AppSync Events API

marcin-nowak-scl:
### Summary

Adds an **AppSync Events** emulator (HTTP publish + WebSocket-style subscribe flows), integrates it with the gateway, and introduces a **wildcard DNS resolver** so Lambda and related vhost-style hostnames resolve correctly in local / Docker setups. Includes **MiniStack network auto-detect** and **`AWS_ENDPOINT_URL` fallback** in the Lambda service so containers can reach the gateway without hard-coding `localhost`.

> [!NOTE]
>  #### Shared commit (`feat/appsync-events` and `feat/opensearch`)
> Commit **`1767b6e`** (`1767b6e3c74be69a1816336df5ff312abad3dc92` — *feat(lambda): MiniStack network auto-detect, `AWS_ENDPOINT_URL` fallback, AppSync vhost DNS, and Linux `host.docker.internal`*) is **included on both branches** so either PR can be proposed against `main` without depending on the other.
> 
> Whichever PR merges **first**, the other will usually need a **`git rebase`** onto updated `main` to deduplicate that commit and resolve overlaps in `ministack/services/lambda_svc.py`, `CHANGELOG.md`, `tests/test_lambda.py`, etc.

### What changed

- New `ministack/services/appsync_events.py` — Event API behaviour and state.
- `ministack/services/dns_resolver.py` — DNS resolution helpers for emulator hostnames.
- `ministack/core/docker_network.py` — network discovery (shared with other feature branches that add Docker-sidecar style integrations).
- Gateway wiring in `ministack/app.py`, routing in `ministack/core/router.py`, extensions in `ministack/services/appsync.py`.
- Tests: `tests/test_appsync_events.py`, `tests/test_dns_resolver.py`, `tests/test_appsync_dns_defaults.py`, plus updates to `tests/test_ministack.py` and persistence symmetry where applicable.
- **README** and **CHANGELOG** updated.

### How to test

```powershell
# Ruff: ministack/ .py changed vs main (see **Ruff / lint** above)
if ($py) { ruff check @py }

pytest tests/test_appsync_events.py tests/test_dns_resolver.py tests/test_appsync_dns_defaults.py tests/test_ministack.py -v --tb=short
```

### PR checklist (from CONTRIBUTING)

- [x] Service / feature code under `ministack/`
- [x] Registered and routed in `ministack/app.py` / `ministack/core/router.py`
- [x] Tests added or updated
- [x] **README.md** updated
- [x] **CHANGELOG.md** updated


Nahuel990:
Thanks @marcin-nowak-scl, the AppSync Events emulator core looks solid and the Events spec is clearly aligned with AWS behavior (channel grammar, WS subprotocol shape, Lambda authorizer payload).

But I cant merge it in its current form for two reasons:

1. The PR bundles three unrelated features (AppSync Events, wildcard DNS server, Lambda network / AWS_ENDPOINT_URL changes). The project requires each of these in its own PR. The AppSync Events emulator is the piece that should land first. The other two need separate discussion.

2. HTTP publish does not work in the bundled state. POST /event with Host: <api_id>.appsync-api.<...> is being captured by the S3 vhost handler before detect_service runs (app.py:1000 vs 1024). The fix added here (appsync-api in _NON_S3_VHOST_NAMES) only checks the first label, but appsync-api is in the tail. This needs to be placed in _S3_VHOST_EXCLUDE_RE instead, since that logic matches the host suffix. This is confirmed by running test_publish_returns_identifiers against a fresh server, which fails with the S3 XML 405 response.

Other items the project needs addressed in the AppSync Events PR:
- _publish (HTTP path) skips the Lambda authorizer that WS subscribe and publish frames invoke. Enforcement should be consistent across both paths
- _connections_lock = asyncio.Lock() at import time can bind to the wrong event loop in tests. This needs lazy initialization
- Dead /v2/apis/{id}/apikeys routes exist. Real SDKs route ApiKey operations to /v1/apis/{id}/apikeys (verified against botocore)
- docker_network.py shouldn't be there as I mentioned in the previous PR we can't add a handler per function as core, this should be placed ideally in lambda_svc and invoked from there
- Rebase onto current main (1.3.21 adds hypercorn_conf.py)

For the DNS resolver PR, the project position is that this must be opt-in (default off). Binding UDP/53 system-wide and forwarding all queries is a significant scope expansion for an AWS API emulator. DNS interception is better treated as a sidecar concern rather than a gateway process concern.

For the Lambda AWS_ENDPOINT_URL PR, this also needs to be opt-in. Real AWS Lambda does not set AWS_ENDPOINT_URL, and making it implicit would break cases where an emulated Lambda intentionally reaches real AWS.

The AppSync Events portion can land once these items are addressed. That core is the work the project wants shipped.

marcin-nowak-scl:
@Nahuel990 Hey, thanks a lot for the feedback! I'll try to address it later today, or tomorrow at the latest. We tested MiniStack against several of our POCs, so I'd like to share what we managed to develop. I'll prepare it in a more granular form and take a look at other changes. Thanks again!

marcin-nowak-scl:
@Nahuel990 The changes are mostly ready - tomorrow I'll test them in action before opening/re-opening the PRs. If I don't run into any issues, it should be ready tomorrow. Thanks again!

marcin-nowak-scl:
@Nahuel990 The changes have been updated and tested.

I moved the API-id based AppSync Events hosts to `_S3_VHOST_EXCLUDE_RE`, which fixes data-plane hosts like `{apiId}.appsync-api.<region>.<host>` and `{apiId}.appsync-realtime-api.<region>.<host>` before S3 vhost handling runs.

I also kept the `_NON_S3_VHOST_NAMES` entries for `appsync-api` and `appsync-realtime-api` because they cover the control-plane/discovery-style hosts where the service label is the first hostname segment, e.g. `appsync-api.<region>.<host>`. `_S3_VHOST_EXCLUDE_RE` only catches the service label when it appears after a leading label, so it does not cover that case.

The DNS resolver and Lambda Docker endpoint/network fallback are excluded from this PR and will be handled as separate opt-in PRs later.

Nahuel990:
Thanks @marcin-nowak-scl for working on this!

Nahuel990 (approved):

PR #524, opened 2026-04-29
feat(athena): glue catalog integration and s3 result persistence

m7w:
## Summary
Add Athena query results persistence and Glue Catalog integration:
- Integrated Glue Service: Now resolves table locations using GetTable.
- S3 Persistence: Saves query results as .csv and .csv.metadata (column names and their types).


Nahuel990:
Thanks @m7w, useful direction. Both pieces match real Athena behavior.

Holding this off as written. The biggest blocker is that the new flow replaces `_rewrite_s3_paths` entirely, so any query using direct s3:// URIs (read_csv('s3://...'), FROM s3://...) silently breaks against the emulator. The rewriter is gone without a fallback. Both paths need to keep working, Glue catalog AND direct-URI queries, so the project requirement is to try Glue per-table and fall through to `_rewrite_s3_paths` for anything Glue does not resolve.

Other items the project needs addressed before this can land:
- KeyError-safe access on Glue’s `StorageDescriptor.Location` and `Parameters.classification` (both are optional in the Glue shape)
- Defensive branch when `s3_location` does not start with s3:// or s3a:// (`stripped` is currently unbound)
- `query.replace(full_name, ...)` corrupts string literals, aliases, and column names that contain the table name as a substring. This needs a word-boundary regex
- OutputLocation parsing via `urllib.parse` instead of the `[4:]` slice (works only because S3’s `lstrip("/")` swallows the extra leading slash)
- Multi-tenancy: results currently write via `s3_svc.handle_request` with no Authorization header, so they land in the default account regardless of the caller's account. ministack is multi-tenant by design and this regresses that
- A test exercising both paths (Glue-resolved table and direct s3:// URI)

Once those land, the PR can go in. Thanks for shipping the engine work.

m7w:
Thanks @Nahuel990  for the detailed feedback! Totally agree on the _rewrite_s3_paths fallback and the regex issue. I'll refactor the table resolution to be more defensive and fix the multi-tenancy headers. Working on the updates now.

Nahuel990:
@m7w I see there are a few things pending but I rather prefer to merge this and take it from here, thanks for the contribution

Nahuel990 (approved):

m7w:
@Nahuel990 Glad to hear that! Thanks for the feedback and for merging the PR. I'm happy to help with those pending items or future improvements if needed.

Nahuel990:
@m7w there are 2 small things that I rather prefer to just fix now on the release (we added mwaa during weekend, and we are trying to also ship glue so we have the full loop) don't worry keep using the emulator and if you see bugs feel free to open issues or PRs

PR #527, opened 2026-04-29
feat(tls): native HTTPS support via USE_SSL

### LINKED ISSUE #526: feat(tls): native HTTPS support via MINISTACK_USE_SSL
## Motivation

Some AWS SDKs hardcode `https://` when calling Cognito Hosted UI endpoints — Amplify v6 does this in three places:

- `signInWithRedirect.mjs` → `` `https://${domain}/oauth2/authorize?…` ``
- `completeOAuthFlow.mjs` → `'https://' + domain + '/oauth2/token'`
- `oAuthSignOutRedirect.mjs` → `` `https://${domain}/logout?…` ``

There's no env var or config option in Amplify to override this. Today every consumer that wants to drive the OAuth flow against a MiniStack-emulated Cognito has to put a separate proxy (Caddy / nginx / Traefik) in front of MiniStack, which doubles the local-dev infra and complicates the cert-trust story.

LocalStack solved this years ago with `USE_SSL=1`. Aligning MiniStack's flag name + behaviour would make it a drop-in replacement for that path too.

## Proposed flags

| Env var | Behaviour |
|---|---|
| `MINISTACK_USE_SSL=1` | Enable HTTPS on the gateway port. |
| `MINISTACK_SSL_CERT=/path/to/cert.pem` | Server certificate (PEM). Required when `USE_SSL=1`. |
| `MINISTACK_SSL_KEY=/path/to/key.pem` | Server private key (PEM). Required when `USE_SSL=1`. |

Possible follow-up (separate PR): `MINISTACK_USE_SSL=auto` — if cert/key paths aren't provided, generate a self-signed cert at startup using `cryptography` (already a Dockerfile dep) so engineers don't need `mkcert`.

## Implementation sketch

Hypercorn natively supports TLS via `config.certfile` / `config.keyfile`, so the change is mostly plumbing. Two paths in `ministack/app.py`'s `main()` need to honor the env vars:

**Foreground path** (the `HypercornConfig()` build):

```python
if os.environ.get("MINISTACK_USE_SSL") == "1":
    cert = os.environ.get("MINISTACK_SSL_CERT")
    key = os.environ.get("MINISTACK_SSL_KEY")
    if not cert or not key:
        print("ERROR: MINISTACK_USE_SSL=1 requires MINISTACK_SSL_CERT and MINISTACK_SSL_KEY")
        raise SystemExit(1)
    config.certfile = cert
    config.keyfile = key
```

**Detached path** (the `subprocess.Popen([..., "-m", "hypercorn", ...])` block):

```python
if os.environ.get("MINISTACK_USE_SSL") == "1":
    cmd += ["--certfile", os.environ["MINISTACK_SSL_CERT"],
            "--keyfile",  os.environ["MINISTACK_SSL_KEY"]]
```

No new dependencies — `hypercorn==0.18.0` and `cryptography>=41.0` are already in the Dockerfile.

## Test plan

A new `tests/test_tls.py` (marked `@pytest.mark.serial` since it spawns a server on an alternate port):

1. Generate a self-signed cert in `tmp_path` with `cryptography`.
2. Spawn `python -m ministack` with `MINISTACK_USE_SSL=1` and the cert/key paths set, on a non-default port.
3. Hit `https://127.0.0.1:<port>/_ministack/health` with `urllib` and `ssl.CERT_NONE`; assert 200.

Existing test suite should pass unchanged (TLS is opt-in).

## Local validation

```bash
mkcert -install
mkcert localhost 127.0.0.1
# → produces localhost+1.pem and localhost+1-key.pem

MINISTACK_USE_SSL=1 \
MINISTACK_SSL_CERT=$PWD/localhost+1.pem \
MINISTACK_SSL_KEY=$PWD/localhost+1-key.pem \
python -m ministack

curl -k https://localhost:4566/_ministack/health
# → 200 OK
```

Happy to send a PR for this if it sounds right — figured I'd open an issue first to confirm the env-var naming and that the simple plumbing approach is what you'd want before I code it up.

prandogabriel:
## Summary
Listener speaks TLS when `USE_SSL=1` (also `true` / `yes`), drop-in compatible with LocalStack's flag. Implementation kept deliberately minimal per @Nahuel990's feedback:

- **`ministack/core/tls.py`** (53 lines) — env-driven cert/key resolution. BYO via `MINISTACK_SSL_CERT` + `MINISTACK_SSL_KEY` (e.g. `mkcert`), otherwise shells out to `openssl req` to mint a self-signed cert under `${TMPDIR}/ministack-tls/` and caches across restarts. CN `ministack-local`; SANs `localhost` / `ministack` / `127.0.0.1` / `::1`. No `cryptography` dep.
- **`ministack/core/hypercorn_conf.py`** (6 lines of code) — exports lowercase `certfile` / `keyfile` when `USE_SSL` is set; nothing otherwise so the listener stays plain HTTP by default.
- **`Dockerfile` / `Dockerfile.full`** — ENTRYPOINT passes `-c file:ministack/core/hypercorn_conf.py` through to hypercorn so the same wiring covers both base images. Alpine ships only `libssl`/`libcrypto`, so `apk add openssl` joins the existing `nodejs bash` line in the regular image; `:full` (debian/slim) already has the CLI.
- **`HEALTHCHECK`** stays inline, made `USE_SSL`-aware (probe uses `https` + an unverified context when TLS is enabled).
- **`app.py` untouched.** Hypercorn's existing `from_pyfile` loader does the work.

Closes #526.

## Why
AWS SDKs hardcode `https://` against Cognito Hosted UI endpoints — Amplify v6 in three places (`signInWithRedirect`, `completeOAuthFlow`, `oAuthSignOutRedirect`). Without native TLS, every consumer driving the OAuth flow against a ministack-emulated Cognito has to put a separate proxy (Caddy / nginx / Traefik) in front. LocalStack solved this with `USE_SSL=true`; aligning the flag name and behaviour makes ministack a drop-in replacement on this path too.

## API

| Env var | Behaviour |
|---|---|
| `USE_SSL=1` (also `true` / `yes`) | Enable HTTPS on the gateway port. Cert/key auto-generated unless BYO paths are provided. |
| `MINISTACK_SSL_CERT` | Optional. PEM-encoded server certificate. Required together with `MINISTACK_SSL_KEY`. |
| `MINISTACK_SSL_KEY` | Optional. PEM-encoded private key. |

Setting only one of the BYO paths exits early with a clear error.

`compose.yml` ergonomics:

```yaml
services:
  ministack:
    image: ministackorg/ministack:latest
    environment:
      - USE_SSL=1   # remove this line for plain HTTP (default)
      # Optional BYO cert:
      # - MINISTACK_SSL_CERT=/certs/cert.pem
      # - MINISTACK_SSL_KEY=/certs/key.pem
```

## Test plan
- [x] `tests/test_tls.py` (5 tests, marked `@pytest.mark.serial`) — each spawns its own hypercorn with `-c file:` on a free port:
  - `test_tls_use_ssl_with_byo_cert` — BYO cert via env vars, hit `https://…/_ministack/health` with `ssl.CERT_NONE`.
  - `test_tls_use_ssl_auto_generated_cert` — `USE_SSL=1` with no paths, exercise the `openssl req` path.
  - `test_tls_use_ssl_accepts_true_value` — verify `USE_SSL=true` parsing.
  - `test_tls_disabled_by_default_serves_http` — pin existing plain-HTTP behaviour.
  - `test_tls_partial_cert_config_rejected` — partial BYO config exits non-zero.
- [x] Verified the regular Alpine Docker image end-to-end: `USE_SSL` unset → HTTP / `healthy`; `USE_SSL=1` → HTTPS / `healthy`; `USE_SSL=true` → HTTPS.
- [x] `ruff check` clean on touched files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+---
+
+## [Unreleased]
+
+### Added
+- **Native HTTPS via `USE_SSL=1`** — the gateway listener now speaks TLS when `USE_SSL=1` (also accepts `true` / `yes`). Drop-in compatible with LocalStack's `USE_SSL=true`, so a `compose.yml` switching from LocalStack to MiniStack doesn't need TLS-specific changes. By default, MiniStack auto-generates a self-signed RSA cert (CN: `ministack-local`, SAN: `localhost`, `ministack`, `127.0.0.1`, `::1`) cached under `${TMPDIR}/ministack-tls/` so the cert survives restarts. To pin a specific cert (e.g. an `mkcert`-issued one for browser trust), set `MINISTACK_SSL_CERT` and `MINISTACK_SSL_KEY` to PEM paths. Auto-generation uses the `cryptography` package (already shipped in both the regular and `:full` images). Unblocks AWS SDKs that hardcode `https://` against Cognito Hosted UI endpoints (e.g. Amplify v6) without needing a separate TLS-terminating proxy. Closes #526.
 
 ---
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ LABEL maintainer="MiniStack" \
       description="Local AWS Service Emulator — drop-in LocalStack replacement"
 
 # Upgrade base packages to pick up latest security patches.
-RUN apk upgrade --no-cache && apk add --no-cache nodejs bash && rm -f /usr/bin/wget /bin/wget \
+RUN apk upgrade --no-cache && apk add --no-cache nodejs bash openssl && rm -f /usr/bin/wget /bin/wget \
     && rm -rf /usr/local/lib/python3.12/site-packages/pip* \
               /usr/local/bin/pip*
 
@@ -62,12 +62,13 @@ ENV MINISTACK_VERSION=${MINISTACK_VERSION} \
     RDS_PERSIST=0 \
     ELASTICACHE_BASE_PORT=16379 \
     LAMBDA_EXECUTOR=local \
+    USE_SSL=0 \
     PYTHONUNBUFFERED=1
 
 EXPOSE 4566 2222
 
-# Pure Python healthcheck — no curl dependency
+# Pure Python healthcheck — honours USE_SSL so it works for both HTTP and HTTPS.
 HEALTHCHECK --interval=10s --timeout=3s --start-period=5s --retries=3 \
-    CMD python -c "import urllib.request; urllib.request.urlopen('http://localhost:4566/_ministack/health')" || exit 1
+    CMD python -c "import os,ssl,urllib.request as r; t=os.environ.get('USE_SSL','').strip().lower() in ('1','true','yes'); r.urlopen(('https' if t else 'http')+'://localhost:'+os.environ.get('GATEWAY_PORT','4566')+'/_ministack/health',context=ssl._create_unverified_context() if t else None)" || exit 1
 
-ENTRYPOINT ["python", "-m", "hypercorn", "ministack.app:app", "--bind", "0.0.0.0:4566", "--keep-alive", "75"]
+ENTRYPOINT ["python", "-m", "hypercorn", "ministack.app:app", "-c", "file:ministack/core/hypercorn_conf.py", "--bind", "0.0.0.0:4566", "--keep-alive", "75"]
--- a/Dockerfile.full
+++ b/Dockerfile.full
@@ -77,12 +77,13 @@ ENV MINISTACK_VERSION=${MINISTACK_VERSION} \
     RDS_PERSIST=0 \
     ELASTICACHE_BASE_PORT=16379 \
     LAMBDA_EXECUTOR=local \
+    USE_SSL=0 \
     PYTHONUNBUFFERED=1
 
 EXPOSE 4566 2222
 
-# Pure Python healthcheck — no curl dependency
+# Pure Python healthcheck — honours USE_SSL so it works for both HTTP and HTTPS.
 HEALTHCHECK --interval=10s --timeout=3s --start-period=5s --retries=3 \
-    CMD python -c "import urllib.request; urllib.request.urlopen('http://localhost:4566/_ministack/health')" || exit 1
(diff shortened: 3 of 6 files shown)


Nahuel990:
@prandogabriel thanks for the PR but it's way too big for what it does. let me break it down:

tls.py doesn't need cryptography both base images already ship openssl. the entire cert generation is a subprocess call, not 80 lines of x509 builder. use_ssl_enabled() + resolve_tls_material() + an openssl one-liner puts the whole module at ~30 lines. cryptography is there for KMS and Cognito where we need the python API for key ops, generating a PEM file on disk doesn't need it

don't touch app.py (the only reason you're changing the ENTRYPOINT and routing through main() is to conditionally append certfile/keyfile) hypercorn already supports a config file (-c file:path) with certfile/keyfile options. create a small hypercorn_conf.py that checks USE_SSL and resolves the cert paths, ~6 lines, app.py stays untouched

healthcheck doesn't need to be a module it's a proof of work... current main already has it inline in the Dockerfile. keep it inline, just make it aware based on USE_SSL

prandogabriel:
@Nahuel990 thanks for the review — pushed a force-update with all three points addressed:

1. **`tls.py` ditches `cryptography`** — now ~53 lines, single `openssl req` subprocess call. The cryptography import is gone from this module.
2. **`app.py` untouched** — new `ministack/core/hypercorn_conf.py` (6 lines of code) exports `certfile` / `keyfile` for hypercorn's `from_pyfile` loader. Wired in via `-c file:…` in the ENTRYPOINT so foreground hypercorn picks it up automatically.
3. **`HEALTHCHECK` stays inline** in both Dockerfiles, just made `USE_SSL`-aware so the probe switches scheme + uses an unverified context when TLS is on. The `ministack/healthcheck.py` module is gone.

One small premise correction: `python:3.12-alpine` ships only `libssl3` / `libcrypto3` (the libraries) — the `openssl` CLI binary isn't in the default install. Confirmed with `apk list --installed | grep openssl` on a fresh `python:3.12-alpine`. So I added `openssl` to the existing `apk add --no-cache nodejs bash` in `Dockerfile` (regular Alpine image). `Dockerfile.full` (debian/slim) already has the CLI, no change there. The added Alpine package is ~1.2 MB compressed.

Final diff vs main: 6 files, 233 insertions, 7 deletions. Locally verified with `docker build` + `docker run` on the regular Alpine image: `USE_SSL` unset → HTTP / `healthy`; `USE_SSL=1` → HTTPS / `healthy`; `USE_SSL=true` → HTTPS.

Nahuel990:
Awesome! much better! About the failing tests dont worry we are aware of it, there is currently a problem with Github and xdist for web sockets(if your tests dont fail I will re run them after a few minutes)

Nahuel990 (approved):

PR #538, opened 2026-05-01
feat(eventbridge): auto-fire rate() scheduled rules via background scheduler

hiddengearz:
## Why

EventBridge `rate(N minute|hour|day)` rules created via `PutRule` were never auto-invoked in Ministack. Two root causes:

1. `_dispatch_event()` has an early-exit guard (`if not rule.get("EventPattern"): continue`) that unconditionally skips schedule-only rules on every `PutEvents` call.
2. No background scheduler thread existed anywhere in the codebase — `app.py`'s `_handle_lifespan` starts no periodic task for rule ticking.

This broke any application that relies on scheduled Lambda invocations locally (e.g. `AWS::Events::Rule` with `ScheduleExpression: rate(1 minute)` targeting a Lambda).

## What

- **`ministack/services/eventbridge.py`** — adds a daemon thread (`eb-scheduler`) that ticks every 10 seconds and invokes enabled `rate()` rules whose interval has elapsed:
  - `_parse_rate_seconds(expr)` — pure parser for `rate(N minute|minutes|hour|hours|day|days)`; returns `None` for unsupported expressions (cron, seconds) so they are silently skipped.
  - `_tick_scheduled_rules()` — iterates `_rules._data` **directly** (bypassing `AccountScopedDict`'s ContextVar scope) so all tenant accounts are visited, not just the default. On first sight of a rule, records the timestamp and defers firing — matching the AWS-documented behaviour that the first fire is one full interval after rule creation, not immediately. Dispatches via the existing `_invoke_target` path (Lambda, SQS, SNS).
  - Scheduled event payload conforms to the AWS wire format (see References).
  - `reset()` clears `_rule_last_fired` so test isolation is maintained.

- **`tests/test_eventbridge.py`** — 19 new tests:
  - 12 parametrised cases for `_parse_rate_seconds` (all 6 valid rate units singular/plural, plus `cron()`, `rate(1 second)`, empty string, unsupported units, garbage).
  - 7 behavioural unit tests for `_tick_scheduled_rules` via direct module import + `unittest.mock.patch`: first-sight countdown init, fire-after-interval, skip-before-interval, DISABLED rule skip, cron skip, no-targets no-crash, `reset()` clearing state.
  - 1 slow end-to-end integration test (`test_eventbridge_scheduled_rule_autofire`, gated by `MINISTACK_SLOW_TESTS=1`) that wires a real Lambda target and asserts at least one invocation arrives via SQS after the interval elapses.

## AWS Compliance

The implementation targets 1:1 behaviour with AWS EventBridge scheduled rules:

| Behaviour | AWS spec | This implementation |
|---|---|---|
| First-fire timing | "The countdown begins when you create the rule" — fires after one full interval | `_rule_last_fired` seeded to `now` on first sight; fires only after `interval` elapses |
| Rate expression units | `minute`, `minutes`, `hour`, `hours`, `day`, `days` only (no `seconds`) | Regex anchored to those 6 units; anything else returns `None` (skip) |
| Event `source` | `"aws.events"` | ✓ |
| Event `detail-type` | `"Scheduled Event"` | ✓ |
| Event `detail` | `{}` (empty JSON object) | ✓ |
| Event `version` | `"0"` | ✓ |
| Event `resources` | `[rule_arn]` | ✓ |
| Event `time` | ISO 8601 UTC string e.g. `"2016-12-30T18:44:49Z"` | Converted from Unix float via `datetime.fromtimestamp(..., tz=timezone.utc).strftime(...)` |

## References

- [AWS EventBridge — Schedule expressions for rules](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html) — rate expression syntax and unit list; first-fire timing documented as "countdown begins when you create the rule".
- [AWS EventBridge — Events from AWS services: EventBridge scheduled events](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-service-event.html#eb-service-event-cloudwatch-events) — canonical scheduled event payload showing `source: aws.events`, `detail-type: Scheduled Event`, `detail: {}`, `resources: [rule_arn]`, and ISO 8601 `time`.
- [AWS EventBridge — Event structure reference](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html) — top-level field definitions (`version`, `id`, `source`, `detail-type`, `detail`, `resources`, `time`).

## Risk Assessment

Low. The scheduler runs as a daemon thread and only touches existing `_rules._data` / `_targets._data` / `_invoke_target` paths already exercised by integration tests. No new external dependencies. `cron()` expressions are stored and round-tripped normally; they are skipped by the scheduler (not rejected) so no existing behaviour changes. `reset()` coverage prevents state leaking between tests.

## Test Plan

```
# Fast unit tests (< 1 s)
pytest tests/test_eventbridge.py -k "scheduler" -v

# All EventBridge tests
pytest tests/test_eventbridge.py -v

# Full end-to-end (requires running Ministack, takes ~80 s)
MINISTACK_SLOW_TESTS=1 pytest tests/test_eventbridge.py::test_eventbridge_scheduled_rule_autofire -v
```

- [x] `ruff check ministack/services/eventbridge.py` — clean
- [x] 19 unit/integration tests added and passing
- [x] `CHANGELOG.md` entry added under `[Unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Added
+- **EventBridge scheduled rule auto-fire** — `rate(N minute|hour|day)` rules now fire automatically in Ministack. A background daemon thread (`eb-scheduler`) ticks every 10 seconds and dispatches enabled rules whose interval has elapsed. Follows the AWS-documented behaviour: the countdown starts at rule creation time so the first fire is one full interval after the rule is enabled, not immediately. The scheduled event payload matches the AWS wire format exactly: `"source": "aws.events"`, `"detail-type": "Scheduled Event"`, `"detail": {}`, `"version": "0"`, `"resources": [rule_arn]`, and `"time"` as an ISO 8601 UTC string. Iterates `_rules._data` directly so multi-tenant (multi-account) setups work correctly. `cron()` expressions are accepted and stored but not yet auto-fired. `reset()` clears the scheduler state. 12 unit tests cover rate-expression parsing and all tick-decision branches; one slow end-to-end integration test (gated by `MINISTACK_SLOW_TESTS=1`) exercises the full path from `PutRule`/`PutTargets` through Lambda invocation.
+
+---
+
 ## [1.3.22] — 2026-04-30
 
 ### Added
--- a/ministack/services/eventbridge.py
+++ b/ministack/services/eventbridge.py
@@ -29,7 +29,7 @@ import os
 import re
 import threading
 import time
-from datetime import datetime
+from datetime import datetime, timezone
 
 from ministack.core.responses import (
     AccountScopedDict,
@@ -38,6 +38,7 @@ from ministack.core.responses import (
     get_region,
     json_response,
     new_uuid,
+    set_request_account_id,
 )
 
 logger = logging.getLogger("events")
@@ -83,6 +84,10 @@ _replays = AccountScopedDict()              # replay_name -> replay record
 _endpoints = AccountScopedDict()            # endpoint name -> endpoint record
 # Partner event sources, per-account (key: "account|name" pattern inside each tenant).
 _partner_event_sources = AccountScopedDict()
+
+# Tracks when each scheduled rule last fired: {(account_id, rule_key): timestamp}.
+# Plain dict (not AccountScopedDict) because the scheduler thread owns it globally.
+_rule_last_fired: dict = {}
 
 
 def _ensure_default_bus():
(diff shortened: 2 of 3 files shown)


hiddengearz:
My (claude's) first attempt at adding a new feature so feel free to rip it to shreds. Found out scheduled event's weren't actually firing :$

Nahuel990:
We don’t rip things apart here :D  just aiming to make things better for everyone. Thanks for putting this together!

Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
We don’t rip things apart here :D  just aiming to make things better for everyone. Thanks for putting this together!

Nahuel990 (approved):

Nahuel990 (approved):

PR #540, opened 2026-05-01
feat(eventbridge): add cron() schedule expression support

hiddengearz:
## Context

This extends #538, which added a background scheduler for EventBridge `rate()` rules. In that PR we noted:

> `cron()` expressions are accepted and stored but not yet auto-fired.

AWS EventBridge treats `cron()` as a first-class schedule type alongside `rate()` — both are documented on the same page and accepted by the same `PutRule` API parameter. Leaving `cron()` silently unfired means any application that uses `cron(0 12 * * ? *)` or similar (very common for time-of-day jobs) gets no local invocations at all, with no error to indicate why.

## New dependency: `croniter==6.2.2`

> **Question for maintainers:** are you comfortable adding `croniter` as a runtime dependency?

The alternative is a hand-rolled zero-dependency parser (~80 lines), which would cover the common subset (`*`, `?`, integers, ranges, steps, lists, named weekdays) but miss the `L` (last day), `W` (nearest weekday), and `#` (nth weekday) operators. `croniter` handles the full AWS cron syntax including those edge cases, is MIT-licensed, actively maintained, and has no transitive dependencies beyond `python-dateutil` (which is already pulled in by `boto3` in dev environments).

If you'd prefer to avoid the dependency, I'm happy to submit an alternative PR with the hand-rolled parser instead — just let me know.

## What changed (beyond #538)

- **`ministack/services/eventbridge.py`**
  - `_parse_cron_expr(expr)` — converts AWS `cron(Min Hr DoM Mon DoW Year)` to a 5-field croniter string. Year (field 6) is dropped; AWS `?` wildcards are replaced with croniter `*`. Returns `None` for non-cron input.
  - `_tick_scheduled_rules()` — now handles both expression types. `rate()` rules use interval arithmetic (unchanged). `cron()` rules call `croniter(expr, last_fired_dt).get_next(datetime)` and fire when `now >= next_fire`.
  - `from croniter import croniter` added to imports.

- **`requirements.txt` / `pyproject.toml`** — `croniter==6.2.2` pinned in both.

- **`tests/test_eventbridge.py`** — 11 new tests on top of the 19 from #538:
  - 8 parametrised `test_scheduler_parse_cron_expr` cases (valid conversions + invalid inputs)
  - `test_scheduler_cron_first_sight_initializes_countdown` — first tick records timestamp, no dispatch
  - `test_scheduler_cron_fires_after_scheduled_time` — `cron(0 * * * ? *)` with `last_fired` 2 h ago → fires
  - `test_scheduler_cron_skips_before_scheduled_time` — same rule with `last_fired` 10 s ago → skips

## AWS Compliance

| Behaviour | AWS spec | This implementation |
|---|---|---|
| Field order | `cron(Min Hr DoM Mon DoW Year)` | Parsed positionally; Year stripped |
| `?` wildcard | Required on DoM **or** DoW (the unused one) | Replaced with `*` before passing to croniter |
| `L`, `W`, `#` operators | Supported by AWS | Supported via croniter |
| Ranges, steps, lists | e.g. `MON-FRI`, `*/5`, `1,15` | Supported via croniter |
| First-fire timing | "Countdown begins when rule is created" | On first tick: record timestamp, skip. Fire only after first full occurrence elapses. |
| Event payload | Same as `rate()` — `source: aws.events`, `detail-type: Scheduled Event`, etc. | Identical path; same `_invoke_target` call |

## References

- [AWS EventBridge — Schedule expressions: cron expressions](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html#eb-cron-expressions) — 6-field format, `?` wildcard semantics, `L`/`W`/`#` operator definitions, and the mutual-exclusion constraint between DoM and DoW.
- [AWS EventBridge — Schedule expressions: rate expressions](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html#eb-rate-expressions) — confirms `rate()` and `cron()` are the two and only two expression types accepted by `ScheduleExpression`.
- [croniter on PyPI](https://pypi.org/project/croniter/) — MIT license, `python-dateutil` only transitive dep, supports extended cron including AWS's `L`/`W`/`#`.

## Test Plan

```bash
# Unit tests only (no server needed, < 1 s)
pytest tests/test_eventbridge.py -k "scheduler" -v

# Full EventBridge suite
pytest tests/test_eventbridge.py -v

# End-to-end (requires running Ministack, ~80 s)
MINISTACK_SLOW_TESTS=1 pytest tests/test_eventbridge.py::test_eventbridge_scheduled_rule_autofire -v
```

- [x] `ruff check ministack/services/eventbridge.py` — clean
- [x] 29 scheduler unit tests passing (12 rate-parse + 8 cron-parse + 9 tick-behaviour)
- [x] `CHANGELOG.md` updated


Nahuel990:
@hiddengearz this please -> `If you'd prefer to avoid the dependency, I'm happy to submit an alternative PR with the hand-rolled parser instead — just let me know`

The original croniter package was abandoned for ~3 years (2019-2022) before the maintainer handed it off, also AWS cron is a 6-field dialect, not 5 and AWS uses ? to enforce mutual exclusion between day-of-month and day-of-week

hiddengearz:
Thanks for the feedback @Nahuel990 — totally agree. Dropped `croniter` and replaced it with a zero-dependency hand-rolled parser.

**What changed in the latest push:**

- Removed `croniter==6.2.2` from `requirements.txt` and `pyproject.toml`
- Replaced `_parse_cron_expr()` (which just converted to a 5-field string for croniter) with three new functions:
  - `_cron_field(field, lo, hi, names)` — expands a single field token (`MON-FRI`, `*/5`, `1,15`, etc.) into a `frozenset` of matching integers
  - `_parse_cron_fields(expr)` — parses the full 6-field AWS cron expression into expanded sets; preserves the raw `dom` / `dow` tokens so the `?` mutual-exclusion rule can be enforced correctly (rather than just converting `?` → `*`)
  - `_cron_next_fire(fields, after_dt)` — finds the next matching datetime using per-field forward jumps so sparse schedules (monthly, yearly) don't iterate every minute
- AWS DoW numbering handled natively: `(python_isoweekday % 7) + 1` → AWS 1=SUN…7=SAT
- Tests updated: replaced the old string-conversion assertions with 9 validity cases and 3 concrete next-fire assertions (`same-day`, `day-wrap`, `MON-FRI` weekday skip)

35 unit tests passing, ruff clean, no new dependencies.

Nahuel990 (approved):

Nahuel990:
Thanks @hiddengearz great work on this!

PR #547, opened 2026-05-03
feat(cloudtrail): add CloudTrail service with in-memory audit log

### LINKED ISSUE #486: Add AWS CloudTrail service emulation (LookupEvents, in-memory audit log)
### AWS Service name

CloudTrail

### Operations needed

**Trail stubs (control plane compatibility)**
- DescribeTrails
- GetTrail
- GetTrailStatus

**Audit log (data plane)**
- LookupEvents — filter by EventName, ResourceName, Username, time range

### Use case

Integration tests that verify which AWS APIs were called during a
workflow have no emulation target today. Common patterns that break
without this:

- Test suites that assert a specific API was called with specific
  parameters (e.g. verify PutObject was called for a given bucket
  after a workflow completes)
- CDK/Terraform stacks that create Trail resources fail to plan locally
- Security automation that reads CloudTrail to detect API usage cannot
  be tested end-to-end

No real audit delivery is needed. The emulator just needs to record
API calls in-memory and expose them via LookupEvents with realistic
filtering.

### Design

Every API call passing through ministack is appended to an in-memory
AccountScopedDict as a CloudTrail event record. LookupEvents filters
and returns matching records. Trail operations (DescribeTrails,
GetTrail, GetTrailStatus) return a default stub trail with IsLogging
true for infrastructure-as-code compatibility.

Event record shape matches the AWS CloudTrail API:
EventId, EventName, EventTime, Username, Resources, CloudTrailEvent

- [x] Yes, I would like to implement this and open a PR

AdigaAkhil:
Closes #486

## What this adds

**Audit log (data plane)**

Every API call passing through the ASGI dispatch loop is appended to a per-account \`collections.deque\` as a CloudTrail event record. Recording is off by default and opt-in via \`CLOUDTRAIL_RECORDING=1\` or the runtime config endpoint.

\`LookupEvents\` supports all 8 AWS LookupAttributes:

* EventName
* EventSource
* Username
* ResourceName
* ResourceType
* EventId
* ReadOnly
* AccessKeyId

Each event record includes the full \`CloudTrailEvent\` JSON shape: \`eventVersion\`, \`userIdentity\`, \`eventTime\`, \`eventSource\`, \`eventName\`, \`awsRegion\`, \`sourceIPAddress\`, \`userAgent\`, \`requestParameters\` (secrets scrubbed), \`responseElements\`, \`requestID\`, \`eventID\`, \`eventType\`, \`recipientAccountId\`.

**Control plane (always on)**

Stub trail operations for CDK/Terraform IaC compatibility: \`CreateTrail\`, \`DeleteTrail\`, \`GetTrail\`, \`DescribeTrails\`, \`GetTrailStatus\`, \`StartLogging\`, \`StopLogging\`, \`PutEventSelectors\`, \`GetEventSelectors\`, \`AddTags\`, \`ListTags\`, \`RemoveTags\`.

## Design notes

* **Feature flag** — recording gated on \`CLOUDTRAIL_RECORDING=1\` (env var) or \`POST /_ministack/config {"cloudtrail._recording_enabled": "true"}\`. Control plane stubs are always available regardless of flag.
* **Bounded memory** — per-account \`collections.deque(maxlen=CLOUDTRAIL_MAX_EVENTS)\`, default 10 000. Oldest events are dropped automatically at the cap.
* **Secrets scrubber** — \`requestParameters\` stripped of \`SecretAccessKey\`, \`Password\`, \`AuthToken\`, \`Signature\`, \`Authorization\`, \`X-Amz-Security-Token\`, \`Credentials\`, \`SecretString\`, \`SecretBinary\` before storage.
* **Self-recording guard** — CloudTrail's own requests and all \`/_ministack/*\` admin paths are skipped in the recorder.
* **Persistence** — trail config (name, S3 bucket, event selectors, tags) persists across restarts. Events are ephemeral; timestamps would be meaningless after a restart.

## Benchmark results

Measured on macOS against the local server (600 requests per scenario, sequential, warmed up).

| Scenario | Off (req/s) | On (req/s) | Delta |
|---|---|---|---|
| S3 HeadBucket | 1440 | 1432 | -0.6% |
| DynamoDB GetItem | 1381 | 1354 | -2.0% |
| SQS ReceiveMessage | 1368 | 1371 | +0.2% |

All three scenarios are within the ~2% threshold. p50/p95/p99 latencies are unchanged (0.7ms / 0.8ms / 0.9ms) in both modes.

Benchmark script at \`tests/bench_recording.py\`. Run with \`python tests/bench_recording.py --compare\`.

## Out of scope for v1 (follow-ups)

* \`AWS::CloudTrail::Trail\` CloudFormation provisioner
* Real S3 log delivery
* Log file integrity validation
* EventBridge integration
* Insight events and Data events

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -249,6 +249,7 @@ SERVICE_REGISTRY = {
     "waf": {"module": "waf_v1"},
     "waf-regional": {"module": "waf_v1"},
     "wafv2": {"module": "waf"},
+    "cloudtrail": {"module": "cloudtrail"},
 }
 
 SERVICE_HANDLERS = {
@@ -282,6 +283,7 @@ _state_map = {
     "scheduler": "scheduler", "autoscaling": "autoscaling",
     "eks": "eks", "backup": "backup", "pipes": "pipes",
     "cloudfront_keyvaluestore": "cloudfront_keyvaluestore",
+    "cloudtrail": "cloudtrail",
 }
 
 SERVICE_NAME_ALIASES = {
@@ -722,6 +724,7 @@ async def _handle_admin_config_request(path: str, method: str, body: bytes):
         "stepfunctions._sfn_mock_config",
         "stepfunctions._SFN_WAIT_SCALE",
         "lambda_svc.LAMBDA_EXECUTOR",
+        "cloudtrail._recording_enabled",
     }
     try:
         config = json.loads(body) if body else {}
@@ -749,6 +752,8 @@ async def _handle_admin_config_request(path: str, method: str, body: bytes):
                     logger.warning("/_ministack/config: invalid SFN_WAIT_SCALE=%r", value)
                     continue
                 value = float_value
+            elif key == "cloudtrail._recording_enabled":
+                value = str(value).lower() in ("1", "true", "yes")
             setattr(mod, var_name, value)
             applied[key] = value
         except (ImportError, AttributeError) as e:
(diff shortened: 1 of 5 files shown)


Nahuel990:
Great PR @AdigaAkhil the only minor thing, can you set the test to run serial? (_SERIAL_TESTS in conftest.py) because Github uses xdist and it's very likely to trigger both paths at the same time and have a race condition

jgrumboe:
> Great PR @AdigaAkhil the only minor thing, can you set the test to run serial? (_SERIAL_TESTS in conftest.py) because Github uses xdist and it's very likely to trigger both paths at the same time and have a race condition

@Nahuel990, with the current settings, xdist is parallelizing tests over files. Tests within the same file are executed in serial mode AFAIK.

Nahuel990:
@jgrumboe there is a flag "serial" in the CI to not share workers among certain tests (for instance, web sockets, cron tasks, etc) and here we will trigger record, delete, and lookup, so I think if at some point the tests for lookup and delete are triggered at the same time, lookup could fail

jgrumboe:
> @jgrumboe there is a flag "serial" in the CI to not share workers among certain tests (for instance, web sockets, cron tasks, etc) and here we will trigger record, delete, and lookup, so I think if at some point the tests for lookup and delete are triggered at the same time, lookup could fail

I know the "serial" flag you're referring to.
Deletions are currently only stubs for cloudtrails. As far as I understood, the code lookups will always work because CloudTrail recording is enabled for the whole test suite.
To be failsafe, if controlplane functions are no stubs anymore, the lookup tests would also need to call create_trail(), I think. It should also be enough to just add `test_delete_trail` to _SERIAL_TESTS.

Nahuel990:
@jgrumboe well we can merge it as is, in the worst case is a one liner PR change, this could go in the next release

AdigaAkhil:
@Nahuel990  , I can make the remaining changes.
Currently I am away from my Laptop

Nahuel990:
@AdigaAkhil don't worry we just merge to main, as it is not released wont break anythign

Nahuel990 (approved):

PR #553, opened 2026-05-04
Ecs container metadata

YakirOren:
## What this adds

  **Metadata V4 endpoints**

  Every container started by `RunTask` is now injected with `ECS_CONTAINER_METADATA_URI_V4=http://<ministack>:<gateway>/v4/<token>`. The gateway responds on:

  * `GET /v4/<token>` — current container payload
  * `GET /v4/<token>/task` — task payload, with sibling `Containers` array
  * `GET /v4/<token>/stats`, `GET /v4/<token>/task/stats` — empty `{}` stub

  Tokens are per-container `secrets.token_urlsafe(32)` values, registered when `RunTask` spawns the container and unregistered on `StopTask` or run failure.

  **RunTask: more task-def fields translated**

  * `privileged` / `Privileged` → Docker `privileged`
  * `linuxParameters.capabilities.add` → Docker `cap_add`
  * `pidMode: host` → Docker `pid_mode: host`
  * `networkMode: host` → Docker `network_mode: host`
  * `volumes` + `mountPoints` → Docker bind mounts (list-of-binds, not dict, so two ECS volumes sharing a host SourcePath don't collide)

  **Standard ECS container labels**

  Each spawned container gets the AWS-spec labels: `com.amazonaws.ecs.task-arn`, `com.amazonaws.ecs.container-name`, `com.amazonaws.ecs.task-definition-family`,
  `com.amazonaws.ecs.task-definition-version`, `com.amazonaws.ecs.cluster`.


Nahuel990:
@YakirOren thanks for raising this PR, even though we have ECS we don't support Xray nor OpenTelemetry (they are not in the roadmap neither as we might need to add UDP dependency which is very niche) could you describe what is your use case for this service? Also could you move your changelog to the top as it's not part of a previous release?

YakirOren:
Hi @Nahuel990 
We're building an ECS instrumentation agent for runtime security & observability,

Without the V4 endpoint in Ministack, we have no way to test the agent end-to-end locally / in CI  

We picked Ministack over the alternatives specifically because of the CloudFormation support,
and this is the only local AWS emulator that lets us deploy it the same way customers do.

Nahuel990:
@YakirOren thanks for clarification! I am merging it now and will tag it for the next release in a few hours

Nahuel990 (approved):

PR #560, opened 2026-05-04
fix(cfn): preserve S3 bucket physical resource ID on stack update

erick-reis-gran:
## Summary

Fixes two bugs in the CloudFormation provisioner that broke Serverless Framework deployments.

### Bug 1: Auto-named S3 bucket gets a new name on every stack update

When a CloudFormation stack is **updated**, `AWS::S3::Bucket` resources without an explicit `BucketName` (such as the `ServerlessDeploymentBucket` created by Serverless Framework) were incorrectly getting a **new random bucket name** on every update instead of preserving the existing one.

This caused Lambda functions referencing the bucket via `{"Ref": "ServerlessDeploymentBucket"}` to resolve to the **new empty bucket** instead of the one where the deployment zip was actually uploaded, resulting in Lambda functions created with **no code**.

**Root cause:** `_RESOURCE_HANDLERS` had no `"update"` entry for `AWS::S3::Bucket`. On stack update, `_update_resource` fell back to `_provision_resource(type, physical_id, props, stack_name)`, passing the old physical resource ID where `_s3_create` expected a logical ID. `_s3_create` called `_physical_name()` which always generates a new random suffix, creating a brand new empty bucket.

**Fix:** Add `_s3_update` handler that preserves the existing physical resource ID (bucket name) when the bucket is updated in place (matching real AWS CloudFormation behavior). Also make `_s3_create` use `setdefault` to avoid wiping existing bucket data.

### Bug 2: Lambda functions deployed via CFN missing `CodeSize` and proper `CodeSha256`

The CF provisioner `_lambda_create` never set `CodeSize` in the config dict and hardcoded `CodeSha256` to `"cfn-deployed"`. This caused UIs (like StackPort) to display **"NaN undefined"** for code size, even when the code was correctly loaded from S3.

**Fix:** Compute `CodeSize` and `CodeSha256` from the actual zip bytes when available.

## Reproduction

1. Deploy a Serverless Framework service to ministack using `serverless-localstack`
2. Deploy again (update)
3. Observe: Lambda functions have no code, "NaN undefined" code size, and a new empty S3 bucket is created on each update
4. S3 bucket list shows multiple deployment buckets, only the oldest has data

## Test plan

- [x] `test_cfn_auto_named_s3_bucket_stable_across_updates` — auto-named bucket keeps same name and objects across stack updates
- [x] `test_cfn_lambda_s3_ref_bucket_has_code_size` — Lambda with `Code.S3Bucket: {Ref: Bucket}` reports correct `CodeSize`, real `CodeSha256`, and code is downloadable
- [x] Existing CFN tests with explicit `BucketName` still pass (no regression)
- [x] Full Serverless Framework deploy + redeploy validated end-to-end locally

### DIFF
--- a/ministack/services/cloudformation/provisioners.py
+++ b/ministack/services/cloudformation/provisioners.py
@@ -118,14 +118,40 @@ def _update_resource(resource_type: str, physical_id: str, old_props: dict,
 
 def _s3_create(logical_id, props, stack_name):
     name = props.get("BucketName") or _physical_name(stack_name, logical_id, lowercase=True, max_len=63)
-    _s3._buckets[name] = {
+    _s3._buckets.setdefault(name, {
         "created": now_iso(),
         "objects": {},
         "region": get_region(),
-    }
+    })
     versioning = props.get("VersioningConfiguration", {})
     if versioning.get("Status") == "Enabled":
         _s3._bucket_versioning[name] = "Enabled"
+    attrs = {
+        "Arn": f"arn:aws:s3:::{name}",
+        "DomainName": f"{name}.s3.amazonaws.com",
+        "RegionalDomainName": f"{name}.s3.{get_region()}.amazonaws.com",
+        "WebsiteURL": f"http://{name}.s3-website-{get_region()}.amazonaws.com",
+    }
+    return name, attrs
+
+
+def _s3_update(physical_id, old_props, new_props, stack_name):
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this fix @erick-reis-gran !

PR #576, opened 2026-05-06
fix(lambda): derive AWS_ACCESS_KEY_ID from function ARN for multi-account isolation

jgrumboe:
When testing the IoT Core feature I discovered that the same Lambda function deployed to two different accounts always got the default account id.

Lambda functions deployed under non-default accounts now receive the correct AWS_ACCESS_KEY_ID (extracted from the function ARN) instead of the host process's env var. This fixes STS GetCallerIdentity and all internal SDK calls resolving to the wrong account.

Changes:
- Add _account_from_arn() helper to extract 12-digit account from ARN
- Fix all four execution paths: warm worker, provided runtime, local
  subprocess, and Docker container
- Scope warm worker pool key by account to prevent cross-account reuse
- Set _HANDLER env var for all Docker runtimes (not just provided)
- Rewrite localhost → host.docker.internal in container AWS_ENDPOINT_URL

Function-level Environment.Variables overrides still take precedence.


ministack/core/lambda_runtime.py
```diff
@@ -17,9 +17,27 @@
 import threading
 import time
 import zipfile
+import re
 
 logger = logging.getLogger("lambda_runtime")
 
+_12_DIGIT_RE = re.compile(r"^\d{12}$")
```
Nahuel990: This is duplicated from responses

jgrumboe: Good catch

jgrumboe: Fixed in last commit.

jgrumboe:
I rebased the PR on main branch instead of the IoT-phase1 branch.

Nahuel990 (approved):

PR #578, opened 2026-05-06
fix(sqs): honor MessageSystemAttributeNames in ReceiveMessage

### LINKED ISSUE #577: SQS ReceiveMessage ignores MessageSystemAttributeNames (AWS SDK v2 / Java) — system attributes returned empty
## Summary

`ReceiveMessage` does not return system attributes (e.g. `ApproximateReceiveCount`, `SentTimestamp`) when the client uses the AWS Java SDK v2 (or any client that sends the modern `MessageSystemAttributeNames` request field instead of the deprecated `AttributeNames`).

The returned `Attributes` map is empty even when the client requests `All`.

## Root cause

In `ministack/services/sqs.py`, `_act_receive_message` reads only the legacy field names:

```python
attr_names = (data.get("AttributeNames")
              or data.get("SystemAttributeNames") or [])
```

Per the official [AWS SQS service model](https://github.com/aws/aws-sdk-java-v2/blob/master/services/sqs/src/main/resources/codegen-resources/service-2.json), `ReceiveMessageRequest` has two fields:

- `AttributeNames` — marked `"deprecated": true`. Doc says: *"This parameter has been discontinued but will be supported for backward compatibility. To provide attribute names, you are encouraged to use `MessageSystemAttributeNames`."*
- `MessageSystemAttributeNames` — the current field.

The AWS Java SDK v2 method `.messageSystemAttributeNames(...)` serializes to the JSON field `MessageSystemAttributeNames` only — it does not also write to the deprecated `AttributeNames` field. Boto3 still uses the legacy param by default, which is why this hasn't been caught by the existing test suite (`tests/test_sqs.py` uses boto3 with `AttributeNames=...`).

## Reproduction (Java SDK v2)

```java
SqsClient sqs = SqsClient.builder()
    .endpointOverride(URI.create("http://localhost:4566"))
    .region(Region.EU_WEST_1)
    .build();

sqs.sendMessage(b -> b.queueUrl(url).messageBody("test"));

ReceiveMessageResponse resp = sqs.receiveMessage(b -> b
    .queueUrl(url)
    .messageSystemAttributeNames(MessageSystemAttributeName.ALL));

// Expected: attributes() contains ApproximateReceiveCount, SentTimestamp, etc.
// Actual:   attributes() is empty.
```

## Fix

One-line addition:

```python
attr_names = (data.get("AttributeNames")
              or data.get("MessageSystemAttributeNames")
              or data.get("SystemAttributeNames") or [])
```

Happy to send a PR with the fix and a regression test that asserts the modern field name is honored.

## Impact

This blocks any consumer using AWS Java SDK v2 from observing redelivery counts, sent timestamps, etc. — a common need (DLQ logic, retry detection, observability). In our case it surfaces as `IllegalArgumentException: No such key: ApproximateReceiveCount in []` from the dk.tgtg/hermes messaging library.

joaomena:
## Summary

- Adds `MessageSystemAttributeNames` to the list of fields read by `ReceiveMessage`. AWS SDK v2 clients (Java/Kotlin/etc.) send the modern field name; ministack only read the deprecated `AttributeNames` / `SystemAttributeNames`.
- Adds a regression test that uses the modern field name.

Closes #577.

## Why

The AWS SQS service model marks `AttributeNames` as `"deprecated": true` with the doc string *"This parameter has been discontinued but will be supported for backward compatibility. To provide attribute names, you are encouraged to use `MessageSystemAttributeNames`."* The Java SDK v2 method `.messageSystemAttributeNames(...)` serializes to the new field only — it does not double-write. As a result, any Java/Kotlin consumer using SDK v2 received empty `Attributes`, which broke `ApproximateReceiveCount`-based logic (redelivery detection, DLQ, etc.).

Boto3 still uses the legacy parameter name by default, so this hadn't been caught by the existing test suite.

## Test plan

- [x] New test `test_sqs_message_system_attribute_names_modern_field` asserts that `receive_message(..., MessageSystemAttributeNames=["All"])` returns `ApproximateReceiveCount` and `SentTimestamp`.
- [x] Existing `test_sqs_message_system_attributes` (legacy `AttributeNames`) continues to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **SQS `ReceiveMessage` honors `MessageSystemAttributeNames`** — only the legacy `AttributeNames` / `SystemAttributeNames` request fields were read; the AWS Java SDK v2 (and any client using the modern, non-deprecated request shape) sends `MessageSystemAttributeNames`, so `Attributes` came back empty even when the client requested `All`. Broke `ApproximateReceiveCount`-based redelivery detection and DLQ logic for SDK v2 consumers.
+- **CFN `AWS::SNS::Subscription` honors `RawMessageDelivery`** — the CloudFormation provisioner read only `FilterPolicyScope` and `FilterPolicy` from the resource properties, silently defaulting subscriptions to `RawMessageDelivery=false` even when the template explicitly set it to `true`. Consumers received SNS-wrapped envelope JSON instead of the raw message, breaking attribute-based routing for any consumer that reads SQS-level `MessageAttributes`.
+
+---
+
 ## [1.3.28] — 2026-05-05
 
 ### Added
--- a/ministack/services/cloudformation/provisioners.py
+++ b/ministack/services/cloudformation/provisioners.py
@@ -316,6 +316,8 @@ def _sns_sub_create(logical_id, props, stack_name):
         return sub_arn, {"SubscriptionArn": sub_arn}
 
     sub_arn = f"{topic_arn}:{new_uuid()}"
+    raw = props.get("RawMessageDelivery", False)
+    raw_str = "true" if (raw is True or str(raw).lower() == "true") else "false"
     sub = {
         "arn": sub_arn,
         "topic_arn": topic_arn,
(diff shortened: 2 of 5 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this @joaomena

PR #583, opened 2026-05-06
fix: Versioned s3 date headers rfc7231

mgius-ae:
Last-Modified headers are currently ISO something, and they need to be rfc7231.  Was causing issues in our tests when buckets were set to versioned.

Looks like this was already been applied in other places in s3 codebase, but the versioned part was a miss.

Added a test to stress the code path.  I also verified this branch against my product's e2e suite and the SDK date parsing error is gone.


Nahuel990:
Thanks for this @mgius-ae !

Nahuel990 (approved):

Issue #566, opened 2026-05-05
Feature: Add VPC Endpoint service support


svenikea:
### AWS Service name

VPC S3 Endpoint Service

### Operations needed

- DescribeVpcEndpointServices

### Use case

Hi, I just want to say you guys are some of the best folks out there for DevOps people like me. Ministack is great we even use it in our Kubernetes cluster to run our pipeline and execute Terragrunt plans to verify our Terraform code. We are currently using `1.3.27`

I want to create an S3 VPC Endpoint from [terraform-aws-modules/vpc/aws//modules/vpc-endpoints](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest/submodules/vpc-endpoints). Most of our cluster nodes rely heavily on S3, and having an S3 VPC gateway endpoint does provide significant performance gains. It would allow us to access S3 objects without routing traffic over the internet.

I would really appreciate the ability to test this feature.

### Would you like to contribute this?

- [ ] Yes, I'd like to implement this service and open a PR
- [ ] I can help review/test an implementation

Nahuel990:
@svenikea thanks! quick scoping: Ministack offers AWS parity at the control-plane and SDK-contract level, so Terraform and SDK calls round-trip with real AWS shapes. We do not and will not emulate VPC data-plane behavior such as packet routing, private S3 traffic via a gateway endpoint, or security groups filtering real traffic, that is out of scope by design

For your case: if the goal is Terragrunt plan and apply succeeding against MS in CI, yes, we already support CreateVpcEndpoint and DescribeVpcEndpoints, and adding DescribeVpcEndpointServices, which is the catalog lookup the module uses, is a small addition we will ship in the next release. If you also need real private routing or runtime performance semantics, that is not something we will provide

Nahuel990:
@svenikea this is available in our latest version. Please let us know if you need anything else

svenikea:
Thanks @Nahuel990 I did a test run and it shows this error 
``` 
 │ Error: flattening VPC Endpoint (vpce-8228cb6a24f4d6a73): reading EC2 Prefix List (com.amazonaws.us-east-1.s3): operation error EC2: DescribePrefixLists, https response error StatusCode: 200, RequestID: 5863fcb7-fe18-4103-9a86-4d9095cda121, deserialization failed, failed to decode response body, expected value for item element, got xml.StartElement type {{http://ec2.amazonaws.com/doc/2016-11-15/ cidr} []} instead
  │ 
[...]

PR #584, opened 2026-05-07
feat(ec2): expand AWS managed prefix lists with deterministic CIDR generation

jgrumboe:
When testing VPC endpoints of type gateway the corresponding AWS managed prefix list needs to be retrieveable. See also #566.

This PR expand AWS managed prefix lists functionality with deterministic mock to keep effort low in the future and still deliver good mock results.

- Add regex-based pattern matching for AWS managed prefix list names
- Generate deterministic prefix list IDs and CIDRs from name hashes
- Support regional and global AWS managed prefix lists (s3, dynamodb, s3express, vpc-lattice, route53-healthchecks, ec2-instance-connect, cloudfront, groundstation)
- Add IPv4 (100.64.0.0/10 CGNAT space) and IPv6 (64:ff9b:1::/48) CIDR generation
- Implement reverse lookup to resolve prefix list IDs to AWS managed names
- Update DescribePrefixLists to return deterministic CIDRs instead of 0.0.0.0/0
- Add DescribeManagedPrefixLists support for AWS managed prefix lists with filtering
- Add GetManagedPrefixListEntries to enumerate prefix list entries
- Improve prefix list filtering with prefix-list-id and owner-id support
- Add comprehensive test coverage for AWS managed prefix list operations

### DIFF
--- a/ministack/services/ec2.py
+++ b/ministack/services/ec2.py
@@ -3347,6 +3347,9 @@ def _describe_prefix_lists(p):
 
 def _create_managed_prefix_list(p):
     name = _p(p, "PrefixListName") or ""
+    # Reject creation of prefix lists that collide with AWS managed names
+    if _is_aws_managed_prefix_name(name):
+        return _error("UnsupportedOperation", "The action is not supported for an AWS-managed prefix list.", 400)
     max_entries = int(_p(p, "MaxEntries") or "10")
     af = _p(p, "AddressFamily") or "IPv4"
     pl_id = "pl-" + "".join(random.choices(string.hexdigits[:16], k=17))
--- a/ministack/services/ec2.py
+++ b/ministack/services/ec2.py
@@ -59,6 +59,7 @@ import hashlib
 import logging
 import os
 import random
+import re
 import string
 import time
 from urllib.parse import parse_qs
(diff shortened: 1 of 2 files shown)


ministack/services/ec2.py
```diff
@@ -3177,27 +3178,151 @@ def _modify_vpc_endpoint(p):
...
+_AWS_PL_GLOBAL_SUFFIXES = [
+    "cloudfront.origin-facing", "groundstation",
+    "ipv6.cloudfront.origin-facing",
+]
+
+
+def _aws_pl_id_from_name(prefix_name: str) -> str:
+    """Deterministic prefix-list ID derived from the name hash."""
+    h = hashlib.sha256(prefix_name.encode()).hexdigest()
+    return f"pl-{h[:17]}"
```
Nahuel990, line 3209: I'm from the phone I will check the PR a bit more in depth AWS managed prefix is pl-8 chars please review docs if they updated not sure about the 17

jgrumboe: It's actually a mixture of 8 and 17 chars.
image

Also, I checked several documentation pages. It seems that AWS extended from 8-chars legacy to 17-chars longer resource id format. https://docs.aws.amazon.com/cli/latest/reference/ec2/modify-id-format.html

ministack/services/ec2.py
```diff
@@ -3177,27 +3178,151 @@ def _modify_vpc_endpoint(p):
...
             continue
         if filters.get("prefix-list-name") and name not in filters["prefix-list-name"]:
             continue
+        if filters.get("prefix-list-id") and pl_id not in filters["prefix-list-id"]:
+            continue
         items += f"""<item>
             <prefixListId>{pl_id}</prefixListId>
             <prefixListName>{name}</prefixListName>
-            <cidrSet><item><cidr>0.0.0.0/0</cidr></item></cidrSet>
+            <cidrSet><item><cidr>{cidr}</cidr></item></cidrSet>
```
Nahuel990: Here, we should remove that extra <cidr> </cidr> tags

jgrumboe: Fixed in a94d9e7

Nahuel990 (approved):

PR #591, opened 2026-05-08
Add CORS headers to Cognito OAuth2 / OIDC endpoints

coezbek:
## Summary

Cognito's browser-facing endpoints — Hosted UI (`/oauth2/authorize`, `/login`), `/oauth2/token`, `/oauth2/userInfo`, `/logout`, `/.well-known/openid-configuration`, `/.well-known/jwks.json` — are dispatched in `app.py` via `_handle_cognito_get_request` and `_handle_cognito_body_request`. Both return raw `(status, headers, body)` tuples that bypass `_with_data_plane_headers`, the helper every other data-plane response goes through and which adds the wildcard `Access-Control-Allow-Origin: *` header that browser-based OIDC clients require.

The `OPTIONS` preflight handler does send CORS headers, so preflight succeeds, but the actual response is missing the header and the browser blocks it. Symptom: any `react-oidc-context` / `oidc-client-ts` / Amplify-style OIDC client running against MiniStack fails on discovery and on `/oauth2/token` with:

> No 'Access-Control-Allow-Origin' header is present on the requested resource.

## Fix

Route both Cognito dispatchers through `_with_data_plane_headers` at their call sites in `_handle_pre_body_request` and `_handle_post_body_shortcuts`. Same wildcard CORS that every other data-plane response gets — no new CORS logic, just plugging the gap.

Threads `request_id` through `_handle_post_body_shortcuts` so the helper can also stamp `x-amzn-requestid` alongside the CORS header, matching every other data-plane response.

## Test plan

- [x] New regression test (`test_cognito_browser_endpoints_send_cors_headers`) asserts `Access-Control-Allow-Origin: *` on `/<pool>/.well-known/openid-configuration`, `/<pool>/.well-known/jwks.json`, and `/oauth2/token`.
- [x] Full `tests/test_cognito.py` suite passes (101/101).
- [x] No other test files reference these dispatch points; threading `request_id` through `_handle_post_body_shortcuts` is purely additive.

## Notes

- Independent of #588 (which fixes the discovery doc's URLs to point at MiniStack instead of real AWS); the two changes compose — discovery returns local URLs, and now the responses to those URLs carry CORS headers.
- The form action in `_login_page_html` is left as `/login` (absolute path on origin). Since the Hosted UI is served from MiniStack's own origin, that resolves correctly. Reverse-proxying MiniStack at a path prefix (e.g. `/cognito/oauth2/authorize`) would break it, but with CORS now working there's no reason to proxy in the first place.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -662,7 +662,11 @@ async def _handle_pre_body_request(method: str, path: str, headers: dict, query_
 
     response = await _handle_cognito_get_request(method, path, headers, query_params)
     if response is not None:
-        return response
+        # Cognito's OAuth2/OIDC endpoints (Hosted UI, /oauth2/*, /.well-known/*)
+        # are typically called by browser-based OIDC clients and must therefore
+        # carry the same `Access-Control-Allow-Origin: *` that every other data
+        # plane response gets via _with_data_plane_headers.
+        return _with_data_plane_headers(response, request_id)
 
     response = await _handle_ses_messages_request(method, path, headers, query_params)
     if response is not None:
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks for this @coezbek sorry for the delay we will be releasing this in the new version today

Nahuel990 (approved):

PR #604, opened 2026-05-09
feat(lambda): inject @aws-sdk/client-* stubs into Node.js worker for CDK Provider Framework compatibility

hiddengearz:
## Summary

> ⚠️ **Stacked on #603** (`feat/cfn-custom-resource`) — the diff here includes those CFN custom resource changes plus the additions below. Please review and merge #603 first; this PR's base will stay as `main` for GitHub's sake but the lambda-only delta lives in `feat/cfn-custom-resource..feat/lambda-node-sdk`.

Real AWS Lambda (Node.js 18+) ships the AWS SDK v3 built-in, so handlers can `require('@aws-sdk/client-ssm')` without bundling it. The Ministack Node.js worker runs on the host, where these packages are not installed — causing `Cannot find module '@aws-sdk/client-ssm'` errors that prevent any CDK custom resource from running.

This PR patches `Module.prototype.require` in the worker script to intercept `@aws-sdk/client-*` requires and return lightweight stubs that route through `AWS_ENDPOINT_URL` (Ministack). Real packages take precedence if present (e.g. via Lambda Layers), so the stub is only the fallback.

- **Lambda** (`@aws-sdk/client-lambda`): dedicated REST stub — `POST /2015-03-31/functions/{name}/invocations`. Exports `Lambda` (v2-style `.invoke()`), `LambdaClient` + `InvokeCommand` (v3-style `.send()`), and `waitUntilFunctionActiveV2` (no-op — functions are always active in Ministack)
- **All other JSON-protocol services**: generic Proxy-based stub — intercepts any property access, maps `*Client` → v3-style client, `*Command` → command class, bare uppercase name → v2-style client. Routes via `X-Amz-Target` to Ministack's existing service handlers. Covers all services in `router.py` that use `target_prefixes` (SSM, SFN, STS, DynamoDB, Kinesis, ECS, CloudWatch, Logs, SecretsManager, EventBridge, SQS, SNS, Glue, Athena, Firehose, Cognito IDP/Identity, EMR, ECR, ACM, WAFv2, WAF, Organizations, KMS, CodeBuild, Transfer, ServiceDiscovery, Resource Groups Tagging, CloudTrail)
- **HTTPS → HTTP downgrade**: CDK Provider Framework's `cfn-response.js` calls `https.request` unconditionally for the ResponseURL PUT and drops the port from the URL. Extended the existing `patchAwsSdk()` interceptor to catch localhost/127.0.0.1 requests, downgrade to HTTP, and restore the missing port — so the PUT reaches Ministack's HTTP server
- **`err.name` parity**: AWS SDK v3 sets both `err.name` and `err.code` to the service error type string (e.g. `"ParameterNotFound"`). Handlers typically catch by `e.name` — the stub now sets both fields

## Files changed

| File | Change |
|------|--------|
| `ministack/core/lambda_runtime.py` | `_installAwsSdkV3Stubs()` IIFE: `Module.prototype.require` intercept, Lambda REST stub, generic JSON-RPC stub factory, `_JSON_RPC_TARGETS` map, localhost HTTPS→HTTP patch |
| `tests/test_lambda.py` | 4 new Node.js worker tests: stub resolution (Lambda/SFN/SSM), extended service resolution (14 additional services), `err.name` parity, HTTPS→HTTP downgrade |

## Test plan

- [x] `pytest tests/test_lambda.py -k "nodejs_worker" -v` — 4/4 pass
  - `test_nodejs_worker_aws_sdk_v3_stub_resolves` — Lambda, LambdaClient, InvokeCommand, waitUntilFunctionActiveV2, SFNClient, SSMClient, PutParameterCommand all resolve
  - `test_nodejs_worker_aws_sdk_v3_stub_resolves_extended` — 14 additional services all resolve (SQS, SNS, KMS, Cognito IDP/Identity, ECR, Glue, Athena, Firehose, ACM, Organizations, CodeBuild, CloudTrail, ServiceDiscovery)
  - `test_nodejs_worker_json_rpc_error_has_name` — `err.name === 'ParameterNotFound'` (not `'Error'`)
  - `test_nodejs_worker_https_localhost_downgraded_to_http` — `https.request` to localhost reaches a real HTTP server
- [x] `pytest tests/test_cfn_custom_resource.py -v` — 11/11 pass (no regression)
- [x] End-to-end: CDK stack deploying a `cr.Provider`-backed custom resource that calls SSM `GetParameter` / `PutParameter` deploys successfully against Ministack

## Notes

**Services not covered by the generic stub** (no `X-Amz-Target` protocol): S3, CloudFront, Route53 (REST-XML), IAM, EC2, RDS, ElastiCache, AutoScaling (Query API), and REST-path services like API Gateway, AppConfig, Scheduler, EKS, AppSync, Backup, RDS Data API. Handlers using those services need the real SDK bundled or provided via a Lambda Layer — the same constraint as real AWS when running outside a managed runtime.


Nahuel990:
@hiddengearz can you review the docs from where you are implementing this? Short prefix `AWSSecurityTokenService.` does not match, so the JSON branch is skipped. The JSON body is then parsed as form-encoded and yields no `Action` → `400`.

Per botocore (`get_service_model("sts").metadata`), the real STS protocol is `query`, not `json`. The AWS SDK v3 package `@aws-sdk/client-sts` also wire-formats requests as Query protocol.

So the current stub matches neither real AWS behavior nor MiniStack.

hiddengearz:
Good catch — confirmed. Auditing all entries via `botocore.session.get_service_model(svc).protocol` found the same issue in two more:

| Entry | Actual protocol | Real `@aws-sdk` wire format |
|---|---|---|
| `sts` | `query` | `Action=GetCallerIdentity` (form-encoded POST) |
| `sns` | `query` | `Action=Publish` (form-encoded POST) |
| `cloudwatch` | `smithy-rpc-v2-cbor` | `POST /service/GraniteServiceVersion20100801/operation/PutMetricData` |

All other entries are correctly `json`. `sqs` is fine — botocore also reports it as `json` (`@aws-sdk/client-sqs` v3 uses `X-Amz-Target: AmazonSQS.*`).

**Fix** — removed all three ([e9ece8f](https://github.com/hiddengearz/ministack/commit/e9ece8f)):

```diff
-    "sts":        "AWSSecurityTokenService",
-    "cloudwatch": "GraniteServiceVersion20100801",
-    "sns":        "AmazonSNS",
```

When the real package is present it formats requests correctly via its native protocol. When absent, the right failure is `Cannot find module` — not a wrong-format request.

**Why tests didn't catch it:** Ministack's SNS handler is query-only (no JSON branch), so the SNS stub was broken against Ministack too — but SNS and STS were never exercised in the test suite. CloudWatch was a silent false pass: Ministack's CloudWatch handler explicitly accepts X-Amz-Target JSON alongside smithy-rpc-v2-cbor, so the wrong wire format happened to work locally.

Does this look correct to you?

Nahuel990 (approved):

Nahuel990:
Awesome! Thanks @hiddengearz

PR #605, opened 2026-05-09
Echo OAuth2 nonce into Cognito id_token

coezbek:
## Summary

OIDC requires the id_token to carry the nonce the client sent in the authorize request, so the client can mitigate replay. MiniStack already captures `nonce` on the authorize form and stores it on the auth code (`_authorization_codes[code][\"nonce\"]`) — but the token endpoint never threads it into `_fake_token`, so the minted id_token omits the `nonce` claim.

Strict OIDC libraries (`oidc-client-ts`, `react-oidc-context`, Auth0/MS libs) silently discard tokens missing an expected nonce. Symptom: `/oauth2/token` returns 200 with `id_token` + `access_token`, but the OIDC client never finishes loading and downstream API calls go out without an `Authorization` header.

## Fix

Thread `nonce` through `_build_auth_result` → `_fake_token` and stamp `claims[\"nonce\"]` on id tokens when present. No change to access or refresh tokens.

## Test plan

- [x] New regression test (`test_oauth2_id_token_echoes_nonce`) drives a login with a known nonce and asserts the id_token carries it.
- [x] Full `tests/test_cognito.py` suite passes (102/102).
- [x] Default `nonce=\"\"` preserves all existing callers.

### DIFF
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -337,6 +337,7 @@ def _identity_id(pool_id: str) -> str:
 def _fake_token(sub: str, pool_id: str, client_id: str, token_type: str = "access",
                  username: str = "", user_attrs: dict | None = None,
                  groups: list[str] | None = None,
+                 nonce: str = "",
                  trigger_source: str = "TokenGeneration_Authentication") -> str:
     """Return a JWT signed with the RSA key when cryptography is available.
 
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks @coezbek

PR #609, opened 2026-05-11
feat(iam): emulate AWS-managed policies (arn:aws:iam::aws:policy/*) globally

spicykay:
## Problem

Real AWS hosts AWS-managed policies (`AdministratorAccess`, `ReadOnlyAccess`, `AWSLambdaBasicExecutionRole`, etc.) under a **virtual `aws` account** that every customer can read regardless of their own session account. Ministack stores every policy in `AccountScopedDict`, keyed by `(get_account_id(), arn)` — so a normal 12-digit-numeric session can never resolve them. `GetPolicy(arn:aws:iam::aws:policy/AdministratorAccess)` returns `NoSuchEntity`, which breaks the very common Terraform pattern:

```hcl
data "aws_iam_policy" "admin" {
  arn = "arn:aws:iam::aws:policy/AdministratorAccess"
}
```

…and every other consumer that references AWS-managed ARNs by ARN (`iam.ManagedPolicy.fromAwsManagedPolicyName` in CDK, raw boto3, Serverless framework defaults, the AWS Lambda execution-role bundles, etc.).

`_list_policies` already string-matches `arn:aws:iam::aws:` (https://github.com/ministackorg/ministack/blob/main/ministack/services/iam.py#L439) — proof that the concept of "AWS-managed = not Local" is partially modeled, just not made readable.

## Repro (without this PR)

```bash
docker run -d --name ms -p 4566:4566 ministackorg/ministack:latest
AWS_ACCESS_KEY_ID=481530559488 AWS_SECRET_ACCESS_KEY=test AWS_REGION=us-east-1 \
  aws --endpoint-url http://localhost:4566 iam get-policy \
  --policy-arn arn:aws:iam::aws:policy/AdministratorAccess
# An error occurred (NoSuchEntity) when calling the GetPolicy operation:
# Policy arn:aws:iam::aws:policy/AdministratorAccess not found.
```

Same thing fails for `arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole`, etc.

Workaround we've been shipping downstream: an HTTP proxy in front of Ministack that rewrites `arn:aws:iam::aws:policy/X` → `arn:aws:iam::<session>:policy/X` in IAM request bodies, plus a startup script that seeds matching customer-managed policies under the session account. Roughly 150 lines of Go + bash, and it shows the rewritten ARN in plan output (cosmetic but jarring). With this PR landed, that whole layer goes away.

## Fix

Model AWS-managed policies as a *separate*, non-account-scoped global store and pre-seed it with the most commonly referenced policies. Helpers `_lookup_policy(arn)` / `_policy_exists(arn)` fall through to the global store when the ARN matches `arn:aws:iam::aws:policy/`. Read paths use the helpers; mutation paths reject AWS-managed ARNs with AWS-parity errors.

### Pre-seeded policies

The four job-function policies (`AdministratorAccess`, `PowerUserAccess`, `ReadOnlyAccess`, `SecurityAudit`) plus the service-specific policies bundled with most Terraform modules and CDK constructs:

`IAMFullAccess`, `IAMReadOnlyAccess`, `AmazonS3FullAccess`, `AmazonS3ReadOnlyAccess`, `AmazonEC2FullAccess`, `AmazonEC2ReadOnlyAccess`, `AmazonSSMManagedInstanceCore`, `AmazonDynamoDBFullAccess`, `AWSLambdaBasicExecutionRole`, `AWSLambdaVPCAccessExecutionRole`, `AmazonSQSFullAccess`, `AmazonSNSFullAccess`, `AmazonECSTaskExecutionRolePolicy`, `CloudWatchAgentServerPolicy`, `CloudWatchLogsFullAccess`, `AWSCloudFormationFullAccess`.

Documents are minimal but accurate stubs — emulate is for plan-time iteration, not for matching every policy statement against real AWS. Happy to expand or trim the seed list based on review feedback.

### Autovivify for the long tail

For AWS-managed ARNs we don't pre-seed (~1000 exist in real AWS), `GetPolicy` autovivifies a permissive stub on first lookup so terraform plans still complete. This matches Ministack's permissive-emulator philosophy. Strict mode (return `NoSuchEntity` for unseeded ARNs) is opt-in via `MINISTACK_AUTOCREATE_AWS_MANAGED=0`.

### Mutation rejected with AWS-parity errors

| Operation on `arn:aws:iam::aws:policy/X` | Response |
|---|---|
| `CreatePolicy` (somehow targeting the `aws` namespace) | `InvalidInput` |
| `DeletePolicy` | `AccessDenied` |
| `TagPolicy` / `UntagPolicy` | `AccessDenied` |
| `CreatePolicyVersion` / `DeletePolicyVersion` | `AccessDenied` |
| `AttachRolePolicy` / `AttachUserPolicy` | succeeds (AttachmentCount counter is no-op'd) |
| `DetachRolePolicy` / `DetachUserPolicy` | succeeds |
| `ListPolicies Scope=AWS / All` | included |
| `ListPolicies Scope=Local` | excluded |

### Reset

`reset()` re-seeds AWS-managed policies after clearing — they're not customer state, so the canonical set should always be present.

## Verify (with this PR)

```bash
AWS_ACCESS_KEY_ID=481530559488 AWS_SECRET_ACCESS_KEY=test AWS_REGION=us-east-1 \
  aws --endpoint-url http://localhost:4566 iam get-policy \
  --policy-arn arn:aws:iam::aws:policy/AdministratorAccess
# {
#     "Policy": {
#         "PolicyName": "AdministratorAccess",
#         "Arn": "arn:aws:iam::aws:policy/AdministratorAccess",
#         "DefaultVersionId": "v1",
#         "IsAttachable": true,
#         "Description": "Provides full access to AWS services and resources.",
#         ...
#     }
# }
```

## Tests

11 new cases in `tests/test_iam.py`:

- `test_iam_get_aws_managed_policy_administrator_access`
- `test_iam_get_aws_managed_policy_version_returns_document`
- `test_iam_list_policies_scope_all_includes_aws_managed`
- `test_iam_list_policies_scope_aws_returns_only_aws_managed`
- `test_iam_list_policies_scope_local_excludes_aws_managed`
- `test_iam_attach_aws_managed_policy_to_role`
- `test_iam_cannot_create_policy_under_aws_account`
- `test_iam_cannot_delete_aws_managed_policy`
- `test_iam_cannot_tag_aws_managed_policy`
- `test_iam_unknown_aws_managed_policy_is_autocreated`

All 44 tests in `tests/test_iam.py` pass (33 existing + 11 new).

```
============================== 44 passed in 0.42s ==============================
```

`ruff check ministack/services/iam.py` clean.

## Discussion points

- **Seed list size.** 20 felt like the right cutoff (Pareto on what real-world Terraform / CDK references), but happy to trim down to the four job-function policies if reviewers prefer a smaller surface — the autovivify path covers everything else.
- **Autovivify default.** I defaulted it **on** because Ministack's other services lean permissive (e.g., `SimulatePrincipalPolicy` always returns `allowed`), so plans-complete-by-default felt consistent. Easy to flip to strict-by-default with one line if maintainers disagree.

## Checklist

- [x] Tests added and passing (`pytest tests/test_iam.py -v` — 44 passed)
- [x] Linting passes (`ruff check ministack/services/iam.py`)
- [x] No new service file (this is a bug fix / behavioral feature on the existing `iam.py`)


Nahuel990:
Hey @spicykay, thanks for this. The overall PR looks good

I want to push back a bit before merging though, because MiniStack is an emulator and the standard has to be: "if it works here, it works on real AWS". We should avoid deviations that mask real AWS behaviour locally (one of the biggest problem LocalStack has)

The autovivify default is the main concern.

If someone typos an ARN, for example:

    arn:aws:iam::aws:policy/AdminstratorAccess

(missing the `i`), real AWS returns `NoSuchEntity` and the user fixes it.

With autovivify enabled by default, MiniStack silently invents a permissive policy under the typoed ARN, the `terraform plan` succeeds, and the issue only appears later against real AWS.

Could you flip the default to OFF?

The 20 pre-seeded policies already cover the common case, and users who genuinely want "on demand" behaviour for longpolicies can opt in. The autocreate test will need to change with that default.


 AttachmentCount stays at0 for AWS-managed policies, but real AWS reports the count of the calling account's roles/users that have the policy attached. Terraform's `aws_iam_policy` data source reads that  so if you attach `AdministratorAccess` to three of your roles and ask MiniStack "how many use this?", you get zero. Audit tools that key off that field will see the wrong picture.

The **seeded documents** are functional approximations rather than the canonical AWS documents only `AdministratorAccess` matches exactly. Keep the inline shape you already have in `_seed_aws_managed_policies`; just swap each stub for the canonical document. You can pull each one with `aws iam get-policy-version`.

Heads-up on size: most documents are small (300 bytes–2 KB) but `ReadOnlyAccess` and `SecurityAudit` are ~22 KB and ~14 KB respectively because they enumerate a lot of actions across every AWS service. All 20 inline will land around 60–80 KB total (we are good to keep them in one file)

Happy to merge after that

jgrumboe:
Just thinking loud here: could we probably use https://aws.permissions.cloud/managedpolicies to verify permissions existence or even fetching on the fly? Maybe behind some feature flag?

Nahuel990:
@jgrumboe we should keep Ministack offline, without network calls that can limit the rate of the container/package

Nahuel990 (approved):

PR #615, opened 2026-05-11
Add Cost Usage Reports Service

staranto:
This PR adds Cost Usage Reports (cur) functionality to MiniStack. 7 of 7 [API actions](https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_Operations_AWS_Cost_and_Usage_Report_Service.html) and CLI commands are supported. Full support for Terraform AWS Provider 6.44.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Added
+- **Cost & Usage Reports** — 7 of 7 CUR API actions supported.
+
 ## [1.3.34] — 2026-05-11
 
 ### Added
@@ -413,7 +418,7 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 ## [1.3.6] — 2026-04-20
 
 ### Added
-- **API Gateway path-based data plane** — REST + HTTP + WebSocket APIs are now reachable without `*.execute-api.localhost` Host overrides: `http(s)://localhost:4566/_aws/execute-api/{apiId}/{stage}/{path}` (v1 + v2 HTTP + v2 WS) and the LocalStack-legacy `http://localhost:4566/restapis/{apiId}/{stage}/_user_request_/{path}` (v1). Unblocks macOS browsers (no `*.localhost` DNS resolution) and strict HTTP clients with no Host override. 
+- **API Gateway path-based data plane** — REST + HTTP + WebSocket APIs are now reachable without `*.execute-api.localhost` Host overrides: `http(s)://localhost:4566/_aws/execute-api/{apiId}/{stage}/{path}` (v1 + v2 HTTP + v2 WS) and the LocalStack-legacy `http://localhost:4566/restapis/{apiId}/{stage}/_user_request_/{path}` (v1). Unblocks macOS browsers (no `*.localhost` DNS resolution) and strict HTTP clients with no Host override.
 - **Custom/predictable API Gateway IDs** — `aws_apigatewayv2_api` and `aws_apigateway_rest_api` honour an `ms-custom-id` tag on `CreateApi` / `CreateRestApi` and pin the generated `apiId` / REST API id to the tag value. Duplicates in the same account return `ConflictException` (409). The LocalStack `ls-custom-id` tag is intentionally rejected with a clear `BadRequestException` (400) pointing callers at the ministack-native key. Reported by @whittin3. Fixes #400
 - **Cognito `AWS::Cognito::UserPoolClient` CFN `GenerateSecret`** — CloudFormation-provisioned user pool clients now generate a client secret when `GenerateSecret: true`, matching the native Cognito API path. Contributed by @mgius-ae (#403)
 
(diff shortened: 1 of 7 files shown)


Nahuel990:
@staranto thanks for the PR. The implementation is clean and it covers the full API surface.
Before I merge it, could you share the use case you have?

Real cost data only exists on AWS itself, so I want to better understand how this would be used within MiniStack.

staranto:
@Nahuel990 my use case doesn't have any interest in actual cost data. I have a large environment with many AWS accounts and a ton of TF and AWS CLI via Bash scripts to support that. I think nearly all of the accounts use CUR (amongst, obviously, many other services), so it is managed in the TF and scripts. 

My use of MiniStack is basically the "AWS Testing 101" use case from your Getting Started. So, for CUR, it doesn't matter if there's any real data in the report, just that the TF and scripts run to completion and do what they're expected. Lack of CUR support in MiniStack seemed to be an easy first PR to close the gap, which is already quite small anyway. 

Without CUR support, this fails. With it, we're set.
```
resource "aws_cur_report_definition" "example_cur_report_definition" {
  report_name                = "example-cur-report-definition"
  time_unit                  = "HOURLY"
  format                     = "textORcsv"
  compression                = "GZIP"
  additional_schema_elements = ["RESOURCES", "SPLIT_COST_ALLOCATION_DATA"]
  s3_bucket                  = "example-bucket-name"
  s3_prefix                  = "example-cur-report"
  s3_region                  = "us-east-1"
  additional_artifacts       = ["REDSHIFT", "QUICKSIGHT"]
}
```

Nahuel990:
@staranto  merging this since the implementation is already done.

For future PRs though: AWS has 250+ services and botocore exposes nearly 400 endpoints. "AWS has it" or "it is easy to stub" is not enough reason on its own to add support in MiniStack. If we follow that rule consistently, the project becomes impossible to maintain.

Please focus contributions on services with an actual use case behind them.

Concrete example from this PR: the report definition declares:

    additional_artifacts = ["REDSHIFT", "QUICKSIGHT"]

MiniStack does not emulate Redshift or QuickSight either, so the CUR support here mostly just lets the Terraform apply finish successfully. There is no real path for the generated data to go anywhere useful afterwards.

That is still valid for lightweight stack validation or testing examples, but it also limits the practical value of the feature. That is the kind of trade-off contributors should think about before opening a PR.

Nahuel990 (approved):

staranto:
Couldn't "AWS has it" be a worthy *goal* by itself, when applied to the control-plane API for any service? It seems that would be more maintainable across the entire landscape.  If a new service is to be added to MiniStack, step 1 is to implement the control-plane.  generally far easier, I would think, and a requirement anyway.  The *optional* steps 2-x are to implement the functional-plane.

IMO.. the value of MiniStack is not only functional emulation.  A major value is just control-plane compatibility through the public APIs, which is required for Terraform and CLI workflows.  If a public API is missing, those workflows fail immediately, even when no one expects deep functional emulation "inside" that service.

Though CUR may not be the best general example for this, it is for my use case.  Without the 7 public CUR actions, Terraform and AWS CLI scripts fail.  With those actions, workflows against MiniStack become testable again.

For CUR specifically, the public API is control-plane focused.  The real "meat" is AWS’s internal billing and report-generation functionality, which is not publicly exposed.  Requiring that internal functionality as the bar for a service would set a much higher standard than MiniStack currently applies to multiple existing services that are intentionally control-plane stubs.

The CUR contribution followed the existing project pattern:

- Implement the publicly exposed control-plane API first so IaC and CLI workflows succeed.

- Add deeper behavior incrementally, if needed.

If the project wants a new standard, for example "no new service unless it includes functionality beyond the public API control-plane", that is valid.  But it should be explicit and consistent.

> MiniStack does not emulate Redshift or QuickSight either, so CUR support here mostly lets Terraform apply complete successfully.

Yes, that's the point.  This example was taken directly from Terraform documentation.  The control-plan works perfectly, the functional-plane does not.  If it needs to, that's an entirely more complex scope.

> There is no real path for generated data to go anywhere useful afterwards.

Yes. For many services and many use cases, that's fine.

Nahuel990:
@staranto for context, MiniStack did not start as "an emulator for every AWS service."

It started as a free replacement for the old LocalStack community edition after they sunset it, plus a handful of services LocalStack moved behind the paid tier that people still needed locally.

That scope is intentional. It is what keeps the project lightweight:

- ~270MB image size
- single container
- fully offline

The trade-off is that every service we add becomes weight every user carries, including the majority who will never call that API.

botocore exposes roughly 400 endpoints. If the standard becomes "Terraform breaks without it", then eventually every service qualifies and the image grows forever.

The bar has to stay tighter than that:

- services that existed in the LocalStack community surface
- services LocalStack moved behind paywalls
- services with enough functional value that local emulation is genuinely useful

Things like:

- S3 object round-tripping
- DynamoDB queries
- Lambda execution
- SQS delivery

Those are meaningfully better to run locally than against real AWS.

CUR is borderline on that scale. I merged it as an exception because the implementation was clean and it was already done.

But it also sets the limit for what we want to accept in the future as "control-plane only with no functional behaviour behind it." MiniStack won't turn into a collection of stub endpoints just because Terraform references them.

PR #619, opened 2026-05-11
feat(s3): enforce Deny statements in bucket policy

mfurqaan31:
## Summary

  S3 bucket policies were stored on `PutBucketPolicy` and round-tripped on `GetBucketPolicy`, but never evaluated against incoming requests. As a result every request succeeded regardless of policy — `Effect: Deny`
   statements were silently ignored. Tests written against MiniStack to verify "deny-all" or "deny `s3:GetObject` on this prefix" patterns would pass locally and ship buckets that are wide open in prod.

  ## Fix

  Add a minimal policy evaluator and wire it into the request pipeline:

  - `_check_bucket_policy` (`ministack/services/s3.py`, after `_delete_bucket_policy`) — parses the stored JSON and returns `AccessDenied` (HTTP 403) when any `Effect: Deny` statement matches the request on both
  `Action` (with `fnmatch` wildcards, e.g. `s3:*`) and `Resource` (bucket and object ARN forms). Conditions, Principal scoping, and `Effect: Allow` are intentionally not implemented — bucket-owner default-allow
  remains in force for everything not explicitly denied.
  - `handle_request` (`ministack/services/s3.py`) — invokes `_check_bucket_policy` before `_dispatch`, with an `action_map` covering `GET/HEAD → s3:GetObject|s3:ListBucket`, `PUT → s3:PutObject|s3:CreateBucket`,
  `POST → s3:PutObject`, `DELETE → s3:DeleteObject|s3:DeleteBucket`.
  - Escape hatch: requests carrying `?policy` (`Get/Put/DeleteBucketPolicy`) skip enforcement so a deny-all does not brick the bucket — without this, a policy applied through the API could not be inspected or
  removed via the API.

  ## Test added

  `tests/test_s3.py` — 9 new tests plus a `_deny_all_policy` helper:

  - `test_s3_policy_deny_blocks_get_object` / `_head_object` / `_list_objects` / `_put_object` / `_delete_object` — deny-all blocks each verb with `AccessDenied` (HTTP 403)
  - `test_s3_policy_deny_scoped_to_specific_action` — `Deny s3:GetObject` does not block PutObject
  - `test_s3_policy_deny_scoped_to_specific_object` — `Deny ... bkt/private/*` does not block `bkt/public/*`
  - `test_s3_policy_management_escapes_deny` — `Get/Put/Delete-BucketPolicy` keep working under deny-all
  - `test_s3_policy_allow_statements_are_ignored` — documents the intentional Deny-only scope
  - `test_s3_policy_malformed_json_does_not_crash_evaluator` — fail-open on garbage policy

  The existing `test_s3_bucket_policy` (round-trip only) is unchanged and continues to pass.

  Verified locally:

  pytest tests/test_s3.py -v -k policy
  ...
  tests/test_s3.py::test_s3_bucket_policy PASSED
  tests/test_s3.py::test_s3_policy_deny_blocks_get_object PASSED
  tests/test_s3.py::test_s3_policy_deny_blocks_head_object PASSED
  tests/test_s3.py::test_s3_policy_deny_blocks_list_objects PASSED
  tests/test_s3.py::test_s3_policy_deny_blocks_put_object PASSED
  tests/test_s3.py::test_s3_policy_deny_blocks_delete_object PASSED
  tests/test_s3.py::test_s3_policy_deny_scoped_to_specific_action PASSED
  tests/test_s3.py::test_s3_policy_deny_scoped_to_specific_object PASSED
  tests/test_s3.py::test_s3_policy_management_escapes_deny PASSED
  tests/test_s3.py::test_s3_policy_allow_statements_are_ignored PASSED
  tests/test_s3.py::test_s3_policy_malformed_json_does_not_crash_evaluator PASSED

  ## References

  - AWS bucket policy examples — https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html

### DIFF
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -25,6 +25,7 @@ Storage: In-memory (optionally backed by S3_DATA_DIR).
 import base64
 import copy
 import datetime as _dt
+import fnmatch
 import hashlib
 import json
 import logging
@@ -37,7 +38,6 @@ from urllib.parse import quote as url_quote
 from urllib.parse import unquote as url_unquote
 from xml.etree.ElementTree import Element, SubElement, tostring
 from xml.sax.saxutils import escape as _esc
-import fnmatch
 
 from defusedxml.ElementTree import fromstring
 
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -37,6 +37,7 @@ from urllib.parse import quote as url_quote
 from urllib.parse import unquote as url_unquote
 from xml.etree.ElementTree import Element, SubElement, tostring
 from xml.sax.saxutils import escape as _esc
+import fnmatch
 
 from defusedxml.ElementTree import fromstring
 
@@ -408,6 +409,27 @@ async def handle_request(
 ) -> tuple:
     bucket, key = _parse_bucket_key(path, headers)
 
+    # Enforce bucket-policy Deny statements before dispatch.
+    if bucket:
+        # Policy management always escapes — otherwise a deny-all bricks the bucket.
+        if "policy" in query_params and method in ("GET", "PUT", "DELETE"):
+            action = None
+        else:
+            action_map = {
+                "GET":    "s3:GetObject" if key else "s3:ListBucket",
+                "HEAD":   "s3:GetObject" if key else "s3:ListBucket",
+                "PUT":    "s3:PutObject" if key else "s3:CreateBucket",
+                "POST":   "s3:PutObject",
+                "DELETE": "s3:DeleteObject" if key else "s3:DeleteBucket",
+            }
+            action = action_map.get(method)
+        if action:
+            denied = _check_bucket_policy(bucket, action, key)
+            if denied:
+                status, h, body_out = denied
+                h.setdefault("x-amz-request-id", new_uuid())
+                return status, h, body_out
+
     result = _dispatch(method, bucket, key, headers, body, query_params)
 
     status, resp_headers, resp_body = result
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks @mfurqaan31. We appreciate the work here and the test coverage, but it can't be merged in its current form because of how it behaves with real-world bucket policies.

The main issue is `Condition`.

Most production S3 deny policies are conditional:

- deny unless `SecureTransport`
- deny unless encryption headers are present
- deny unless source VPC matches
- deny unless IP is within a CIDR range

Real AWS evaluates the `Condition` block and only applies the deny when the condition matches. This PR ignores conditions entirely, so every conditional deny effectively becomes unconditional.

That breaks common setups immediately. For example, a bucket policy enforcing HTTPS-only access works correctly on AWS today, but with this PR every `GetObject` request returns `403` locally regardless of transport.

The same problem exists with `Principal` / `NotPrincipal`. A deny intended for one account ends up applying to everyone here.

There is also an issue with the `action_map` collapsing distinct actions like:

- `s3:GetBucketTagging`
- `s3:GetBucketVersioning`
- `s3:ListBucket`

into the same internal permission string.

To merge this properly, the evaluator would need real `Condition` support, at least:

- `Bool`
- `StringEquals`
- `StringNotEquals`
- `IpAddress`
- `ArnLike`

and a proper S3 action resolution table covering the actual API surface, which are 181 actions (https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html) rather than 5

That is a much larger piece of work.

The alternative is to keep bucket policies as opaque storage only. `terraform apply` for `aws_s3_bucket_policy` already round-trips correctly today, which is enough for IaC validation without introducing incorrect authorization behaviour.

mfurqaan31:
Sure @Nahuel990 you can close the PR

PR #622, opened 2026-05-11
fix(s3): UploadPartCopy x-amz-copy-source-range validation

mfurqaan31:
## Summary

  `UploadPartCopy` parses `x-amz-copy-source-range` with `start, end = rng.split("-")` and no validation. Malformed values raise an unhandled `ValueError`, which surfaces as HTTP 500 with an empty body.
  Out-of-bounds and reversed ranges pass silently and corrupt the part body.

  Observed behavior:

  | Input | Expected | Actual |
  |---|---|---|
  | `bytes=garbage` | 400 `InvalidArgument` | 500 (unhandled `ValueError`) |
  | `bytes=abc-def` | 400 `InvalidArgument` | 500 (unhandled `ValueError`) |
  | `bytes=5-2` (reversed) | 400 `InvalidArgument` | 200 (empty slice) |
  | `bytes=0-99` on 10-byte source | 400 `InvalidArgument` | 200 (silently truncated to 10 bytes) |

  The silent cases are the worst — `CompleteMultipartUpload` succeeds and the assembled object is wrong size, with no error anywhere in the client.

  boto3 retries 5xx but fails fast on 4xx, so clients hit infinite retry loops against MiniStack while failing immediately against real S3 — masking the underlying client bug and hanging CI.

  ## Fix

  Validate `x-amz-copy-source-range` in `_upload_part_copy` (`ministack/services/s3.py`) before slicing:

  - require `bytes=` prefix
  - reject multi-range values (containing `,`)
  - require exactly two non-empty segments
  - require both segments to be integers
  - require `start >= 0` and `end >= start`
  - require the range to fit inside the source object

  Any failure returns 400 `InvalidArgument` with the source object size in the message when applicable. Valid ranges and requests without the header are unchanged.

  ## Test added

  `tests/test_s3.py`:

  - `test_s3_upload_part_copy_with_valid_range` — regression check for the valid-range happy path
  - `test_s3_upload_part_copy_rejects_malformed_range` — parametrized over 10 malformed inputs (no-hyphen, non-numeric, too many segments, both-empty, missing-start, missing-end, reversed, multi-range, wrong unit,
  missing prefix); each must return 400 `InvalidArgument`
  - `test_s3_upload_part_copy_rejects_out_of_bounds_range` — range past the end of the source returns 400 with the object size in the message

  The existing `test_s3_upload_part_copy` happy-path test is unchanged.

  Verified locally:

  pytest tests/test_s3.py -k upload_part_copy -v
  ...
  tests/test_s3.py::test_s3_upload_part_copy PASSED
  tests/test_s3.py::test_s3_upload_part_copy_with_valid_range PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=garbage] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=abc-def] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=10-20-30] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=-] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=5-] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=-5] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=5-2] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[bytes=0-1,3-4] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[rows=0-5] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_malformed_range[0-5] PASSED
  tests/test_s3.py::test_s3_upload_part_copy_rejects_out_of_bounds_range PASSED

  13 passed

  ## References

  - `UploadPartCopy` API reference — https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html

### DIFF
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -3087,10 +3087,34 @@ def _upload_part_copy(bucket_name: str, dest_key: str, query_params: dict, heade
 
     # Handle x-amz-copy-source-range
     copy_range = headers.get("x-amz-copy-source-range", "")
-    if copy_range and copy_range.startswith("bytes="):
-        rng = copy_range[6:]
-        start, end = rng.split("-")
-        src_body = src_body[int(start):int(end) + 1]
+    if copy_range:
+        _malformed = _error(
+            "InvalidArgument",
+            "The x-amz-copy-source-range value must be of the form "
+            "bytes=first-last where first and last are the zero-based offsets "
+            "of the first and last bytes to copy",
+            400,
+        )
+        if not copy_range.startswith("bytes=") or "," in copy_range:
+            return _malformed
+        rng = copy_range[len("bytes="):]
+        parts = rng.split("-")
+        if len(parts) != 2 or parts[0] == "" or parts[1] == "":
+            return _malformed
+        try:
+            start, end = int(parts[0]), int(parts[1])
+        except ValueError:
+            return _malformed
+        if start < 0 or end < start:
+            return _malformed
+        object_size = len(src_body)
+        if start > object_size - 1 or end > object_size - 1:
+            return _error(
+                "InvalidArgument",
+                f"Range specified is not valid for source object of size: {object_size}",
+                400,
+            )
+        src_body = src_body[start:end + 1]
 
     etag = f'"{md5_hash(src_body)}"'
     _multipart_uploads[upload_id]["parts"][part_number] = {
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Amazing work @mfurqaan31 !

PR #624, opened 2026-05-12
fix(s3): handle absolute-form request targets in _parse_bucket_key

mark-bray:
AWS SDK for .NET v4 sends HTTP/1.1 requests with absolute-form targets (e.g. PUT http://ministack:4566/bucket/key). hypercorn passes the raw target as scope["path"], so _parse_bucket_key received the full URL and parsed "http:" as the bucket name, returning NoSuchBucket.

Strip the scheme and authority before parsing when the path starts with http:// or https://.


Nahuel990 (approved):

Nahuel990:
Thanks @mark-bray !

PR #634, opened 2026-05-13
add /v2/credentials endpoint and inject container env for ECS tasks

YakirOren:
Real ECS tasks read their credentials from `/v2/credentials/<uuid>` on
  `169.254.170.2` (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`).
  MiniStack wasn't serving that path, so SDKs inside a `RunTask` container had no
  way to fetch credentials.
  This adds the handler, which returns the same shape IMDS already serves, and
  injects `AWS_CONTAINER_CREDENTIALS_FULL_URI`,
  `AWS_CONTAINER_AUTHORIZATION_TOKEN`, and `AWS_ENDPOINT_URL` into every spawned
  container so an unmodified SDK finds it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ---
 
+## [Unreleased]
+
+### Added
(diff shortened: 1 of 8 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this @YakirOren just a heads up, before releasing I will close some gaps that are as a workaround in your implementation

Issue #639, opened 2026-05-14
Step Functions: executions silently stall at ExecutionStarted under per-account isolation (contextvars not propagated to background threads)


michael-denyer:
### What happened?

Step Functions executions silently stall at `ExecutionStarted` when the caller uses a non-default 12-digit access-key (per-account isolation, as documented in the README's multi-tenancy section). The execution thread starts but never enters its first state — no `TaskStateEntered`, no errors, no log output.

**Root cause:** `_start_execution` writes to `_executions[exec_arn] = {...}` where `_executions` is an `AccountScopedDict`. The write happens in the request-handler context where `get_account_id()` returns the caller's account (e.g. `123456789012` from a 12-digit access key). Then `_start_execution` spawns the execution worker via:

```python
threading.Thread(target=_run_execution, args=(exec_arn,), daemon=True).start()
```

Python's `threading.Thread` does **not** propagate `contextvars.ContextVar` values into the new thread (see https://docs.python.org/3/library/contextvars.html — only `asyncio.Task` and `concurrent.futures` workers explicitly opt in). So `_run_execution` runs with the default `_request_account_id` value (`000000000000`). At its first line:

```python
execution = _executions.get(exec_arn)
if not execution:
    return
```

…it looks up the execution under account `000000000000`, finds nothing, and silently returns. No execution events get added, no logs, no diagnostics.

Same issue affects two other thread-spawn sites in `stepfunctions.py`:
- Line ~1701: `threading.Thread(target=run_branch, ...)` inside `_invoke_parallel` (Parallel state branches)
- Line ~1755: `ThreadPoolExecutor(...).submit(run_item, ...)` inside `_invoke_map` (Map state iterations)

Even if the main execution thread is fixed, nested Parallel branches and Map iterations will lose the account context.

### Reproduction steps

```python
import boto3, json, time

# Use a 12-digit access key to trigger account-scoped storage
kw = dict(
    endpoint_url="http://localhost:4566",
    aws_access_key_id="123456789012",
    aws_secret_access_key="test",
    region_name="us-east-2",
)
sfn = boto3.client("stepfunctions", **kw)
lam = boto3.client("lambda", **kw)

# Trivial Lambda
import io, zipfile
buf = io.BytesIO()
with zipfile.ZipFile(buf, "w", zipfile.ZIP_DEFLATED) as zf:
    zf.writestr("lambda_function.py",
                b"def handler(event, context):\n    return {'ok': True}\n")
lam.create_function(
[...]

PR #640, opened 2026-05-14
fix(stepfunctions): propagate request contextvars into background execution threads (#639)

michael-denyer:
## Summary

Fix Step Functions executions silently stalling at `ExecutionStarted` when the caller uses a non-default 12-digit access-key (the per-account-isolation pattern documented in the README).

`_start_execution`, `_execute_parallel`, and `_execute_map` each spawn worker threads via `threading.Thread` / `ThreadPoolExecutor`. Python doesn't propagate `contextvars` to those threads automatically, so the workers run with the default `_request_account_id` value (`000000000000`) and can't find the execution / state-machine / etc. records that the calling request stored under the caller's tenant in `AccountScopedDict`.

The fix snapshots the calling context with `contextvars.copy_context()` and wraps each spawned target with `ctx_snapshot.run(...)`. The snapshot is taken at the request boundary in `_start_execution` and re-snapshotted at each nested fan-out (Parallel branches, Map iterations) so tenancy context flows through arbitrarily deep ASL trees.

## Why this matters

The bug is invisible to tests that use the default `aws_access_key_id="test"` (both the request thread and the worker thread fall back to the same default account `000000000000`, so the AccountScopedDict lookup happens to succeed). It only manifests when per-account isolation is exercised. No logs, no errors, no diagnostics — executions just sit at `ExecutionStarted` forever, defeating multi-tenant test suites that want to scope resources per fake account.

See [issue #639](https://github.com/ministackorg/ministack/issues/639) for the full reproducer.

## Test plan

- [x] Added regression test `tests/test_stepfunctions.py::test_sfn_execution_proceeds_under_non_default_account_id` — creates an SFN client with `aws_access_key_id="123456789012"`, deploys a one-state Pass machine, asserts the execution reaches `SUCCEEDED` with the full event history (`ExecutionStarted` → `ExecutionSucceeded`). Without the fix this test stalls until pytest-timeout.
- [x] Built the fork locally (`docker build -t ministack-patch-test:639 .`) and ran a trivial single-state ASL with a `lambda:invoke` task against account `123456789012`. Execution went from RUNNING → SUCCEEDED in <1 s with the expected 8-event history (`ExecutionStarted`, `TaskStateEntered`, `TaskScheduled`, `TaskSucceeded`, `TaskStateExited`, `SucceedStateEntered`, `SucceedStateExited`, `ExecutionSucceeded`).
- [ ] CI test matrix (handed to the maintainers; my local sanity covers the new regression test specifically).
- [x] No public API surface change — only internal thread-spawn wrapping.

## Scope

Targeted: only the three thread/threadpool spawn points in `ministack/services/stepfunctions.py`. Other services that use background threads + AccountScopedDict (briefly scanned: I didn't see any in `lambda_svc.py`, `batch.py`, `dynamodb_streams.py`, `ecs.py`; please flag if I missed any) would need the same treatment, but those weren't part of the failure pattern I observed.

### DIFF
--- a/ministack/services/stepfunctions.py
+++ b/ministack/services/stepfunctions.py
@@ -25,6 +25,7 @@ SUCCEEDED / FAILED / TIMED_OUT / ABORTED.
 
 import ast
 import asyncio
+import contextvars
 import copy
 import json
 import logging
@@ -532,8 +533,18 @@ def _start_execution(data):
         ],
     }
 
+    # Propagate the request's contextvars (notably the account ID set by
+    # set_request_account_id) into the background execution thread. Python's
+    # threading.Thread does NOT automatically copy contextvars, so without
+    # this snapshot the worker runs under the default account and silently
+    # fails to find the execution stored in AccountScopedDict under the
+    # caller's account. See issue #639.
+    ctx_snapshot = contextvars.copy_context()
     threading.Thread(
-        target=_run_execution, args=(exec_arn,), daemon=True).start()
+        target=ctx_snapshot.run,
+        args=(_run_execution, exec_arn),
+        daemon=True,
+    ).start()
 
     logger.info("Step Functions execution started: %s", exec_arn)
     return json_response({"executionArn": exec_arn, "startDate": start_date})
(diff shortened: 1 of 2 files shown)


michael-denyer:
Could a maintainer please add the `bug` label? External contributors can't apply labels via GraphQL (`AddLabelsToLabelable` requires repo write access). This PR pairs with the labelled issue [#639](https://github.com/ministackorg/ministack/issues/639). Thanks.

Nahuel990:
Hey @michael-denyer awesome! Thanks for this... one minor thing can you move copy_context() inside the comprehensions so each thread gets its own snapshot? (a single Context can't be entered by two threads at once)

michael-denyer:
Thanks for the review! Pushed an update that:

1. Moves `contextvars.copy_context()` **inside** the comprehensions for the Parallel-branches and Map-iterations sites — each spawned thread now gets its own fresh `Context` snapshot, so two threads can't be running the same `Context.run` concurrently. The main-execution site (single thread) keeps the captured-once form.
2. Rebased onto the latest `main`.

Diff for the parallelism sites:

```python
# Parallel branches (_execute_parallel)
threads = [
    threading.Thread(
        target=contextvars.copy_context().run,
        args=(run_branch, i, b),
        daemon=True,
    )
    for i, b in enumerate(branches)
]

# Map iterations (_execute_map)
with ThreadPoolExecutor(max_workers=workers) as pool:
    futs = [
        pool.submit(contextvars.copy_context().run, run_item, i, item)
        for i, item in enumerate(items)
    ]
```

Should be ready for another look.

Nahuel990 (approved):

PR #646, opened 2026-05-15
feat(cognito): route invitation and verification emails through SES

kjdev:
## Summary
- `AdminCreateUser`, `SignUp`, `ResendConfirmationCode`, `ForgotPassword`, and `AdminResetUserPassword` now hand their welcome / temporary-password / verification mail to the in-process SES emulator, so simulated apps see the message in `/_ministack/ses/messages` and have it relayed via SMTP when `SMTP_HOST` is set.
- Mirrors AWS behaviour: `MessageAction=SUPPRESS` skips delivery, `RESEND` re-sends, `DesiredDeliveryMediums=["SMS"]` excludes email, `AdminCreateUserConfig.InviteMessageTemplate` / `VerificationMessageTemplate` placeholders (`{username}`, `{####}`) expand, and the sender resolves to `EmailConfiguration.From` falling back to `no-reply@verificationemail.com` (overridable via `COGNITO_DEFAULT_FROM`). `COGNITO_EMAIL_ENABLED=false` short-circuits delivery globally.
- Adds the previously-missing `ResendConfirmationCode` action while wiring the delivery path.
- `ses.py` gains a `send_internal_email()` entry point so the record/relay logic stays in one place for any future emulator that needs to push mail through SES.

## Test plan
- `pytest tests/test_cognito.py` — new cases cover invitation delivery, `SUPPRESS` short-circuit, `RESEND` re-send, SMS-only medium skipping email, `EmailConfiguration.From` overriding `COGNITO_DEFAULT`, and `SignUp` / `ForgotPassword` / `ResendConfirmationCode` pushing verification codes.
- Process-level check that `COGNITO_EMAIL_ENABLED=false` stops delivery before SES.
- Manual: hit `AdminCreateUser` against a running gateway and confirm the message appears at `GET /_ministack/ses/messages`.

### DIFF
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -13,7 +13,7 @@ User Pools operations:
   AdminSetUserPassword, AdminUpdateUserAttributes,
   AdminInitiateAuth, AdminRespondToAuthChallenge,
   InitiateAuth, RespondToAuthChallenge, SignUp, ConfirmSignUp,
-  ForgotPassword, ConfirmForgotPassword, ChangePassword,
+  ResendConfirmationCode, ForgotPassword, ConfirmForgotPassword, ChangePassword,
   GetUser, UpdateUserAttributes, DeleteUser,
   AdminAddUserToGroup, AdminRemoveUserFromGroup,
   AdminListGroupsForUser, AdminListUserAuthEvents,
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks @kjdev

Issue #649, opened 2026-05-15
S3 → Lambda direct event notification: API accepts configuration but uploads do not invoke Lambda


michael-denyer:
## Affected Service

S3 (cross-service interaction with Lambda)

## What happened?

`put_bucket_notification_configuration` with a `LambdaFunctionConfigurations` block returns success against MiniStack v1.3.39, but subsequent `s3:PutObject` events do not invoke the referenced Lambda. No error, warning, or partial-progress signal — uploads succeed and the Lambda is never called.

Silent acceptance is more problematic than an explicit "not implemented" error because test wiring appears to succeed at configuration time and only fails downstream when assertions on Lambda side-effects don't materialise.

The README documents SNS → Lambda fanout but does not list S3 → Lambda direct. The current API behaviour (silent accept + no invocation) is the worst of both worlds — neither documented as unsupported nor functional.

## Reproduction steps

```python
import boto3

endpoint = "http://localhost:4566"
s3 = boto3.client("s3", endpoint_url=endpoint)

s3.create_bucket(Bucket="test-bucket")
# Lambda creation omitted for brevity — assume function ARN exists.

s3.put_bucket_notification_configuration(
    Bucket="test-bucket",
    NotificationConfiguration={
        "LambdaFunctionConfigurations": [{
            "LambdaFunctionArn": "arn:aws:lambda:us-east-1:000000000000:function:test-handler",
            "Events": ["s3:ObjectCreated:*"],
        }]
    },
)  # returns 200 — config "accepted"

s3.put_object(Bucket="test-bucket", Key="example.txt", Body=b"hello")  # also 200

# Lambda is never invoked. No CloudWatch logs, no event delivery.
```

Production AWS invokes the Lambda within ~1s of `put_object`.

## Expected behavior

Either:

1. Implement S3 → Lambda direct invocation so configured notifications actually fire, OR
2. Return an error on `put_bucket_notification_configuration` when `LambdaFunctionConfigurations` is included and the feature is unimplemented — so test wiring fails loudly at configuration time rather than silently at assertion time.

Loud configuration-time failure beats silent runtime no-op.

## MiniStack version

v1.3.39

## SDK / tool used

boto3

michael-denyer:
Closing this — I filed it without re-verifying current behavior, and the claim is wrong.

I wrote a reproducer mirroring the exact configuration in the issue body and ran it against `ministackorg/ministack:1.3.39@sha256:057472d4...`:

[...]

PR #651, opened 2026-05-15
test(s3): add S3 → Lambda direct notification coverage

michael-denyer:
## What this PR does

Adds two regression tests for the S3 → Lambda direct notification path that already exists in `ministack/services/s3.py::_deliver_event_to_lambda` but had no end-to-end test coverage.

Two tests, mirroring the existing `test_s3_event_notification_to_sqs` + `test_s3_event_notification_filter` pair:

- **`test_s3_event_notification_to_lambda`** — basic happy path. Creates a Lambda whose handler writes a marker key back to the source bucket, configures `put_bucket_notification_configuration` with a `LambdaFunctionConfigurations` block targeting it, fires a `PutObject`, and asserts the marker key appears within 5s of async invocation.
- **`test_s3_event_notification_to_lambda_with_filter`** — same shape, but asserts the `S3Key` suffix filter applies to Lambda destinations just as it does to SQS destinations (only the matching object key triggers the Lambda; the non-matching one is filtered out).

Both pass locally against v1.3.39 in ~1.8s combined.

## Why

I filed [#649](https://github.com/ministackorg/ministack/issues/649) claiming this code path silently no-ops at v1.3.39 — that was wrong, and I'm closing it with an apology comment. The issue was filed against stale notes from a downstream consumer's runbook (which had been written against an earlier ministack version) without re-verifying current behavior. The reproducer that proved the issue wrong is the basis for this regression test, so the same gap can't happen again.

There's no functional code change in this PR — only test additions. The Lambda dispatch path itself works correctly today.

## Test plan

- [x] `pytest tests/test_s3.py::test_s3_event_notification_to_lambda` passes locally against `ministackorg/ministack:1.3.39`
- [x] `pytest tests/test_s3.py::test_s3_event_notification_to_lambda_with_filter` passes locally against `ministackorg/ministack:1.3.39`
- [x] No other tests in `test_s3.py` affected — adds two new test functions immediately after the existing SQS notification test, no other changes

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -1,7 +1,9 @@
+import io
 import json
 import os
 import re
 import time
+import zipfile
 from datetime import datetime
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


jgrumboe:
@michael-denyer 
Your tests are failing because the test without filter is creating an infinite loop because the marker file is written back to the same bucket and therefore it triggers the notification again and again...
And you don't need to set AWS_ENDPOINT_URL as Lambda env because it's injected correctly by the Lambda service since a few versions.

BUT more important:
There's already the test [`test_s3_event_notification_to_lambda_boto3_default`](https://github.com/ministackorg/ministack/blob/e88722b041dde07b1fed01eaf98c369d23fbdb38/tests/test_s3.py#L1065) that covers already the same use case. And in my opinion in a better way by checking against the invocation logs of the Lambda itself.
Your tests do not add new test coverage, except the filter test.
If this is your intent to add filter test coverage, please adapt your PR to mimic the `test_s3_event_notification_to_lambda_boto3_default` test.

Nahuel990:
@michael-denyer Thanks for this test, we already have similar coverage and those tests are failing so I prefer to close this PR

PR #666, opened 2026-05-18
Fix RDS Data API readiness race

jayjanssen:
🤖 Generated by my AI agent.

## Why
MiniStack could report Docker-backed RDS/Aurora resources as usable before the backing database accepted authenticated connections, letting RDS Data API acknowledge early writes through the in-memory stub. Aurora MySQL cluster endpoints also needed to point at the local backing instance so callers using `DescribeDBClusters.Endpoint` can connect.

## What
- Gate Docker-backed RDS availability on authenticated database readiness
- Keep parent Aurora clusters in creating while member instances are not ready
- Return DatabaseUnavailableException for real endpoint connection failures instead of using the SQL stub
- Sync Aurora cluster endpoints to backing instance endpoints and grant MySQL master-user privileges
- Add regression coverage and changelog entries

## Risk Assessment
Medium: this changes RDS readiness timing and Aurora endpoint behavior for Docker-backed databases, while preserving stub behavior for metadata-only environments.

## References
- Tests: `uv run --with pytest --with pytest-xdist --with boto3 --with cbor2 --with docker pytest tests/test_rds_data.py tests/test_rds.py -q`
- Tests: `uv run --with pytest --with pytest-xdist --with boto3 --with cbor2 --with docker pytest tests/test_rds_data.py -q`
- Lint: `uv run --with ruff ruff check ministack/services/rds.py ministack/services/rds_data.py tests/test_rds_data.py tests/test_rds.py`

Generated with Codex


Nahuel990:
@jayjanssen I will merge this becuase of the fix you implemented but I will restore the non-blocking part of db which is the correct AWS behaviour (return "Creating" after a few seconds)

Nahuel990 (approved):

PR #680, opened 2026-05-21
feat(s3): support If-None-Match / If-Match preconditions on PutObject

mattcookio:
## Summary

Implements the conditional-write headers AWS S3 added in November 2024, following the behaviors documented in:

- [PutObject API reference — request headers](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html)
- [Conditional writes — user guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html)

| Header | AWS spec | This PR |
|---|---|---|
| `If-None-Match: "*"` | *"expects the asterisk value"* — fails 412 when key exists | ✅ matches |
| `If-Match: "<etag>"` | *"expects the ETag value as a string"* — fails 412 when ETag differs | ✅ matches |
| `If-Match: "<etag>"` against missing key | **404 NoSuchKey** ([user guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html#conditional-error-response): *"the operation fails with a 404 Not Found error"*) | ✅ matches |
| `If-None-Match: "<etag>"` | not explicitly documented for PutObject; RFC 7232 default | ✅ 412 (RFC behavior) |
| `If-Match: "*"` | not explicitly documented for PutObject; RFC 7232 default | ✅ 412 (RFC behavior) |

## Implementation

Small helper `_check_put_preconditions` next to the existing `_validate_content_md5`. `_put_object` calls it after MD5 and storage-class validation, before mutating bucket state. ETag comparison strips surrounding quotes on both sides — clients are inconsistent about including them.

The existing `_copy_object` already supports `x-amz-copy-source-if-match` / `x-amz-copy-source-if-none-match`; this fills the symmetric gap on plain `PutObject`.

## Motivation

We use a single shared S3 slot serializing concurrent arrivals through `If-None-Match: "*"` to pair two-file inbound batches in our transformation pipeline. Real S3 enforces it; MiniStack Community currently silently accepts the second PUT and overwrites — so integration tests against MiniStack would produce false positives for code we know is correct against AWS. This closes that parity gap.

## Tests

Six new tests in `tests/test_s3.py`:

- `test_s3_put_object_if_none_match_star_no_existing` — succeeds when key empty
- `test_s3_put_object_if_none_match_star_existing_fails` — 412 when key exists; original bytes preserved
- `test_s3_put_object_if_none_match_etag` — succeeds on mismatched ETag, fails on matching
- `test_s3_put_object_if_match_star_requires_existing` — 412 when missing; succeeds when present
- `test_s3_put_object_if_match_etag` — succeeds on matching ETag, 412 on stale
- `test_s3_put_object_if_match_etag_missing_object_returns_404` — covers the AWS-specific 404 case

All six pass locally. Pre-existing failures in `tests/test_s3.py` (multipart-copy range parsing, chunked-upload tests) are unrelated and exist on `main` without these changes.

botocore strips `IfNoneMatch` / `IfMatch` from `put_object` calls on older SDK versions (params added in late 2024), so the tests inject the headers via `before-send.s3.PutObject` event handlers to ensure they reach the wire. This makes the tests robust across botocore versions.

## Known limitation

AWS S3 returns `409 Conflict` (not `412`) for concurrent operations — e.g. when a delete or another conditional write completes between the read-back and the write. MiniStack's in-memory single-threaded model has no distributed-system race window to detect, so this PR always returns 412 for any precondition violation. Users testing concurrent-conflict code paths against MiniStack should be aware of this — it's a backend-emulation limit, not a correctness issue.

## Out of scope

- `DeleteObject` conditional headers — same precondition shape; separate PR if there's interest.
- `CompleteMultipartUpload` conditional writes — AWS supports the same headers there per the [user guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html); could follow.

### DIFF
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -405,6 +405,83 @@ def _validate_content_md5(headers: dict, body: bytes):
     return None
 
 
+def _check_put_preconditions(headers: dict, existing_obj: dict | None):
+    """Evaluate `If-Match` and `If-None-Match` on PUT.
+
+    AWS S3 added native conditional writes on PutObject in November 2024:
+      - `If-None-Match: "*"` — succeed only when no object exists at the key.
+        Used to implement create-once semantics (idempotent writes, distributed
+        leader election via S3, two-file pair serialization).
+      - `If-None-Match: "<etag>"` — succeed only when the existing object's
+        ETag does NOT match.
+      - `If-Match: "*"` — succeed only when an object already exists.
+      - `If-Match: "<etag>"` — succeed only when the existing object's ETag
+        matches.
+
+    Returns a 412 PreconditionFailed tuple when a condition is violated;
+    otherwise returns None and the caller proceeds with the PUT.
+
+    ETag comparison strips surrounding quotes on both sides — S3 stores ETags
+    with quotes but client code is inconsistent about including them.
+    """
+    if_none_match = headers.get("if-none-match", "").strip()
+    if_match = headers.get("if-match", "").strip()
+
+    if not if_none_match and not if_match:
+        return None
+
+    existing_etag = (
+        existing_obj["etag"].strip('"') if existing_obj is not None else None
+    )
+
+    if if_none_match:
+        # "*" form: any existing object violates the condition.
+        if if_none_match == "*":
+            if existing_obj is not None:
+                return _error(
+                    "PreconditionFailed",
+                    "At least one of the pre-conditions you specified did not hold",
+                    412,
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks for this @mattcookio! This will be released in a few hours in the next version, welcome to Ministack!

Nahuel990 (approved):

PR #686, opened 2026-05-22
security(pull_request_target) fix

BenKamin03:
## Summary

Switch Docker preview publishing from `pull_request_target` to `pull_request`.

This avoids the risky pattern highlighted in the [TanStack supply-chain compromise postmortem](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem): running untrusted PR code in a privileged workflow with access to secrets, and poison caches.

## Changes

- Replaced `pull_request_target` with `pull_request`.
- Added a same-repository guard so Docker preview images are only published for trusted internal PR branches.
- Forked PRs no longer attempt to publish images or access Docker Hub secrets.

## Testing

Not run locally, GitHub Actions workflow-only change.

---

if you don't think this is needed, feel free to close this pr - i found this while doing my other PR, and thought it might be helpful.

### DIFF
--- a/.github/workflows/docker-publish-on-pr.yml
+++ b/.github/workflows/docker-publish-on-pr.yml
@@ -1,7 +1,7 @@
 name: Docker Publish on PR
 
 on:
-  pull_request_target:
+  pull_request:
     types: [opened, synchronize, reopened, labeled]
 
 env:
@@ -9,6 +9,10 @@ env:
 
 jobs:
   publish-pr:
+    # Only publish preview images for trusted PRs whose source branch lives in
+    # this repository. Forked PRs run without repository secrets on
+    # `pull_request`, and this guard prevents accidental publish attempts, and cache poisoning.
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
 
     permissions:


jgrumboe:
Thanks for opening this PR.
I will review it.

jgrumboe:
Thanks for the security improvement here. I think the right behavior for this workflow is slightly different, though.
We introduced preview image builds so people can test PR changes with a Docker image while the PR is still under review.
Since our regular CI/tests already run in a separate workflow, I don’t think this workflow needs to run on general `pull_request` events at all to strengthen the security posture.

Instead, I’d suggest making preview publishing explicitly maintainer-gated:
- trigger only when a PR gets a specific label, e.g. `safe-to-preview`
- publish only for that labeled PR
- build the `full` image together, no need for a second label anymore

That would make the workflow’s purpose much clearer: it’s not general CI and it’s not automatic preview publishing — it’s a maintainer-approved preview publish flow.
This should also let us support fork PRs when we intentionally want to, without automatically publishing preview images for every PR, thereby addressing the security risk in a similar way.

@Nahuel990 What do you think?

Nahuel990:
@BenKamin03 sorry for the late reply, and thanks for flagging this. I've been exploring the current setup for a few hours and have a couple of points...

First, we don't accept changes from contributors to .ci / workflow files for exactly this kind of security reason... If a contributor wants to raise something there, the flow is to open an issue, @jgrumboe or I review it, and we apply the change ourselves. I'll add that to CONTRIBUTING.md to avoid confusion in the future

About the token, it only runs in the shell during login and push, never as a build-arg or secret mount, so there's no path for a contributor's Dockerfile to export it. The worst case is someone publishing a malicious image under our preview channel, which we don't encourage anyone outside the PR author to pull anyway (the correct image is ministackorg/ministack:latest or :full) 

About the label, in the end it's the same thing. Any maintainer would just apply the label without re-reading the Dockerfile, same as clicking "approve workflow run" I don't think it doesn't actually add a review step there

What I think "*does*" help is a path-based guard on the workflow itself. Any PR that touches Dockerfile, .dockerignore, or anything under `.github/` skips the auto-preview, and only runs after a maintainer reviews and builds manually. That keeps the auto-publish for the 99% case (pure Python, docs, tests) and gates the build surface by humans by default...

PR #693, opened 2026-05-23
Fix CloudFront invalidation caller reference idempotency

CoffeeRaptor:
## Summary
- Return the existing CloudFront invalidation when `CreateInvalidation` is retried with the same `CallerReference` and identical paths.
- Return `InvalidationBatchAlreadyExists` when the same `CallerReference` is reused with different paths.
- Add regression coverage for both idempotent retry and conflicting batch behavior.

## Tests
- `python3 -m py_compile ministack/services/cloudfront.py tests/test_cloudfront.py`
- `ruff check ministack/services/cloudfront.py tests/test_cloudfront.py`
- `pytest tests/test_cloudfront.py -v -k invalidation`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 ## [1.3.47] — 2026-05-22
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks @CoffeeRaptor !

PR #698, opened 2026-05-23
Support AWS Serverless basic resources

### LINKED ISSUE #690: Support AWS::Serverless::LayerVersion resource type in cloud formation.
### AWS Service name

CloudFormation

### Operations needed

Support of AWS::Serverless::LayerVersion  as part of CloudFormation.

### Use case

Deploying local sam application to Ministack.

Note: I may try to write code for this, I just need a guidance on a start-point and will dig into MiniStack next week.

- [x] Yes, I'd like to implement this service and open a PR
- [x] I can help review/test an implementation

ShadyAbuKalam:
### Overview
Fixes #690 

This is WIP to support the following 3 components from SAM Applications into Cloud Formation.
- AWS::Serverless::Function
- AWS::Serverless::StateMachine
- AWS::Serverless::LayerVersion

They are translated by cloud formation to their respective CN components, but with a few tweaks / conversions.

This PR acts as mapping between the serverless resources and CN native resources
---
1. Support AWS::Serverless::Function ( is implemented as AWS::Lambda::Function) with the following notes:
   1. Map CodeUri -> Code ✅
      - CodeUri supports localpath (sam build/deploy will upload this to S3 and sends S3 path), S3 paths and FunctionCode
   2. Map Tracing -> TracingConfig ✅
      - Add `"arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess"` to the role's managed policies if no role was there.
   3. Map Policies -> Role's ManagedPolciiesArn/Policies (In the change set)
      - SAM Policy template ✅
      - inline Policies (Maybe not in this phase)
      - Full ARN policy ✅
      - AWS Managed Policy alias ✅
   4. If missing role, create a new role with the managed policies from Policies at the change set creation. ✅
   5. Write UTs.

2. Support AWS::Serverless::StateMachine with following notes: (Currently investigating the mapping)
   1. Map DefinitionUri -> DefinitionS3Location ✅
   2. Map Logging -> LoggingConfiguration ✅
   3. Map Type -> StateMachineType ✅
   4. Map Role -> RoleArn ✅
   5. Map Name -> StateMachineName ✅
   6. If role is missing, Add one, with mapping similar to Role in AWS::Serverless::Function ✅
   7. Write UTs.
3. Support AWS::Serverless::LayerVersion ( is implemented as AWS::Lambda::LayerVersion) with following notes:
   1. Map ContentUri -> Content ✅
      - ContentUri supports localpath (sam build/deploy will upload this to S3 and sends S3 path), S3 path and LayerContent.
   2. Write UTs.
4. Support for Globals section. ✅
---

Note: I have only one day per week to work on this, so I will update this next Saturday.


ShadyAbuKalam:
@Nahuel990 , I have a few questions marked as `#TODO` in the code.
If you can have a look on the basis of what I am doing, it will be great.

ShadyAbuKalam:
@Nahuel990 , Would you check this as this is implementation ?
I want to have a feedback loop before going in with UTs.

Nahuel990:
@ShadyAbuKalam Hi, sorry I'm not understanding what you're trying to implement here...

There's a clear drift from your original issue #690 (which asked for AWS::Serverless::LayerVersion - plus some resources) and what's in the PR... it now spans Function, LayerVersion, and StateMachine, but for each one only the property-rename layer is in place (e.g. CodeUri→Code, ContentUri→Content, DefinitionUri→DefinitionS3Location)...

The SAM-specific behaviors that make these resources actually useful aren't here: Events (the whole reason most people use SAM Function), AutoPublishAlias, auto-created LogGroup, RetentionPolicy / PublishLambdaVersion semantics on LayerVersion, inline Definition on StateMachine, ImageUri / InlineCode on Function... A user shipping a real sam init template would deploy a Lambda with no triggers and a StateMachine they couldn't define inline...

Could you narrow this back to a clear scope before we go further? Either (a) reshape this PR and in one focused strictly on LayerVersion per the original issue and we go from there, if you try to implement full SAM it will be a massive PR and I'm already seeing some basic Python errors like the logger %n (this is old, which Python version are you using) instead of %d

ShadyAbuKalam:
@Nahuel990 

Well, I have a SAM application that uses these components, that I am trying to emulate.

At first I thought only LayerVersion is missing.
Then I dug deep in ministack code and found it has no SAM support at all. So I decided to add emulation support for whatever we use in our app.

I have a shallow knowledge of AWS, so I had to do some research.

I dug deeper in the AWS documentation and played around sam deployments and found that everything is translated into cloud formation native resources.

So I wrote the mapping layer for them, add support for automatic role creation and policy management.

I am not sure what's your plan for SAM apps, but we can discuss how to implement it over a series of phases.

----
For Python part, I use python 3.14.
I simply didn't write Python code in the last 6 years, and before that I was writing c-extensions for it.

Yeah, my Python skills are a bit rusty.

Nahuel990:
We don't support SAM, people using SAM have already sam local (which is officially maintained by AWS) people would use Ministack when they try to use any of the other IaC (TF, OT, CfN, etc) I thought your original need was to have the Serverless resource for CfN, that's why I don't understand this scope

ShadyAbuKalam:
Yes, but the sam local is only for a single invocation and don't support full state machine testing.

Nahuel990:
@ShadyAbuKalam but if that's the case you have to drop sam, and deploy your lambdas / code into Ministack, instead of using `sam deploy ` you'll need to run `aws cloudformation deploy --endpoint http://localhost:4566` (Ministack endpoint)

ShadyAbuKalam:
I don't use `sam deploy`, I use the wrapper that local stack has made https://github.com/localstack/aws-sam-cli-local which is a thin wrapper around the original sam command.

I also tried the `aws cloudformation deploy`

AWS cloudformation command doesn't wrap code and uploads it to S3, on the other hand sam does this.

The second thing, aws cloudformation deploy still sends the `AWS::Serverless` resources to ministack, so yes we still need to support them in the cloudformation module in ministack to accept them.

Nahuel990:
@ShadyAbuKalam I owe you a correction... you're right that `aws cloudformation deploy` sends `AWS::Serverless` resources and the transform is server-side. I was wrong about that.

I thought you were running the `sam build` / `sam package` path and deploying manually in Ministack (in this case, the translation for Serverless would have been a few lines of code that map Serverless -> to each behaviour)

However, we won't be adding SAM transform support to MiniStack. AWS provides the official tooling for this:

**Lambda testing:** `sam local invoke` / `sam local start-lambda` / `sam local start-api`

**Step Functions testing:** [Step Functions Local](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html) -- `docker run -p 8083:8083 amazon/aws-stepfunctions-local`. Supports endpoint overrides for Lambda, DynamoDB, SQS, SNS, ECS, Glue. AWS has a [tutorial](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local-lambda.html) for running it with `sam local start-lambda

**Unit testing state logic:** [TestState API](https://docs.aws.amazon.com/step-functions/latest/dg/test-state-isolation.html) -- test individual states with mocked service integrations, Map, Parallel, retry/error handling, input/output processing. No deployment needed

If your Lambdas need DynamoDB, SQS, or other services during testing, point Step Functions Local's endpoint overrides at MiniStack:

1. `sam local start-lambda --port 3001`
2. `docker run -p 8083:8083 amazon/aws-stepfunctions-local` with `LAMBDA_ENDPOINT=http://host.docker.internal:3001`, `DYNAMODB_ENDPOINT=http://host.docker.internal:4566`
3. `aws stepfunctions --endpoint http://localhost:8083 create-state-machine ...`

Full state machine testing, real Lambda invocations, MiniStack-backed services. No SAM transform needed.

The `samlocal` wrapper you're using is a LocalStack tool that works around this by adding endpoint support to SAM CLI. That's not something we want to replicate as it's a drift from original AWS behaviour -- AWS already provides the right (and official) tools, they just need to be composed together

ShadyAbuKalam:
For my knowledge aws state machine local tool is not supported any more. I have to check if it's still a viable option or not.

I will investigate the other methods you mentioned and get back to you in a week.

ShadyAbuKalam:
To avoid having this as stale PR.
Nothing of the methods mentioned is enough to test except Step Functions local which is deprecated.

I also wanted to have all my emulation / testing confined in one place.

I will resort to testing locally via lambda testing.

Feel free to close the PR.
---
@Nahuel990  Thanks for your efforts on this project.

PR #702, opened 2026-05-24
feat(inspector2): add Inspector2 emulator

ry-allan:
Add Amazon Inspector2 emulator with deterministic stub vulnerability findings for ECR container images, Lambda functions, and EC2 instances.

New APIs (14 operations):
- Enable, Disable
- ListFindings, BatchGetFindingDetails
- ListCoverage, ListCoverageStatistics
- ListFindingAggregations, SearchVulnerabilities
- TagResource, UntagResource, ListTagsForResource
- CreateFilter, ListFilters, DeleteFilter

Features per-resource-type enable/disable, filtering, sorting, pagination, and per-account state isolation with full persistence support.

Did consider wiring in grype for actual scanning with an engine mode (mock vs grype) but felt that was overkill


Nahuel990:
Hi @ry-allan, thanks for the contribution...

Before I look at the code, I need some insight into this:

1. What's the use case? What tool, application, or workflow of yours is currently blocked because MiniStack doesn't emulate Inspector2? Be specific... the name of the tool, what calls it makes, and what it does with the response...

2. Why these 26 operations? A first PR for a new service should ship the minimum needed to unblock your use case, not a full API surface...

I'm asking mostly because Inspector is not typically a service someone would emulate... it returns metadata from running services, and here you're only adding the control plane, in the past we've had some contributors who asked an AI to "add a new service", and we're trying to avoid that type of behavior so we don't overload MiniStack with services where emulation is not actually needed...

ry-allan:
Hey @Nahuel990,

First off, thanks for providing ministack, I’ve been using it most days for the past month and it’s been fantastic 🙌🏻.

Right now I’m looking into building an internal developer platform - essentially a PaaS abstraction layer over AWS that lets non-technical users deploy static sites and containerised services without needing to understand the underlying infrastructure, whilst giving them visibility into what’s running.

On the security side, I initially integrated Trivy with a small CLI wrapper to run frequent scans, but I’m moving toward Inspector as it feels more self-contained for surfacing vulnerability statistics without the overhead of maintaining my own tooling around it / scheduling etc.

For the API - fair enough on keeping the surface concise, I initially wanted to cover as much as possible but that said, even a handful of the core endpoints would help (Enable, Disable, ListFindings, BatchGetFindingDetails, ListCoverage, ListCoverageStatistics, ListFindingAggregations, SearchVulnerabilities, TagResource, UntagResource, ListTagsForResource, CreateFilter, ListFilters, DeleteFilter).

I've just pushed an update to remove those additional operations, kept to the ones above. 

But either way, I'm happy to fork and just maintain my own copy synced with upstream for my local development so no issues.

Nahuel990:
@ry-allan Perfect, I will merge it now and release it in the next version in a few hours. Thanks!

Nahuel990 (approved):

PR #707, opened 2026-05-25
fix(glue): persist ViewOriginalText/ViewExpandedText in _create_table

### LINKED ISSUE #706: Glue stub: _create_table silently drops ViewOriginalText / ViewExpandedText, breaks Trino CREATE VIEW
**Summary**

MiniStack's Glue stub accepts `CreateTable` requests carrying `ViewOriginalText` and `ViewExpandedText` (the fields that hold the SQL body of a Hive/Presto-style view), but `_create_table` does not persist them. Subsequent `GetTable` calls return the row without those fields. Any engine that reads the view back — including Trino's iceberg connector — fails because the view body it just wrote is gone.

The matching `_update_table` handler (right below `_create_table` in the same file) DOES persist these fields, which strongly suggests the omission in `_create_table` is unintentional.

**Affected versions:** Reproduced on `ministackorg/ministack:1.3.48` (current `latest` as of 2026-05-25). Every release from 1.3.43 through 1.3.48 has the same bug.

**Reproduction (with Trino as the client)**

```bash
docker run -d --name ministack -p 4566:4566 ministackorg/ministack:1.3.48
docker run -d --name minio -p 9000:9000 -p 9001:9001 \
  -e MINIO_ROOT_USER=minioadmin -e MINIO_ROOT_PASSWORD=minioadmin \
  minio/minio server /data --console-address ":9001"

# Configure Trino to use MiniStack as Glue catalog and MinIO as S3.
# (catalog properties file omitted for brevity — standard iceberg connector
#  + glue catalog + s3 endpoint http://localhost:9000)

# 1. Create a schema + a table backing one view.
trino> CREATE SCHEMA iceberg.repro WITH (location='s3://repro-bucket/');
trino> CREATE TABLE iceberg.repro.dim_x (id varchar, name varchar)
       WITH (format='PARQUET', location='s3://repro-bucket/dim_x/');

# 2. Create the view.
trino> CREATE OR REPLACE VIEW iceberg.repro.vw_x AS SELECT id, name FROM iceberg.repro.dim_x;
# CREATE VIEW succeeds.

# 3. Query the view.
trino> SELECT * FROM iceberg.repro.vw_x LIMIT 0;
# Query fails with: viewOriginalText must be present
```

**Reproduction (raw `aws glue create-table` showing the drop)**

```bash
aws --endpoint-url http://localhost:4566 glue create-database --database-input Name=repro

aws --endpoint-url http://localhost:4566 glue create-table --database-name repro --table-input '{
  "Name": "vw_x",
  "TableType": "VIRTUAL_VIEW",
  "ViewOriginalText": "/* Presto View: eyJjYXRhbG9nIjoiaWNlYmVyZyJ9 */",
  "ViewExpandedText": "/* Presto View */"
}'

aws --endpoint-url http://localhost:4566 glue get-table --database-name repro --name vw_x
# Returned Table object does NOT contain ViewOriginalText or ViewExpandedText.
```

**Root cause** (from inspecting the image)

`/opt/ministack/ministack/services/glue.py` in 1.3.48:

```python
def _create_table(data):                # line 249
    db_name = data.get("DatabaseName")
    ...
    table_input = data.get("TableInput", {})
    ...
    _tables[key] = {
        "Name": name,
        "DatabaseName": db_name,
        "Description": table_input.get("Description", ""),
        "Owner": table_input.get("Owner", ""),
        "CreateTime": int(time.time()),
        "UpdateTime": int(time.time()),
        "LastAccessTime": int(time.time()),
        "StorageDescriptor": table_input.get("StorageDescriptor", {}),
        "PartitionKeys": table_input.get("PartitionKeys", []),
        "TableType": table_input.get("TableType", "EXTERNAL_TABLE"),
        "Parameters": table_input.get("Parameters", {}),
        "IsRegisteredWithLakeFormation": False,
        "CatalogId": get_account_id(),
    }
    # ViewOriginalText and ViewExpandedText are NEVER read from table_input.
```

Compare to `_update_table` at line ~307, which explicitly lists them as `safe_keys`:

```python
def _update_table(data):
    ...
    safe_keys = {"Description", "Owner", "StorageDescriptor", "PartitionKeys",
                 "TableType", "Parameters", "ViewOriginalText", "ViewExpandedText"}
    for k in safe_keys:
        if k in table_input:
            _tables[key][k] = table_input[k]
```

**Suggested fix**

Add `ViewOriginalText` and `ViewExpandedText` to the `_create_table` dict construction (mirroring `_update_table`'s `safe_keys`). Minimal diff:

```python
@@ _create_table
     "TableType": table_input.get("TableType", "EXTERNAL_TABLE"),
     "Parameters": table_input.get("Parameters", {}),
+    "ViewOriginalText": table_input.get("ViewOriginalText"),
+    "ViewExpandedText": table_input.get("ViewExpandedText"),
     "IsRegisteredWithLakeFormation": False,
```

(Using `.get(...)` so callers omitting the fields get `None`, which matches real Glue's behavior for non-view tables.)

**Impact**

Anything using Trino, Spark, or Athena against MiniStack's Glue catalog that creates a view via `CreateTable` (the standard path — every engine does this). The bug isn't visible at `CREATE VIEW` time because the response is `200 OK`; it surfaces only when a subsequent reader tries to expand the view, with an error message that doesn't obviously point at MiniStack.

yonatoasis:
## Summary

`_create_table` in `ministack/services/glue.py` builds the stored table dict without reading `ViewOriginalText` / `ViewExpandedText` from `TableInput`, so the SQL body of Hive/Presto-style views is dropped on creation. The matching `_update_table` handler already lists both fields in its `safe_keys`, so this just mirrors that there.

Two-line change in `_create_table` plus a regression test.

Fixes #706.

## Test plan

All steps verified locally. Baseline is `ministackorg/ministack:1.3.48`; fixed image is `ministack:fix-views` built from this branch. The end-to-end stack used Trino 476 + MinIO `RELEASE.2025-09-07T16-13-09Z` on isolated ports (MiniStack 4569, MinIO 9200/9201, Trino 8091).

- [x] **`aws glue create-table` with `ViewOriginalText`/`ViewExpandedText` followed by `get-table` returns both fields.**

  Baseline (`1.3.48`):
  ```
  ["vw_x", "VIRTUAL_VIEW", null, null]
  ```
  Fixed image:
  ```
  ["vw_x", "VIRTUAL_VIEW", "/* Presto View: eyJjYXRhbG9nIjoiaWNlYmVyZyJ9 */", "/* Presto View */"]
  ```

- [x] **Trino `CREATE OR REPLACE VIEW` + `SELECT * FROM <view>` against MiniStack-backed Glue catalog no longer fails with `viewOriginalText must be present`.**

  Against an isolated Trino + MinIO + fixed-image MiniStack stack:
  ```sql
  CREATE SCHEMA iceberg.repro WITH (location='s3://repro-bucket/');           -- CREATE SCHEMA
  CREATE TABLE iceberg.repro.dim_x (id varchar, name varchar)
    WITH (format='PARQUET', location='s3://repro-bucket/dim_x/');             -- CREATE TABLE
  INSERT INTO iceberg.repro.dim_x VALUES ('a','Alice'),('b','Bob');           -- INSERT: 2 rows
  CREATE OR REPLACE VIEW iceberg.repro.vw_x AS SELECT id, name FROM iceberg.repro.dim_x;  -- CREATE VIEW
  SELECT * FROM iceberg.repro.vw_x ORDER BY id;
  -- "a","Alice"
  -- "b","Bob"
  ```

- [x] **Existing non-view `CreateTable` calls are unaffected (fields default to `None`, matching real Glue behavior).**

  Against the fixed image, creating an `EXTERNAL_TABLE` with no `View*Text` in `TableInput` and then calling `get-table`:
  ```
  ["dim_x", "EXTERNAL_TABLE", null, null]
  ```

- [x] **`pytest tests/test_glue.py::test_glue_view_original_text_roundtrip tests/test_glue.py::test_glue_table_v2`** both pass against the fixed image (new view-roundtrip test + existing non-view table test, confirming no regression to the EXTERNAL_TABLE path).

### DIFF
--- a/ministack/services/glue.py
+++ b/ministack/services/glue.py
@@ -267,6 +267,8 @@ def _create_table(data):
         "PartitionKeys": table_input.get("PartitionKeys", []),
         "TableType": table_input.get("TableType", "EXTERNAL_TABLE"),
         "Parameters": table_input.get("Parameters", {}),
+        "ViewOriginalText": table_input.get("ViewOriginalText"),
+        "ViewExpandedText": table_input.get("ViewExpandedText"),
         "IsRegisteredWithLakeFormation": False,
         "CatalogId": get_account_id(),
     }
--- a/tests/test_glue.py
+++ b/tests/test_glue.py
@@ -123,6 +123,36 @@ def test_glue_table_v2(glue):
         glue.get_table(DatabaseName="glue_tbl_v2db", Name="tbl_v2")
     assert exc.value.response["Error"]["Code"] == "EntityNotFoundException"
 
+def test_glue_view_original_text_roundtrip(glue):
+    glue.create_database(DatabaseInput={"Name": "glue_view_db"})
+    original = "/* Presto View: eyJjYXRhbG9nIjoiaWNlYmVyZyJ9 */"
+    expanded = "/* Presto View */"
+    glue.create_table(
+        DatabaseName="glue_view_db",
+        TableInput={
+            "Name": "vw_x",
+            "TableType": "VIRTUAL_VIEW",
+            "ViewOriginalText": original,
+            "ViewExpandedText": expanded,
+        },
+    )
+    resp = glue.get_table(DatabaseName="glue_view_db", Name="vw_x")
+    assert resp["Table"]["ViewOriginalText"] == original
+    assert resp["Table"]["ViewExpandedText"] == expanded
+
+    glue.update_table(
+        DatabaseName="glue_view_db",
+        TableInput={
+            "Name": "vw_x",
+            "TableType": "VIRTUAL_VIEW",
+            "ViewOriginalText": original + " v2",
+            "ViewExpandedText": expanded + " v2",
+        },
+    )
+    resp2 = glue.get_table(DatabaseName="glue_view_db", Name="vw_x")
+    assert resp2["Table"]["ViewOriginalText"] == original + " v2"
+    assert resp2["Table"]["ViewExpandedText"] == expanded + " v2"
+
 def test_glue_list_v2(glue):
     glue.create_database(DatabaseInput={"Name": "glue_lst_v2db"})
     glue.create_table(


Nahuel990 (approved):

Nahuel990:
Thanks for this @yonatoasis ! and welcome to Ministack!

PR #713, opened 2026-05-26
Add account-region scoped persistence support

Areson:
## Summary

Adds the service-neutral persistence foundation needed for later multi-region service work.

This PR:

- Adds `AccountRegionScopedDict`, a dict-like state container that scopes keys by account ID and request region.
- Adds JSON persistence encode/decode support for account+region scoped state using a distinct `__account_region_scoped__` marker.
- Preserves existing `AccountScopedDict` behavior and its existing `__scoped__` persistence format.
- Adds scoped helper methods (`get_scoped`, `set_scoped`, `contains_scoped`, `pop_scoped`) so follow-up service PRs can access explicit account/region state without changing request context.
- Handles legacy account-scoped/plain persisted state by inferring region from ARN-shaped keys or values when possible.
- Adds regression coverage for same-name regional isolation, tuple resource keys, legacy ARN-region adoption, and key-ARN precedence.

No services are converted to use `AccountRegionScopedDict` in this PR.

## Validation

- `ruff check ministack/core/responses.py ministack/core/persistence.py tests/test_persistence.py`
- `PYTHONPATH=$PWD .../python -m pytest tests/test_persistence.py -v`
  - `128 passed`

### DIFF
--- a/ministack/core/persistence.py
+++ b/ministack/core/persistence.py
@@ -10,7 +10,7 @@ import logging
 import os
 import tempfile
 
-from ministack.core.responses import AccountScopedDict
+from ministack.core.responses import AccountRegionScopedDict, AccountScopedDict
 
 logger = logging.getLogger("persistence")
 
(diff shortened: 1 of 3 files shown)


Nahuel990:
Awesome!

PR #750, opened 2026-05-30
fix(apigateway): emit Set-Cookie and multiValueHeaders from Lambda-proxy responses

### LINKED ISSUE #749: API Gateway HTTP API drops Set-Cookie headers
### Affected Service

API Gateway (v2)

### What happened?

A Lambda function returning the API Gateway v2 response envelope with a
`cookies` array (the documented contract for emitting multiple `Set-Cookie`
headers) produces zero `Set-Cookie` headers on the HTTP response when invoked
through an HTTP API integration. The same Lambda's envelope fetched via direct
`Invoke` is correct, so the loss is in the APIGW response translation, not the
Lambda. The v1 REST API equivalent (`multiValueHeaders`) has the same problem.
Other response headers (e.g. `Content-Type`) pass through fine. Real AWS APIGW
honors both fields.

### Reproduction steps

```python
# Env: AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1
# MiniStack running on http://localhost:4566
import io, zipfile, urllib.request, boto3

EP = "http://localhost:4566"
lam = boto3.client("lambda", endpoint_url=EP)
gw  = boto3.client("apigatewayv2", endpoint_url=EP)

code = (
    b"def handler(event, context):\n"
    b"    return {\n"
    b"        'statusCode': 200,\n"
    b"        'cookies': ['session=abc123; Path=/; HttpOnly', 'flag=on; Path=/'],\n"
    b"        'body': 'cookies should be in the response',\n"
    b"    }\n"
)
buf = io.BytesIO()
with zipfile.ZipFile(buf, "w") as z:
    z.writestr("index.py", code)

lam.create_function(
    FunctionName="cookie-repro", Runtime="python3.12",
    Role="arn:aws:iam::000000000000:role/lambda-role",
    Handler="index.handler", Code={"ZipFile": buf.getvalue()},
)

# Direct invoke proves the Lambda envelope is correct (cookies present):
import json
out = lam.invoke(FunctionName="cookie-repro")["Payload"].read()
print("envelope:", json.loads(out).get("cookies"))   # -> ['session=...', 'flag=on; ...']

# Wire an HTTP API (payload format 2.0) in front of it:
api = gw.create_api(Name="cookie-repro-api", ProtocolType="HTTP")["ApiId"]
integ = gw.create_integration(
    ApiId=api, IntegrationType="AWS_PROXY",
    IntegrationUri="arn:aws:lambda:us-east-1:000000000000:function:cookie-repro",
    PayloadFormatVersion="2.0",
)["IntegrationId"]
gw.create_route(ApiId=api, RouteKey="ANY /{proxy+}", Target=f"integrations/{integ}")
gw.create_stage(ApiId=api, StageName="$default", AutoDeploy=True)

# Invoke through APIGW (path-based execute-api form, no DNS needed):
resp = urllib.request.urlopen(f"{EP}/_aws/execute-api/{api}/anything")
print("Set-Cookie headers:", resp.headers.get_all("Set-Cookie"))
# BUG: prints None  (expected: ['session=abc123; Path=/; HttpOnly', 'flag=on; Path=/'])
```

### Expected behavior

The APIGW HTTP response contains two `Set-Cookie` headers — `session=abc123; …`
and `flag=on; …` — one per entry in the Lambda's `cookies` array, matching real
AWS APIGW v2. (And for v1, one `Set-Cookie` per `multiValueHeaders` entry.)
Currently zero are emitted; other headers like `Content-Type` pass through fine.

### MiniStack version

1.3.52

### SDK / tool used

boto3 1.34 / AWS CLI v2

rmlasseter:
Fixes #749

## Problem

Multi-value response headers from a Lambda-proxy integration never reached the
client. HTTP API (v2) dropped the `cookies` array; REST API (v1) dropped
`multiValueHeaders`. And `app.py:_send_response` serialized headers through a
`dict[str, str]`, so a single header name could carry only one value anyway —
making multiple `Set-Cookie` headers impossible regardless. Any cookie-setting
flow (OIDC, BFF sessions, CSRF/consent pairs) silently broke. Real AWS APIGW
honors both fields.

## Fix

- **`app.py`** — `_send_response` expands a list/tuple header value into one
  header line per item (RFC 6265 §3 forbids comma-folding `Set-Cookie`).
  Scalar values are unchanged, so no other response path is affected.
- **`services/apigateway.py`** — v2 builder maps the format-2.0 `cookies`
  array to a `Set-Cookie` list value.
- **`services/apigateway_v1.py`** — v1 builder folds format-1.0
  `multiValueHeaders` into the response (winning over `headers` on key
  collision, matching AWS).

## Tests

Three regression tests in the existing execute-api idiom: v2 `cookies` → two
distinct `Set-Cookie` headers, v2 empty `cookies` → none emitted, v1
`multiValueHeaders` → one line per value. Full API Gateway suite green
(158 tests), `ruff` clean. Scope kept narrow — related divergences found while
diagnosing (response `isBase64Encoded`, duplicate `Content-Type` on header-name
case mismatch) are left for separate follow-ups.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -474,7 +474,17 @@ async def _send_response(send, status, headers, body):
     body_bytes = body if isinstance(body, bytes) else body.encode("utf-8")
     if "content-length" not in {k.lower() for k in headers}:
         headers["Content-Length"] = str(len(body_bytes))
-    header_list = [(k.encode("latin-1"), _encode_header_value(str(v))) for k, v in headers.items()]
+    # A list/tuple header value expands to one header line per item. This is
+    # required for Set-Cookie, which RFC 6265 §3 forbids folding into a single
+    # comma-joined header; APIGW Lambda-proxy responses surface multiple
+    # cookies this way. Scalar values keep their existing single-line behavior.
+    header_list = []
+    for k, v in headers.items():
+        if isinstance(v, (list, tuple)):
+            for item in v:
+                header_list.append((k.encode("latin-1"), _encode_header_value(str(item))))
+        else:
+            header_list.append((k.encode("latin-1"), _encode_header_value(str(v))))
     await send(
         {
             "type": "http.response.start",
--- a/ministack/services/apigateway.py
+++ b/ministack/services/apigateway.py
@@ -1005,6 +1005,12 @@ async def _invoke_lambda_proxy(
     status = lambda_response.get("statusCode", 200)
     resp_headers = {"Content-Type": "application/json"}
     resp_headers.update(lambda_response.get("headers", {}))
+    # Payload format 2.0 emits multiple Set-Cookie headers via the top-level
+    # `cookies` array, not the headers map. The list value is expanded into one
+    # Set-Cookie line per entry by _send_response.
+    cookies = lambda_response.get("cookies")
+    if cookies:
+        resp_headers["Set-Cookie"] = list(cookies)
     resp_body = lambda_response.get("body", "")
     if isinstance(resp_body, str):
         resp_body = resp_body.encode("utf-8")
--- a/ministack/services/apigateway_v1.py
+++ b/ministack/services/apigateway_v1.py
@@ -977,6 +977,13 @@ async def _invoke_lambda_proxy_v1(integration, api_id, stage_name, stage, resour
     status = lambda_response.get("statusCode", 200)
     resp_headers = {"Content-Type": "application/json"}
     resp_headers.update(lambda_response.get("headers", {}))
+    # Payload format 1.0 carries multi-value headers (notably Set-Cookie) in
+    # `multiValueHeaders`. AWS merges these over `headers`, with multiValueHeaders
+    # winning on key collision. Each list value is expanded into one header line
+    # per entry by _send_response.
+    for k, v in (lambda_response.get("multiValueHeaders") or {}).items():
+        if v:
+            resp_headers[k] = list(v)
     resp_body = lambda_response.get("body", "")
     if isinstance(resp_body, str):
         resp_body = resp_body.encode("utf-8")
--- a/tests/test_apigatewayv1.py
+++ b/tests/test_apigatewayv1.py
@@ -498,6 +498,70 @@ def test_apigwv1_execute_lambda_proxy(apigw_v1, lam):
     lam.delete_function(FunctionName=fname)
 
 
+def test_apigwv1_execute_lambda_proxy_multi_value_headers(apigw_v1, lam):
+    """Payload format 1.0 `multiValueHeaders` yields one header line per value.
+
+    Real APIGW v1 carries multi-value headers (notably Set-Cookie) in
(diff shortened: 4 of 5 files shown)


Nahuel990:
## Review

Verified the diff against AWS docs (HTTP API v2 `cookies` array, REST/HTTP v1 `multiValueHeaders`) and walked the three layers — description, code, AWS contract — line by line.

**AWS contract is correctly stated:**
- v2 format `cookies` array → one `Set-Cookie` header per entry (RFC 6265 §3 forbids folding)
- v1 format `multiValueHeaders` → one header line per value; wins over `headers` on key collision

The code matches both.

**Code:**

1. `app.py:_send_response` — `list`/`tuple` value expands to N header lines, scalars unchanged. Cleanest possible change: zero impact on existing call sites (every other emitter passes scalars), unlocks the two call sites below. The `_encode_header_value(str(item))` wrap keeps the existing per-line CRLF stripping. ✓
2. `apigateway.py` (v2) — `resp_headers["Set-Cookie"] = list(cookies)` uses `list()` to avoid sharing the lambda-response list; `if cookies:` correctly skips the empty-array case (no empty `Set-Cookie:` header). ✓
3. `apigateway_v1.py` (v1) — iterating `multiValueHeaders` AFTER `resp_headers.update(headers)` lets MVH win on collision, matching the AWS contract. `if v:` skips empty value lists. ✓

**Tests:** end-to-end via real Lambda invocation through `{api_id}.execute-api.localhost`, asserting on `resp.headers.get_all("Set-Cookie")` (the right way to read multi-value headers in `urllib`). Covers happy path (2 cookies) + the empty-cookies-no-empty-header edge case + v1 MVH happy path. 158/158 APIGW tests stayed green per the PR description.

**Scope discipline:** explicitly carved out the response `isBase64Encoded` handling and the case-sensitive `Content-Type` dedup as separate follow-ups. Good call.

**Two non-blocking observations** (worth a follow-up issue, not this PR):

1. **v2: `headers.Set-Cookie` (scalar) + `cookies` array both present.** Per AWS docs ("If you specify cookies in the response headers parameter as well, those cookies are also sent"), AWS emits both. The new code overwrites the scalar with the array. Edge case the reported bug doesn't touch — flagging for completeness.
2. **v1: case-collision dedup.** If `headers` has `Set-Cookie` (scalar) and `multiValueHeaders` has `set-cookie` (list), `resp_headers` ends up with both keys, both ship. The PR's author already flagged this in the description as out of scope. Same comment, surfaced here.

**Merge.** Strong PR for a first-time contributor — single concern, AWS-doc-cited comments, end-to-end tests, narrow scope. The two observations above are real but unrelated to the reported bug (#749) and can land as a follow-up.

Nahuel990 (approved):

Nahuel990:
Thanks @rmlasseter for this! And welcome to Ministack!

PR #756, opened 2026-05-30
feat(kms): support Ed25519 keys

KABBOUCHI:


Nahuel990:
HI @KABBOUCHI , thansk for this! AWS's wire value for Ed25519 is `ED25519_SHA_512`, not `EDDSA`

`EdDSA` appears in the AWS documentation as the algorithm description, but it is **not** a valid API enum value. The API's `SigningAlgorithm` parameter does not accept `EDDSA`; clients that send `SigningAlgorithm=EDDSA` to AWS KMS will be rejected.
The metadata    array should be ["ED25519_SHA_512", "ED25519_PH_SHA_512"]

Review how AWS handle `ED25519_PH_SHA_512` (used when `MessageType=DIGEST`)
Either implement it correctly or return `UnsupportedOperationException`. Do not route it through the standard Ed25519 implementation, as the resulting signatures will be incompatible with AWS KMS.

**For extra guidance you can cehck:**
- KMS Developer Guide: *Supported signing algorithms for ECC key specs*
- KMS API Reference: *Sign* (`API_Sign`) valid values list

https://docs.aws.amazon.com/kms/latest/developerguide/symm-asymm-choose-key-spec.html

Nahuel990:
@KABBOUCHI I am merging this because it's been a while since it's been open, I will complete it before release, thanks for the work

Nahuel990 (approved):

PR #757, opened 2026-05-30
Normalize Step Functions RDS SDK errors

Areson:
## Why

Step Functions SDK integrations should surface RDS failures with AWS-shaped error names. Without that normalization, workflows can receive generic MiniStack errors instead of the `Rds.*` error names returned by AWS Step Functions SDK integrations.

## What

- Normalize Step Functions RDS SDK integration failures to `Rds.*` error names.
- Add regression coverage for RDS SDK error mapping through Step Functions.
- Require Docker-backed database readiness probes to execute an authenticated `SELECT 1` before marking instances available.
- Preserve AWS-shaped `DescribeDBClusters.Port` values while still syncing cluster endpoint hosts to backing local instances.

## Validation

- `ruff check ministack/services/stepfunctions.py tests/test_stepfunctions.py ministack/services/rds.py tests/test_rds.py tests/test_rds_password_rotation.py`
- `pytest tests/test_rds.py tests/test_rds_password_rotation.py -q`

### DIFF
--- a/ministack/services/rds.py
+++ b/ministack/services/rds.py
@@ -433,12 +433,19 @@ def _grant_mysql_master_user_privileges(host, port, master_user, master_pass, db
 
 
 def _try_database_connect(host, port, engine, user, password, db_name):
-    """Single auth-probe attempt. Returns True on success, False on a
-    transient failure. TCP readiness alone is not enough for MySQL/Postgres
-    images — they accept sockets before bootstrap creates users/databases —
-    so we open and close an authenticated connection. When the DB driver
-    isn't installed (lightweight image) we fall back to a one-shot TCP check.
+    """Single auth + query probe attempt.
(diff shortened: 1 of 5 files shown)


Nahuel990:
@Areson sorry for the late reply... I see you are Rreturning `3306`/`5432` (which matches the AWS API shape) but it breaks reachability through `cluster.Endpoint:cluster.Port`... Under the multi-region scoping model, the local instance's mapped Docker port is the only port a client can actually connect to.

   Can you either:

   * Publish the Docker container on the engine-default host port (so `Port=3306` is both AWS-shaped and functional), or
   * Keep the existing sync behavior?

   Hardcoding the port without the corresponding host-side fix will break current users

Nahuel990:
Thanks!

PR #758, opened 2026-05-30
fix(apigateway): v2 Lambda-proxy emits both header and array cookies, overrides case-mismatched headers

rmlasseter:
Follow-up to #750 — addresses the two v2 observations from that PR's review.

## Problem

Two parity gaps remained in the HTTP API (v2) Lambda-proxy response-header
assembly:

1. ~~A `Set-Cookie` carried in the response `headers` map was **overwritten** by the `cookies` array (`resp_headers["Set-Cookie"] = list(cookies)`). Per the AWS docs, both are sent — so the header cookie was silently dropped.~~

   **Correction (thanks @Nahuel990):** the docs don't support "both are sent." For v2 the `cookies` array is the documented cookie mechanism — *"don't manually add set-cookie headers; use the cookies array"* ([Lambda Function URLs](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-cookies) / [HTTP API v2](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) docs, same payload schema). So the `cookies` array should **supersede** any `Set-Cookie` carried in `headers` (case-insensitive), matching #759's v1 rule where the structured source wins on collision.
2. A header returned in different letter-case than a seeded default — notably a
   lowercase `content-type` vs MiniStack's default `Content-Type` — produced
   two dict keys and shipped **twice**, default first. HTTP field names are
   case-insensitive (RFC 9110 §5.1); the function's value should win as one
   header.

## Fix (`apigateway.py`)

- Apply the Lambda's `headers` with case-insensitive override of any seeded
  default, mirroring the `multiValueHeaders` merge introduced for v1 in #750.
- When a `cookies` array is present, case-insensitively drop any `Set-Cookie`
  from `headers`, then emit the array — the array **supersedes** a hand-set
  `Set-Cookie`.

## Tests

- `headers` `Set-Cookie` + `cookies` array → only the array's cookies ship
  (the header `Set-Cookie` is superseded).
- lowercase `content-type` → a single overriding `Content-Type`.

Full API Gateway suite green; scope limited to the v2 builder.

### DIFF
--- a/ministack/services/apigateway.py
+++ b/ministack/services/apigateway.py
@@ -1004,12 +1004,27 @@ async def _invoke_lambda_proxy(
 
     status = lambda_response.get("statusCode", 200)
     resp_headers = {"Content-Type": "application/json"}
-    resp_headers.update(lambda_response.get("headers", {}))
-    # Payload format 2.0 emits multiple Set-Cookie headers via the top-level
-    # `cookies` array, not the headers map. The list value is expanded into one
-    # Set-Cookie line per entry by _send_response.
+    # Apply the Lambda's headers, letting each override any case-insensitive
+    # default already present. HTTP field names are case-insensitive (RFC 9110
+    # §5.1), so a lowercase `content-type` from the function must replace the
+    # seeded `Content-Type`, not ship alongside it — the same case-insensitive
+    # handling #750 applied to the v1 multiValueHeaders merge.
+    for k, v in (lambda_response.get("headers") or {}).items():
+        lower_k = k.lower()
+        for existing in [h for h in resp_headers if h.lower() == lower_k]:
+            del resp_headers[existing]
+        resp_headers[k] = v
+    # Payload format 2.0 delivers cookies via the top-level `cookies` array,
+    # which AWS turns into one Set-Cookie header per entry. The array is the
+    # documented cookie mechanism — AWS guidance is "don't manually add
+    # set-cookie headers; use the cookies array" (Lambda Function URLs / HTTP
+    # API v2 docs) — so when it is present it supersedes any Set-Cookie carried
+    # in `headers` (the structured source wins, as in #750's v1 collision rule).
+    # _send_response expands the list into one Set-Cookie line per entry.
     cookies = lambda_response.get("cookies")
     if cookies:
+        for existing in [h for h in resp_headers if h.lower() == "set-cookie"]:
+            del resp_headers[existing]
         resp_headers["Set-Cookie"] = list(cookies)
     resp_body = lambda_response.get("body", "")
     if isinstance(resp_body, str):
--- a/tests/test_apigatewayv2.py
+++ b/tests/test_apigatewayv2.py
@@ -461,6 +461,115 @@ def test_apigw_execute_lambda_proxy_empty_cookies(apigw, lam):
     apigw.delete_api(ApiId=api_id)
     lam.delete_function(FunctionName=fname)
 
+def test_apigw_execute_lambda_proxy_cookies_supersede_header_set_cookie(apigw, lam):
+    """The `cookies` array supersedes a `Set-Cookie` carried in `headers`.
+
+    Follow-up to #750. For v2 the `cookies` array is the documented cookie
+    mechanism ("don't manually add set-cookie headers" — AWS Lambda Function
+    URLs / HTTP API v2 docs), so when both are returned only the array's cookies
+    ship — mirroring #759's v1 rule where the structured source wins on a
+    case-insensitive collision.
+    """
+    import urllib.request as _urlreq
+    import uuid as _uuid
+
+    fname = f"intg-apigw-cookiewin-{_uuid.uuid4().hex[:8]}"
+    code = (
+        b"def handler(event, context):\n"
+        b"    return {\n"
+        b"        'statusCode': 200,\n"
+        b"        'headers': {'set-cookie': 'should-be-dropped=1; Path=/'},\n"
+        b"        'cookies': ['a=1; Path=/', 'b=2; Path=/'],\n"
+        b"        'body': 'x',\n"
+        b"    }\n"
+    )
+    buf = io.BytesIO()
+    with zipfile.ZipFile(buf, "w") as zf:
+        zf.writestr("index.py", code)
+    lam.create_function(
+        FunctionName=fname, Runtime="python3.12",
+        Role="arn:aws:iam::000000000000:role/test-role",
+        Handler="index.handler", Code={"ZipFile": buf.getvalue()},
+    )
+
+    api_id = apigw.create_api(Name=f"cookiewin-api-{fname}", ProtocolType="HTTP")["ApiId"]
+    int_id = apigw.create_integration(
+        ApiId=api_id, IntegrationType="AWS_PROXY",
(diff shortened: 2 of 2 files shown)


Nahuel990 (dismissed):
I'm sorry,

I tried checking this doc https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html 

but I dont see that claim from spot 1 

A Set-Cookie carried in the response headers map was overwritten by
the cookies array (resp_headers["Set-Cookie"] = list(cookies)). Per the
AWS docs, both are sent — so the header cookie was silently dropped.

  - The cookies array is the documented way to emit multiple Set-Cookie headers.
  - The response format example shows cookies and headers as separate fields.
  - The docs do NOT address what happens when a function returns both cookies array AND Set-Cookie inside        
  headers. That case is simply not covered. 
  
  
  Can you point me from which doc are you retrieving it? Do you have the command/cli/tf where this is created? I can check it in a few days

rmlasseter:
You're right, and thanks for pushing on this — I went back to the docs and
couldn't substantiate the "both are sent" claim either. The
[HTTP API v2 Lambda proxy page](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)
treats `cookies` and `headers` as separate response fields but never addresses
the both-present case, and the closest authoritative guidance is the Lambda
[Function URLs doc](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-cookies)
(same v2 payload schema): *"To return cookies from your function, don't manually
add set-cookie headers. Instead, include the cookies in your response payload
object."* So the array is authoritative and a hand-set `Set-Cookie` shouldn't be
treated as additive.

I've revised the PR: when a `cookies` array is present it now **supersedes** any
`Set-Cookie` in `headers` (case-insensitive) — which also lines up with the v1
collision rule from #759 where the structured source wins. The test now asserts
the array wins (mirroring your `mvh_wins_case_insensitive` test) instead of the
old "both" assertion.

The other half of the PR — collapsing the duplicate `Content-Type` from a
case-mismatched header (RFC 9110 §5.1) — is unchanged. Force-pushed the revision.

rmlasseter:
Following up — this was nagging at me, so rather than keep hunting the docs for an authoritative reference, I ran a both-present case against **real AWS** to get ground truth. Sharing the setup, the result, and the genuine ambiguity, because I think the *desired* MiniStack behavior is a call for you to make.

### Test: Lambda returning BOTH a header `Set-Cookie` and a `cookies` array, behind a real HTTP API v2 (payload 2.0, AWS_PROXY)

<details><summary>Reproducible setup (AWS CLI, self-cleaning)</summary>

```bash
REGION=us-east-1; SFX=$(date +%s); FN="cookie-probe-$SFX"; ROLE="cookie-probe-role-$SFX"
sleep 12
cat > /tmp/probe.py <<'PY'
def handler(event, context):
    return {"statusCode": 200,
            "headers": {"set-cookie": "h=1; Path=/", "content-type": "text/plain"},
            "cookies": ["a=1; Path=/", "b=2; Path=/"], "body": "probe"}
PY
(cd /tmp && zip -q probe.zip probe.py)
LAMBDA_ARN=$(aws lambda create-function --function-name "$FN" --runtime python3.12 \
  --role "$ROLE_ARN" --handler probe.handler --zip-file fileb:///tmp/probe.zip \
  --region "$REGION" --query FunctionArn --output text)
read API_ID API_EP < <(aws apigatewayv2 create-api --name "$FN-api" --protocol-type HTTP \
  --target "$LAMBDA_ARN" --region "$REGION" --query '[ApiId,ApiEndpoint]' --output text)
aws lambda add-permission --function-name "$FN" --statement-id "apigw-$SFX" \
  --action lambda:InvokeFunction --principal apigateway.amazonaws.com \
  --source-arn "arn:aws:execute-api:$REGION:$ACCT:$API_ID/*/*" --region "$REGION" >/dev/null
sleep 5; curl -sS -D - -o /dev/null "$API_EP/"
# cleanup
aws apigatewayv2 delete-api --api-id "$API_ID" --region "$REGION"
```
</details>

**Result (on the wire):**

```
HTTP/2 200
content-type: text/plain
set-cookie: a=1; Path=/
set-cookie: b=2; Path=/
set-cookie: h=1; Path=/
```

So live AWS emits **all three** — the `cookies` array entries first, then the header `Set-Cookie`. (The single `content-type: text/plain` also confirms the case-insensitive header half of this PR, independent of the cookie question.)

### Where the docs land

- **v1 is unambiguous** — the `headers` + `multiValueHeaders` merge is documented (merged into one list; `multiValueHeaders` wins on collision). #759 matches that.
- **v2 is *not* clear** — the docs describe the `cookies` array → `set-cookie` conversion but never address the both-present case, and they explicitly steer toward using `cookies` exclusively: *"don't manually add set-cookie headers … include the cookies in your response payload object."* So the observed emit-both behavior is real, but it's the result of doing something the docs tell you not to do.

### The question for you

Given that tension, which behavior do you want MiniStack to emulate for v2?

1. **Match observed AWS** — emit both (array first, then the header `Set-Cookie`).
2. **Follow the documented guidance** — treat `cookies` as authoritative and drop a hand-set `Set-Cookie` (i.e. cookies-exclusive), accepting a small divergence from live AWS in a discouraged edge case.

I have both implemented and tested locally and can push whichever you prefer — the Content-Type case-fix is unaffected either way. Happy to go with your call.

rmlasseter:
Quick clarification on the current state of the diff, since I've gone back and forth on this: the **tip of the branch right now is option 2 — cookies supersedes** the header `Set-Cookie`, while the **original commit is option 1 — emit both**. Both commits also carry the *other* fix in this PR — collapsing the duplicate `Content-Type` from a case-mismatched header — so that part is identical regardless of which cookie approach is selected.

Let me know which you'd prefer and I'll set the branch to match.

Nahuel990:
@rmlasseter thanks for this and sorry for the back and forth... Yes I was checking too and let's go with AWS behaviour. (It happens they document a different behaviour, we always pick whatever AWS does instead document) I will merge this and apply your previous fix to avoid the round trip. Thanks for checking too!

Nahuel990 (approved):

rmlasseter:
Quick heads-up on the merge: the commit at the top of my branch was the **supersede** version (`218d3a9`), so that's what got squash-merged — the emit-both version didn't make it in. With the back-and-forth while we were still working out which behavior was actually right, the branch happened to be sitting on supersede at merge time; sorry for the churn there.

The upshot is that `main` currently does the supersede behavior (a `Set-Cookie` in `headers` is dropped when a `cookies` array is also present), whereas we'd landed on matching AWS, which emits both. It's a tiny corner — AWS steers people away from mixing the two — so absolutely no worries if you'd rather leave it as-is. But I have the emit-both change ready and am happy to open a small follow-up PR (this one's merged, so it'd be a new one) if that's useful.

Either way — thank you (and the maintainers generally) for the time and care you put into MiniStack. It's been hugely helpful to build against, and I appreciate you working through this one with me.

Nahuel990:
@rmlasseter yes, I will add it before release in ~2 hours, I didn't want to ask you for another commit as it's already there  I just need to cherry pick it before release

PR #759, opened 2026-05-30
fix(apigateway): v1 Lambda-proxy overrides case-mismatched default headers

rmlasseter:
Follow-up to #750.

## Problem

The `multiValueHeaders` merge added in #750 (by @Nahuel990) case-folds
collisions, but only for keys present in `multiValueHeaders`. A `content-type`
carried in the plain `headers` map still collided with the seeded default
`Content-Type` and shipped **twice**. HTTP field names are case-insensitive
(RFC 9110 §5.1), so the function's header should override the default and be
emitted once.

## Fix (`apigateway_v1.py`)

Apply the same case-folding @Nahuel990 used for the `multiValueHeaders` merge to
the scalar `headers` merge, so a lowercase `content-type` replaces the default
`Content-Type` rather than duplicating it.

## Tests

- lowercase `content-type` → a single overriding `Content-Type`.

Full API Gateway suite green; one-function change to the v1 builder.

### DIFF
--- a/ministack/services/apigateway_v1.py
+++ b/ministack/services/apigateway_v1.py
@@ -976,7 +976,16 @@ async def _invoke_lambda_proxy_v1(integration, api_id, stage_name, stage, resour
 
     status = lambda_response.get("statusCode", 200)
     resp_headers = {"Content-Type": "application/json"}
-    resp_headers.update(lambda_response.get("headers", {}))
+    # Apply the Lambda's `headers` with case-insensitive override of any seeded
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):
Approve.

Clean follow-up to #750 — applies the same case-insensitive override pattern to the scalar `headers` merge that we already do for `multiValueHeaders`. RFC 9110 §5.1 justification is correct. Test is targeted and asserts the right thing (`get_all("Content-Type")` == single value).

LGTM.

Nahuel990:
Thanks for this!

PR #762, opened 2026-05-31
build: Add curl to docker image

Baptiste-Garcin:
Hi !

I need cURL in the container a the health check during my github CI


Nahuel990:
Hi @Baptiste-Garcin Ministack doesn't need curl, if you are buidling a local project that runs against Ministack you should set that dependency in your own Dockerfile

Baptiste-hero:
Hello, I have e2e tests running in a github runner. Those relies on ministack for SNS/SQS (and later for SSM). In my github workflow, I need to add a health check on ministack before starting the tests. 

Doing something like this

```
      ministack:
        image: ministackorg/ministack-preview-build:pr-762-afc651c0
        ports:
          - "127.0.0.1:4568:4566"
        options: >-
          --health-cmd "curl -f http://localhost:4566/_ministack/ready || exit 1"
          --health-interval 10s
          --health-timeout 5s
          --health-retries 5
          --health-start-period 15s
```

But, weird enough, in this specific case, the curl I use must be inside the ministack container.

Nahuel990:
This should fall back to the MiniStack health check on line 73 (which is raw Python). We deliberately strip all HTTP client tooling from the base image, including wget, to reduce the attack surface and eliminate the most common entrypoint  for attack / abuse

With the same pattern should work 

--health-cmd "python -c 'import urllib.request as r; r.urlopen(\"http://localhost:4566/_ministack/ready\")'"

Baptiste-hero:
Okay, really clear, thanks. I can't close the PR on my side...

PR #766, opened 2026-06-01
fix(apigateway): v2 Lambda-proxy honors isBase64Encoded in both directions

rmlasseter:
HTTP API (v2) Lambda-proxy ignored binary payloads on both sides. This adds `isBase64Encoded` handling in both directions.

## Response
A body with `isBase64Encoded: true` was emitted as the literal base64 string. It's now decoded to raw bytes before sending. `isBase64Encoded` is part of the documented v2 response payload format (refs below), and HTTP APIs honor it unconditionally — there is no `binaryMediaTypes` negotiation like REST/v1.

## Request
The inbound body was always `body.decode("utf-8")` with `isBase64Encoded: false`, corrupting binary uploads. It's now base64-encoded with `isBase64Encoded: true` for non-text content types.

The docs establish the mechanism — *"the body of the request. If the content type of the request is binary, the body is base64-encoded"* / *"`isBase64Encoded`: TRUE if the body is a binary payload and base64-encoded"* — but they **don't enumerate which content types count as binary**. So I **verified the split against real AWS** (HTTP API v2, payload 2.0, AWS_PROXY). The body is delivered as a UTF-8 string (`isBase64Encoded: false`) only for `text/*` and `application/json` / `application/xml` / `application/javascript`; everything else — including a missing `Content-Type` and (surprisingly) `application/x-www-form-urlencoded` — is base64-encoded.

| Content-Type sent | `isBase64Encoded` observed |
|---|---|
| `application/json`, `text/plain`, `text/html`, `application/xml`, `application/javascript` | **false** (UTF-8 string) |
| `application/octet-stream`, `image/png`, `image/jpeg`, `application/pdf`, `application/x-www-form-urlencoded`, *(no Content-Type)* | **true** (base64) |

<details><summary>Live-AWS verification (reproducible, self-cleaning)</summary>

```bash
#!/usr/bin/env bash
REGION=us-east-1; SFX=$(date +%s); FN="b64req-$SFX"; ROLE="b64req-role-$SFX"
sleep 12
cat > /tmp/echo.py <<'PY'
import json, base64, hashlib
def handler(event, context):
    b=event.get("body"); isb=bool(event.get("isBase64Encoded"))
    raw=base64.b64decode(b) if (isb and b) else (b.encode("utf-8","surrogateescape") if b else b"")
    return {"statusCode":200,"headers":{"content-type":"application/json"},
            "body":json.dumps({"isB64":isb,"sha":hashlib.sha256(raw).hexdigest()[:12]})}
PY
(cd /tmp && zip -q echo.zip echo.py)
ARN=$(aws lambda create-function --function-name "$FN" --runtime python3.12 --role "$ROLE_ARN" \
  --handler echo.handler --zip-file fileb:///tmp/echo.zip --region "$REGION" --query FunctionArn --output text)
read API EP < <(aws apigatewayv2 create-api --name "$FN-api" --protocol-type HTTP --target "$ARN" \
  --region "$REGION" --query '[ApiId,ApiEndpoint]' --output text)
aws lambda add-permission --function-name "$FN" --statement-id "agw-$SFX" --action lambda:InvokeFunction \
  --principal apigateway.amazonaws.com --source-arn "arn:aws:execute-api:$REGION:$ACCT:$API/*/*" --region "$REGION" >/dev/null
sleep 5; printf '\x00\x01\x02\xfe\xff' > /tmp/p.bin
for CT in application/json text/plain application/xml application/javascript \
          application/x-www-form-urlencoded application/octet-stream image/png application/pdf; do
  printf '%-36s ' "$CT"; curl -s -X POST -H "Content-Type: $CT" --data-binary @/tmp/p.bin "$EP/"; echo
done
printf '%-36s ' "(no content-type)"; curl -s -X POST --data-binary @/tmp/p.bin "$EP/"; echo
# cleanup
aws apigatewayv2 delete-api --api-id "$API" --region "$REGION"
```
</details>

## References (AWS docs)
- [Create AWS Lambda proxy integrations for HTTP APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) — the payload-format-2.0 request and response shapes both include `isBase64Encoded`.
- [Invoking Lambda function URLs](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html) — same v2 payload schema; documents that a binary request body is base64-encoded and `isBase64Encoded` is set, and that the response `isBase64Encoded` flag is honored. (Neither page enumerates the text-vs-binary content types, hence the live test.)

## Tests
- base64 response body → decoded to raw bytes on the wire
- binary request body (`application/octet-stream`) → delivered as base64, round-tripping the raw bytes; a text body (`application/json`) → passed through as a UTF-8 string

Scope is v2 only; the REST (v1) binary path (`binaryMediaTypes`-driven) is a separate PR.

### DIFF
--- a/ministack/services/apigateway.py
+++ b/ministack/services/apigateway.py
@@ -921,6 +921,22 @@ def _path_matches(route_path: str, request_path: str) -> bool:
     return _extract_path_params(route_path, request_path) is not None
 
 
+def _v2_request_body_is_text(content_type: str | None) -> bool:
+    """Whether an HTTP API (v2) request body is delivered as a UTF-8 string.
+
+    Real AWS delivers the body as text (``isBase64Encoded`` false) only for a
+    whitelist of content types; everything else — including a missing content
+    type and ``application/x-www-form-urlencoded`` — is base64-encoded. The docs
+    don't enumerate this, so the set below was verified against live AWS.
+    """
+    ct = (content_type or "").split(";", 1)[0].strip().lower()
+    return ct.startswith("text/") or ct in (
+        "application/json",
+        "application/xml",
+        "application/javascript",
+    )
+
+
 async def _invoke_lambda_proxy(
     integration,
     api_id,
@@ -957,6 +973,15 @@ async def _invoke_lambda_proxy(
     # AWS API Gateway v2 joins multi-value query params with commas
     qs = {k: ",".join(v) for k, v in query_params.items()} if query_params else None
     raw_qs = "&".join(f"{k}={val}" for k, vals in query_params.items() for val in vals)
+    # Binary request bodies are base64-encoded with isBase64Encoded=true; only
+    # the text content types AWS recognizes are passed through as UTF-8 strings.
+    if body:
+        if _v2_request_body_is_text(headers.get("content-type")):
+            req_body, req_is_base64 = body.decode("utf-8", errors="replace"), False
+        else:
+            req_body, req_is_base64 = base64.b64encode(body).decode("ascii"), True
+    else:
+        req_body, req_is_base64 = None, False
     event = {
         "version": "2.0",
         "routeKey": route_key,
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

Nahuel990 (approved):

PR #775, opened 2026-06-02
fix(lambda): fix in x-ray header

Nahuel990:


arivazhaganjeganathan-abc:
> Docker image for this PR has been published: `ministackorg/ministack-preview-build:pr-775-ebe2ce04`

validated with this docker image.. amazon trace id error no longer occurs.

Nahuel990:
Awesome thanks for the feedback! I was expeting that last PR that landed a few minutes ago and I will be releasing this in a few hours

PR #784, opened 2026-06-02
Adopt ARN parsing for SNS subscription endpoints

Areson:
Summary
- Adopt shared ARN parsing for SNS SQS and Lambda subscription endpoints.
- Reject malformed or wrong-service SQS/Lambda endpoint ARNs at subscribe time.
- Resolve SNS-to-SQS fanout by exact stored `QueueArn` so foreign-account or wrong-Region same-name ARNs do not tail-match local queues, while preserving matching cross-Region SQS delivery.

Validation
- `uvx ruff check ministack/services/sns.py tests/test_sns.py`
- `git diff --check`
- Source-backed MiniStack on `GATEWAY_PORT=14563` with focused SNS regression tests: `11 passed`
- Source-backed MiniStack on `GATEWAY_PORT=14563`: `MINISTACK_ENDPOINT=http://localhost:14563 uv run --extra dev pytest tests/test_sns.py -q` (`63 passed`)

### DIFF
--- a/ministack/services/sns.py
+++ b/ministack/services/sns.py
@@ -27,6 +27,7 @@ _HOST = os.environ.get("MINISTACK_HOST", "localhost")
 _PORT = os.environ.get("GATEWAY_PORT", "4566")
 
 import ministack.services.lambda_svc as _lambda_svc
+from ministack.core.arn import ArnParseError, parse_arn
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thankss!

PR #795, opened 2026-06-03
Add /_ministack/ports endpoint for Docker-backed service port discovery

neriyaco:
This endpoint exposes host-port mappings for Docker-backed RDS and ElastiCache instances.

When `DOCKER_NETWORK` is enabled, service APIs may return Docker-internal endpoints that are only reachable from containers on the same network. External clients often need the published host port instead, but this information is not available through the standard AWS APIs.

The new endpoint provides a simple way for test automation, local tooling, and remote clients to discover the correct connection ports without inspecting Docker directly.


Nahuel990:
@neriyaco Can you explain the utility for this? 

If you're running your app on the host you don't to enable DOCKER_NETWORK, Endpoints already return localhost:host_port

If you're running your app in a container, put it in the same Docker network, Endpoints return the internal IP (same as AWS) 

 If you're running tests on the host with Ministack in Docker, is the same as first alternative (dont enable DOCKER_NETWORK)

neriyaco:
It's useful when MiniStack runs on a different machine

Nahuel990:
@neriyaco in that case we'd need to replace the current localhost with _MINISTACK_HOST ( we use it for S3 virtual-host) but it would be also useful here to set which IP you want to use.

neriyaco:
@Nahuel990  Do you mean to add a field "host_address" with the `_MINISTACK_HOST` value? Maybe I misunderstood

Nahuel990:
@neriyaco I'm really sorry, in these files we have hardcoded "localhost", as the entry point. We should replace that hardcode with _MINISTACK_HOST (which is a variable we use already in other services), after that you will be able to set that variable MINISTACK_HOST="ip from the remote computer".

Since we already support describe-db-instances with the hostname, by running the AWS command you will be able to get the host like you would do it in AWS, and remote clients get a reachable address without exposing the internal address of the host in an endpoint. (It's the same with ElasticCache)

neriyaco:
No problem, I'll look where `localhost` needs to be changed and open a separate PR

neriyaco:
Since #797 is merged this PR might not be relevant anymore

Nahuel990:
We are closing it Thanks! I'll be releasing in a few minutes

PR #799, opened 2026-06-03
Adopt ARN parsing for MWAA source buckets

Areson:
## Summary
- Adopt the shared ARN parser for MWAA `SourceBucketArn` validation on create/update.
- Use the parsed S3 bucket ARN resource when syncing DAGs instead of splitting the ARN string.
- Add MWAA regressions for malformed, wrong-service, object-level, and region/account-bearing source bucket ARNs.

## Tests
- `uv run --extra dev ruff check ministack/services/mwaa.py tests/test_mwaa.py`
- `GATEWAY_PORT=4577 MINISTACK_ENDPOINT=http://localhost:4577 uv run --extra dev pytest tests/test_mwaa.py -q` (`18 passed`)

### DIFF
--- a/ministack/services/mwaa.py
+++ b/ministack/services/mwaa.py
@@ -21,6 +21,7 @@ import socket
 import threading
 import time
 
+from ministack.core.arn import ArnParseError, parse_arn
 from ministack.core.persistence import load_state
 from ministack.core.responses import (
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

PR #800, opened 2026-06-03
Adopt ARN parsing for CloudTrail trails

Areson:
## Summary
- Adopt the shared ARN parser for CloudTrail trail ARN resolution and validation.
- Reject malformed, wrong-service, and out-of-scope trail ARNs on mutation/tag paths while preserving cross-request-region reads for stored trail ARNs.
- Add CloudTrail regressions for ARN lookup, malformed identifiers, foreign-region behavior, event selector handling, and tag validation.

## Tests
- `uv run --extra dev ruff check ministack/services/cloudtrail.py tests/test_cloudtrail.py`
- `uv run --extra dev pytest tests/test_cloudtrail.py -q` (`49 passed`)

### DIFF
--- a/ministack/services/cloudtrail.py
+++ b/ministack/services/cloudtrail.py
@@ -20,6 +20,7 @@ import logging
 import os
 import time
 
+from ministack.core.arn import ArnParseError, parse_arn
 from ministack.core.persistence import load_state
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

PR #801, opened 2026-06-03
Adopt ARN parsing for Resource Groups

Areson:
## Summary
- Replace Resource Groups ARN regex handling with the shared ARN parser for group and resource identifiers.
- Validate partition, service, region/account scope, and group resource shape for Resource Groups group ARNs.
- Validate `GroupResources` and `UngroupResources` member ARNs before mutation so failed requests remain atomic.

## Tests
- `uv run --extra dev ruff check ministack/services/resource_groups.py tests/test_resource_groups.py`
- `uv run --extra dev pytest tests/test_resource_groups.py -q` (`23 passed`)

### DIFF
--- a/ministack/services/resource_groups.py
+++ b/ministack/services/resource_groups.py
@@ -21,6 +21,7 @@ import json
 import logging
 import re
 
+from ministack.core.arn import ArnParseError, parse_arn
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

PR #807, opened 2026-06-03
Add Kinesis event source ARN no-fallback regressions

Areson:
## Summary

- add focused Kinesis event source mapping regression coverage for ARN no-fallback behavior
- cover malformed, wrong-service, wrong-account, and wrong-region Kinesis source ARNs while a same-named local stream exists
- verify create-time rejection and poller behavior do not tail-match invalid source ARNs

## Validation

- `uv run --extra dev pytest -q tests/test_lambda.py::test_lambda_create_esm_rejects_invalid_kinesis_arns_without_stream_name_fallback tests/test_lambda.py::test_lambda_kinesis_poller_does_not_tail_match_invalid_event_source_arn`
  - `11 passed`
- `uv run --extra dev ruff check tests/test_lambda.py --extend-ignore E701,I001`

### DIFF
--- a/tests/test_lambda.py
+++ b/tests/test_lambda.py
@@ -3640,6 +3640,151 @@ def test_lambda_sqs_poller_does_not_tail_match_foreign_region_event_source(monke
         set_request_region(original_region)
 
 
+def _kinesis_stream_record(stream_name: str, stream_arn: str) -> dict:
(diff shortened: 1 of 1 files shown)


Nahuel990:
Thanks!

PR #813, opened 2026-06-04
fix(apigatewayv2): resolve JWKS via OIDC discovery for non-Cognito issuers (#802)

### LINKED ISSUE #802: Bug: API Gateway v2 JWT authorizer hardcodes JWKS path to /.well-known/jwks.json and it breaks non-Cognito OIDC issuers
### Affected Service

API Gateway (v2)

### What happened?

Hit this while integrating Salesforce OAuth. Our APIs failed because the JWT authorizer fetched signing keys from `{issuer}/.well-known/jwks.json` instead of Salesforce's actual `jwks_uri` (`https://login.salesforce.com/id/keys`), so no token ever validated and every request to a JWT-protected route came back 401.

`_resolve_jwks_url()` in `ministack/services/apigateway.py` assumes every JWT issuer serves its keys at `{issuer}/.well-known/jwks.json`:

​```python
def _resolve_jwks_url(authorizer: dict) -> str | None:
    ...
    if issuer.startswith("https://cognito-idp.") and ".amazonaws.com/" in issuer:
        pool_id = issuer.rsplit("/", 1)[-1]
        return f"http://{_HOST}:{_PORT}/{pool_id}/.well-known/jwks.json"
    return f"{issuer}/.well-known/jwks.json"   # <-- hardcoded path
​```

This only holds for Cognito (whose published jwks_uri happens to live at that conventional path). Other OIDC providers publish their keys elsewhere, e.g. Salesforce at `/id/keys`, Okta at `/oauth2/v1/keys` ,so validation fails.

We currently work around it by running a script to patch the JWKS URL every time MiniStack starts, which isn't sustainable.

### Reproduction steps

```python
1. Start MiniStack.
2. Create an HTTP API (v2) with a JWT authorizer whose issuer is a non-Cognito
   OIDC provider:
     - Issuer:  https://login.salesforce.com   ← confirm your exact issuer
     - Audience: <your client id>
3. Attach the authorizer to a route (AuthorizationType=JWT).
4. Call the route with a valid Salesforce-issued JWT in the Authorization header.
5. MiniStack fetches keys from `https://login.salesforce.com/.well-known/jwks.json` (404 / wrong doc) instead of the real `https://login.salesforce.com/id/keys`, the signing key is never found, and the request is rejected 401.
```

### Expected behavior

Validation should succeed for any standards-compliant OIDC issuer, matching real AWS. AWS API Gateway HTTP APIs don't guess the path, they perform OpenID Connect discovery: GET `{issuer}/.well-known/openid-configuration`, read the `jwks_uri` field, then fetch keys from that URL. MiniStack should do the same (keeping the Cognito local-rewrite as a fast path). No JWKS-path config knob is needed, AWS exposes only `issuer` and `audience`.

### MiniStack version

v1.3.56

### SDK / tool used

Terraform aws_apigatewayv2_authorizer, AWS CLI

Pratham2703005:
Fixes #802. The JWT authorizer hardcoded the JWKS path to {issuer}/.well-known/jwks.json, which 404s for non-Cognito issuers like Salesforce (keys live at /id/keys). Now it reads jwks_uri from the issuer's OIDC discovery doc (matching AWS), falling back to the old path if discovery is unavailable. Cognito behavior unchanged.

Added tests covering discovery, fallback, and the Cognito path.


ministack/services/apigateway.py
```diff
@@ -87,6 +87,7 @@ async def _urlopen_async(request_or_url, timeout_seconds: float):
 REGION = os.environ.get("MINISTACK_REGION", "us-east-1")
 _PROXY_TIMEOUT_SECONDS = _timeout_from_env("MINISTACK_APIGW_PROXY_TIMEOUT_SECONDS", 30.0)
 _JWKS_TIMEOUT_SECONDS = _timeout_from_env("MINISTACK_APIGW_JWKS_TIMEOUT_SECONDS", 5.0)
+_OIDC_TIMEOUT_SECONDS = _timeout_from_env("MINISTACK_APIGW_OIDC_TIMEOUT_SECONDS", 5.0)
```
Nahuel990, line 90: I dont follow why do we need this new variable, can you explain?

ministack/services/apigateway.py
```diff
@@ -87,6 +87,8 @@ async def _urlopen_async(request_or_url, timeout_seconds: float):
...
+_OIDC_CACHE_TTL_SECONDS = _timeout_from_env("MINISTACK_APIGW_OIDC_CACHE_TTL_SECONDS", 7200.0)
```
Nahuel990, line 91: And this one too

Nahuel990:
@Pratham2703005 I will merge it becuase the skeleton is correct but I willl remove those extra variables which are not needed.

Nahuel990 (approved):

PR #794, opened 2026-06-03
feat(ec2,eks): Fleet capacity + OIDC identity provider config

b-rajesh:
## Summary

   Implements **EC2 Fleet capacity allocation** (CreateFleet, DescribeFleets) and **EKS OIDC identity provider config** (Associate, Describe, Disassociate) — corresponds to Tier 1 + Tier 2 of the uplift plan.

   The PR is structured as **two `feat` commits + two `fix` commits**. The fixes are a self-review pass on the feat commits that addresses abstraction leaks, AWS parity gaps, and serialization hygiene — kept separate so the review history is readable.

   ### EC2 — `services/ec2.py`
   - `CreateFleet` parses `TargetCapacitySpecification`, `LaunchTemplateConfigs` (with `Overrides`), and `TagSpecifications`. Launches instances via a shared `_launch_instances_internal` helper extracted from `_run_instances`.
   - `DescribeFleets` returns every fleet type regardless of whether the caller passes `FleetIds` (matches real AWS — earlier draft incorrectly filtered out `instant` fleets).
   - `_fleets` registered in `get_state` / `restore_state` / `reset` for persistence + test isolation.
   - All XML interpolations in fleet response builders run through `_esc()`; numeric capacity fields cast to `int` to emit AWS-shaped integers instead of Python floats.
   - Load-bearing comments preserved through the `_run_instances` → `_launch_instances_internal` extraction (BDM/Cloud Custodian rationale, IAM profile synthesis rationale).

   ### EKS — `services/eks.py`
   - `AssociateIdentityProviderConfig` / `DescribeIdentityProviderConfig` / `DisassociateIdentityProviderConfig` implemented as REST endpoints under `/clusters/{name}/identity-provider-configs/{verb}`.
   - OIDC issuer URL, client ID, and optional username/groups claims are forwarded to the k3s API server via `--kube-apiserver-arg=oidc-*` flags on cluster restart.
   - `identityProviderConfigArn` generated at associate time and stored in the cfg record. Describe returns the stored ARN — stable across calls so IaC tools (Terraform / CDK / Pulumi) don't see false-positive drift on every poll.
   - New `_collect_oidc_state(cluster_name)` helper resolves request-scoped state (OIDC args + live cfg refs) at the request boundary. The k3s background thread closes over the pre-resolved values and never re-enters `AccountScopedDict`
    or `get_account_id()` — necessary because Python `contextvars` do not propagate to threads.
   - `_idp_configs` registered in `get_state` / `restore_state` / `reset`.
[...]

PR #814, opened 2026-06-04
feat(eks): default topology node labels on k3s cluster creation

b-rajesh:
## Summary
  
  Adds default Kubernetes topology labels to every EKS k3s node so topology-aware controllers (Karpenter, Cluster Autoscaler, scheduler `topologySpreadConstraints`) schedule pods without a manual `kubectl label node` workaround.

  - `--node-label topology.kubernetes.io/zone={region}a`
  - `--node-label topology.kubernetes.io/region={region}`

  Region resolves through `get_region()` 

  ### Configuration surface

  - **Per cluster** — `CreateCluster` tag key `ministack.node-label/<labelKey>` adds or overrides any node label (e.g. `ministack.node-label/topology.kubernetes.io/zone=us-west-2c`).
  - **Org-wide** — `MINISTACK_EKS_DEFAULT_NODE_LABELS="k1=v1,k2=v2"` env var applied to every cluster on creation. Per-cluster tag overrides win on key collision.

  ### Design

  - New `_collect_node_labels(cluster)` helper runs in the request context, uses the public `AccountScopedDict` API only, and returns a pre-resolved list of `--node-label=k=v` arg strings.
  - `_k3s_run_kwargs(... node_labels=None)` takes the pre-resolved list via closure capture, matching the existing `oidc_args` pattern from PR #794 / #804. The background `_bg_start` / `_bg_restart` threads never touch `AccountScopedDict` 
  or `get_region()` directly — Python contextvars don't propagate to threads, so resolving at the request boundary is necessary.
  - Both `_create_cluster` (initial cluster create) and `_restart_k3s` (OIDC associate/disassociate restart path) pass labels through, so a cluster's labels survive an IdP-triggered container restart.

  ### Karpenter context

  Eliminates one of three workarounds 
- (`DescribeCluster` endpoint reachability) 

## Test plan

  - [x] `ruff check ministack/services/eks.py tests/test_eks.py` — clean
  - [x] `pytest tests/test_eks.py -v -n 4 --dist=loadfile -m "not serial"` — 36 passed (includes 5 new tests below)
  - [x] `pytest tests/test_eks.py -v -m "serial"` — 2 passed (`test_eks_create_describe_delete_cluster`, `test_eks_cfn_cluster`)
  - [x] `test_eks_collect_node_labels_emits_aws_topology_defaults` — region/zone always present
  - [x] `test_eks_collect_node_labels_honours_tag_overrides` — `ministack.node-label/<key>` tags override defaults; non-prefixed tags ignored
  - [x] `test_eks_parse_default_node_labels_env_var` — env-var parser tolerates whitespace, drops malformed entries
  - [x] `test_eks_collect_node_labels_env_var_layered_under_tags` — tag overrides env-default for same key; env-default kept when no tag; AWS defaults still emitted
  - [x] `test_eks_k3s_run_kwargs_appends_node_labels` — labels flow into the k3s server command verbatim; existing server flags preserved


Nahuel990:
@b-rajesh I am merging this because the skeleton is correct, but I am removing the manual setup for tags.

Per AWS, any tag for any node should be done via the proper AWS shape: `CreateNodegroup.labels` or launch-template kubelet args, not a MiniStack-specific `ministack.node-label/` tag convention on the cluster.

Same reason I'm dropping `MINISTACK_EKS_DEFAULT_NODE_LABELS`: AWS has no org-wide node-label config knob, and the surface duplicates what the proper Nodegroup field will cover.

What I'm keeping are the AWS-default topology labels:

* `topology.kubernetes.io/zone={region}a`
* `topology.kubernetes.io/region={region}`

Those match real EKS, the cloud-controller-manager emits them on every node, and they're what unblock Karpenter and `topologySpreadConstraints`.

If we want per-node-group labels later, the right move is to wire them into `CreateNodegroup` via the existing AWS labels field. That keeps the API surface AWS-shape-clean.

Thanks for the work. The labels flow-through IdP restart catch and the contextvar-resolved-at-request-boundary pattern are both correct.

Nahuel990 (approved):

PR #815, opened 2026-06-04
fix(cognito): derive issuer host from environment variables

neriyaco:
## Problem

The JWT generator hardcodes the `iss` (issuer) claim to `amazonaws.com`. When local backends validate tokens against a local endpoint (e.g., aws.local), verification fails with an Invalid issuer error because the token claim doesn't match the local authority URI.

## Solution

Dynamically determine the issuer host using a flexible fallback hierarchy:

1. `MINISTACK_COGNITO_ISSUER_BASE_URL`: Explicit override (e.g., `http://aws.local:4566`).
2. `MINISTACK_HOST`: Smart default. Automatically prepends `http://` if a URI scheme is missing to ensure spec compliance.
3. AWS Default: Fallback to `https://cognito-idp.{region}.amazonaws.com` if neither variable is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -710,6 +710,7 @@ end-to-end without any client config.
 | `MINISTACK_SSL_CERT` | _(unset)_ | Optional PEM-encoded server certificate path; required together with `MINISTACK_SSL_KEY`. When unset, MiniStack auto-generates a self-signed cert under `${TMPDIR}/ministack-tls/` (cached across restarts) |
 | `MINISTACK_SSL_KEY` | _(unset)_ | Optional PEM-encoded private key path; required together with `MINISTACK_SSL_CERT` |
 | `MINISTACK_IMDS_V2_REQUIRED` | `0` | Reject token-less GETs on `/latest/meta-data/...`. When set, callers must first `PUT /latest/api/token` and pass the token as `X-aws-ec2-metadata-token`, matching real-AWS hop-limit-1 IMDSv2-only instances |
+| `MINISTACK_COGNITO_ISSUER_BASE_URL` | _(unset)_ | Override the base URL used in Cognito User Pool JWT tokens' `iss` claim and JWKS fetches. If not set and the `MINISTACK_HOST` is set, the base URL is derived from the host and pool ID, otherwise defaults to `https://cognito-idp.{region}.amazonaws.com/{pool_id}` |
 
 ### API Gateway HTTP proxy execution model
 
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -631,9 +631,8 @@ async def _handle_cognito_get_request(method: str, path: str, headers: dict, que
             if pool_id:
                 cognito = _get_module("cognito")
                 if cognito._get_pool_unscoped(pool_id) is not None:
-                    region = extract_region(headers) or "us-east-1"
                     host = headers.get("host") or headers.get("Host")
-                    return cognito.well_known_openid_configuration(pool_id, region, host)
+                    return cognito.well_known_openid_configuration(pool_id, host)
 
     if path == "/oauth2/authorize" and method == "GET":
         return _get_module("cognito").handle_oauth2_authorize(method, path, headers, query_params)
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -173,7 +173,7 @@ def well_known_openid_configuration(pool_id: str, host: str | None = None):
     # from the pool's region (encoded in pool_id), so do the same here. The
     # `region` argument from the request scope is only used as a last-resort
     # fallback for malformed pool_ids.
-    issuer = _get_issuer_base_url(pool_id)
+    issuer = _get_issuer_url(pool_id)
     base = f"http://{host}" if host else f"http://{_MINISTACK_HOST}:{_MINISTACK_PORT}"
     pool_base = f"{base}/{pool_id}"
     doc = {
@@ -369,8 +369,8 @@ def _pool_region(pool_id: str) -> str:
     return get_region()
 
 
-def _get_issuer_base_url(pool_id: str) -> str:
-    """Return the issuer host for a pool, derived from environment variables or the pool's region."""
+def _get_issuer_url(pool_id: str) -> str:
+    """Return the issuer host for a pool, derived from environment variables and pool_id."""
     if issuer_env := os.getenv("MINISTACK_COGNITO_ISSUER_BASE_URL"):
         return f"{issuer_env}/{pool_id}"
 
@@ -416,7 +416,7 @@ def _fake_token(sub: str, pool_id: str, client_id: str, token_type: str = "acces
     origin_jti = new_uuid()
     claims = {
         "sub": sub,
-        "iss": _get_issuer_base_url(pool_id),
+        "iss": _get_issuer_url(pool_id),
         "token_use": token_type,
         "iat": now,
         "exp": now + 3600,
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -159,7 +159,7 @@ def well_known_jwks(pool_id: str):
     return 200, {"Content-Type": "application/json"}, json.dumps({"keys": [_JWKS_KEY]}).encode()
 
 
-def well_known_openid_configuration(pool_id: str, region: str | None = None, host: str | None = None):
+def well_known_openid_configuration(pool_id: str, host: str | None = None):
     """Return OpenID Connect discovery document.
 
     `issuer` matches the JWT `iss` claim (real AWS URL) so OIDC clients that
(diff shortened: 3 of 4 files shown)


Nahuel990:
@neriyaco we try to keep maximum parity with AWS, and what you're proposing is a shortcut. Real AWS always issues `iss=https://cognito-idp.{region}.amazonaws.com/{pool_id}`, andMiniStack already emits exactly that

Could you describe your setup, which validator, and what you're actually trying to do end-to-end? Depending on the answer there's probably a feature we already support that fits better than overriding iss. For example:

* CUSTOM_AUTH flow (`DefineAuthChallenge` / `CreateAuthChallenge` / `VerifyAuthChallengeResponse` Lambda triggers, recently landed) lets you delegate authentication to your own logic or your own auth server, while Cognito stays as the token issuer (same as AWS)

* OIDC identity providers (`CreateIdentityProvider` with `Type=OIDC`) lets a User Pool federate to an external OIDC server you control, with that server's iss/JWKS configured on the IdP record.

If none of those fit and you really need the validator to see ministack as the AWS endpoint, the standard fallback is an `/etc/hosts` entry mapping `cognito-idp.{region}.amazonaws.com` to your Ministack host, so iss/JWKS lookups land on Ministack while everything stays AWS-shaped.

Once we understand the setup we can point you at the right path without Ministack having to issue non-AWS-shaped tokens

neriyaco:
Ok, after thinking about your comment I think the correct way to do what I want is to fetch the OIDC config from the well-known endpoint, then compare the issuer claim from there

I used `jwt.decode(issuer=COGNITO_ENDPOINT)` but it really doesn't make sense

I'll close this PR

pny-nc:
Hi @Nahuel990,
I'm working on a vault cognito integration with ministack. Unfortunately when I try to register ministack cognito into vault, it's checking the discovery urls against the issuer, and it fails:
```[ERROR] auth.oidc.auth_oidc_3c4ec7e6: error checking oidc discovery URL: error="error creating provider with given values: NewProvider: unable to create provider: oidc: issuer URL provided to client (\"http://cognito-idp.localhost.localstack.cloud:4566/us-east-1_lAJUTsltZ\") did not match the issuer URL returned by provider (\"https://cognito-idp.us-east-1.amazonaws.com/us-east-1_lAJUTsltZ\")```
Could you consider to set or configure issuer to the ministack issuer, please?

**Comment by neriyaco** [comment]
@pny-nc You can construct the issuer URL like that:
```py
region = user_pool_id.split("_")[0]
f"https://cognito-idp.{region}.amazonaws.com/{user_pool_id}"
```
It doesn't make sense to change the issuer URL since AWS will never give you something other than this format, so your code shouldn't support this behavior

pny-nc:
> @pny-nc You can construct the issuer URL like that:
> 
> ```python
> region = user_pool_id.split("_")[0]
> f"https://cognito-idp.{region}.amazonaws.com/{user_pool_id}"
> ```
> 
> It doesn't make sense to change the issuer URL since AWS will never give you something other than this format, so your code shouldn't really support this behavior

I'm using terraform for this with the following snippet:
```
resource "vault_jwt_auth_backend" "cognito" {
  path = var.vault_oidc_path
  type = "oidc"

  oidc_discovery_url = "http://cognito-idp.localhost.localstack.cloud:4566/${aws_cognito_user_pool.vault.id}"
```


I've edited `cognito.py` / `well_known_openid_configuration` with the following:
```"
"issuer": pool_base,
```
And the registration finished without errors

Edit: I've tried to set `bound_issuer` but it did not helped

neriyaco:
If you set `bound_issuer` to `"https://cognito-idp.us-east-1.amazonaws.com/${aws_cognito_user_pool.vault.id}"` and it didn't work, you should open an issue in [terraform](https://github.com/hashicorp/terraform-provider-vault/issues).

pny-nc:
I've found a relevant documentation in [openid](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation):
> The issuer value returned MUST be identical to the Issuer URL that was used as the prefix to /.well-known/openid-configuration to retrieve the configuration information.

For my understanding is, if I use `http://cognito-idp.localhost.localstack.cloud:4566` as discovery url, the issuer should be the same, otherwise the configuration is not valid.
Do you have any workaround to use aws style discovery url? For example define cognito endpoint url or something like similar than any other endpoint?
My bet is vault using default go aws sdk

Nahuel990:
Hi @pny-nc, short answer: keep using the AWS-style discovery URL. We already issue the real AWS issuer https://cognito-idp.us-east-1.amazonaws.com/{pool} and returns the same value as the discovery issuer, exactly like AWS. We won't change that, so the fix is on the Vault side as @neriyaco mentioned 

go-oidc requires the discovery doc's issuer to be byte-identical to the URL you fetched it from. You point Vault at `http://cognito-idp.localhost.localstack.cloud:4566/...`, the issuer is `https://cognito-idp.us-east-1.amazonaws.com/...`, so it rejects at provider creation; that's your exact error. `bound_issuer` can't change that (the check happens before it), and the `"issuer": pool_base` edit only gets you past registration. Login still fails because the token's `iss` is doesnt change

The supported, full-parity fix is to give Vault the real AWS URL and route that hostname to MiniStack:

1. Use the AWS discovery URL:

    resource "vault_jwt_auth_backend" "cognito" {
      path                  = var.vault_oidc_path
      type                  = "oidc"
      oidc_discovery_url    = "https://cognito-idp.us-east-1.amazonaws.com/${aws_cognito_user_pool.vault.id}"
      oidc_discovery_ca_pem = file("${path.module}/ministack-ca.pem")
    }

2. Resolve `cognito-idp.us-east-1.amazonaws.com` to MiniStack from inside the Vault container (Vault does the fetch, not Terraform):

    vault:
      extra_hosts:
        - "cognito-idp.us-east-1.amazonaws.com:<ministack-host-ip>"

3. Front MiniStack on that hostname on both ports, and preserve the Host header:

   - 443 with TLS for discovery (cert SAN = `cognito-idp.us-east-1.amazonaws.com`, trusted via `oidc_discovery_ca_pem`)
   - 80 for the JWKS fetch (the doc serves `jwks_uri` over HTTP; go-oidc fetches it as-is)

Result: `oidc_discovery_url` == discovery issuer == token `iss`, the keys should resolve out of the box, and tokens stay byte-for-byte AWS-shaped. Drop the `extra_hosts`/CA lines and the same config runs against real AWS. That's deliberate. We keep Cognito at full AWS parity (like all the services) instead of issuing a non-AWS issuer

pny-nc:
Hi @Nahuel990,

I did your steps, but in this case, I needed to create an additional `hosts` file entry for the `cognito-idp.us-east-1.amazonaws.com` address to 127.0.0.1 on the host, that i'm using vault, because it's opens a new window for login, so i'd like to say, it's rather unusable.
While i've checked localstack, they're issuing a configuration, which is much closer to my exceptations (discovery url: `http://localhost.localstack.cloud:4566/{user_pool_id}`):
```
{
  "authorization_endpoint": "http://localhost.localstack.cloud:4566/_aws/cognito-idp/oauth2/authorize",
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  "issuer": "http://localhost.localstack.cloud:4566/us-east-1_2c5bf2ab4c9e4b57904f01a191a1d817",
  "jwks_uri": "http://localhost.localstack.cloud:4566/us-east-1_2c5bf2ab4c9e4b57904f01a191a1d817/.well-known/jwks.json",
  "response_types_supported": [
    "code",
    "token",
    "token id_token"
  ],
  "scopes_supported": [
    "openid",
    "email",
    "phone",
    "profile"
  ],
  "subject_types_supported": [
    "public"
  ],
  "token_endpoint": "http://localhost.localstack.cloud:4566/_aws/cognito-idp/oauth2/token",
  "token_endpoint_auth_methods_supported": [
    "client_secret_basic",
    "client_secret_post"
  ],
  "userinfo_endpoint": "http://localhost.localstack.cloud:4566/_aws/cognito-idp/oauth2/userInfo",
  "end_session_endpoint": "http://localhost.localstack.cloud:4566/_aws/cognito-idp/logout"
}
```
Still not get it, why don't you want to create an issuer URL based on discovery URL, which makes everything much easier. So I still kindly ask you to consider to generate issuer (and the whole workflow) from the discovery url / openid-configuration, same as all the endpoints, and in this case it will be more useable, and uniform.

On the other hand, something is missing from the cognito integration, but I'll open a new issue about my findings, it's unrelated to this.

Nahuel990:
@pny-nc, please open a fresh issue for this instead of commenting on a closed older PR as they might get lost and open issues stays visible and easier to track


`localhost.localstack.cloud` is a public wildcard DNS record that resolves to `127.0.0.1`. Every time your OIDC client validates a token, it performs a real public DNS lookup against a third party just to reach your own machine (you can check it with nslookup or dig + url) That introduces an external dependency into your local auth path, it breaks when offline, and it exposes those lookups to your resolver. A public hostname pointing to loopback can also allow any web page you open to reach services bound to your local `127.0.0.1`. 

I've used LS for years and I know it ships many convenience options but they drift over time: something works in Ls but not in real AWS, or the other way around, and you only find out in production. We aim to maintain full parity with AWS. I’m happy to add configuration options that make migration to MiniStack easier, as long as they do not change AWS behavior. A host-relative issuer does change AWS behavior, because it allows Ministack to mint a token AWS would never issue and that is exactly the kind of drift we will not ship.

The parity-preserving approach is the one described above: point your client at the real AWS issuer URL and route that hostname to MiniStack. If you open a new issue with your `vault_jwt_auth_backend` configuration and discovery URL, I’ll help you set it up cleanly, and we can track your other Cognito findings there as well

PR #820, opened 2026-06-04
Parse Lambda event source mapping tag ARNs by request scope

Areson:
## Summary
- parse Lambda event-source-mapping tag ARNs with the shared ARN parser
- require the ARN to match the current account and Region before resolving the ESM UUID
- add coverage that wrong-scope ESM ARNs do not fall back to local mappings

## Tests
- `uv run --extra dev pytest tests/test_lambda.py -q` (`234 passed, 1 skipped`)


Nahuel990:
Thanks for this I found an issue we currrently have in main.

PR #822, opened 2026-06-04
Scope S3 Tables resource ARNs by region

Areson:
## Summary
- parse S3 Tables bucket ARNs before bucket, namespace, and table operations
- move S3 Tables state to account+Region scoped stores so same-name buckets can exist independently by Region
- preserve Iceberg REST catalog behavior across unsigned and signed Region-scoped requests

## Tests
- `uv run --extra dev pytest tests/test_s3tables.py -q` (`10 passed`)

### DIFF
--- a/ministack/services/s3tables.py
+++ b/ministack/services/s3tables.py
@@ -31,12 +31,14 @@ import copy
 import json
 import logging
 import os
+import re
 import time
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

PR #825, opened 2026-06-05
fix(appsync): send standard AppSync resolver event to Lambda data sources

### LINKED ISSUE #781: AppSync: pass the standard AWS resolver event (and run the Lambda authorizer) to Lambda data sources
## Summary

When AppSync runs a GraphQL operation backed by a Lambda data source, ministack invokes the Lambda with only the parsed GraphQL arguments. Real AWS AppSync sends the full resolver event. Any function that reads `event.info.fieldName`, `event.identity`, or `event.request.headers` therefore fails against ministack. Separately, for APIs configured with `AWS_LAMBDA` authentication, ministack never invokes the Lambda authorizer, so `event.identity` is never populated.

## Current behavior

`_resolve_lambda(data_source, args)` in `ministack/services/appsync.py` passes `args` as the entire Lambda event. There is no `info`, no `identity`, no `request.headers`, and the Lambda authorizer is never called.

## Desired behavior (AWS parity)

1. For a Lambda data source, build and pass the standard AppSync resolver event.
2. When the API uses `AWS_LAMBDA` auth, invoke the configured Lambda authorizer once per request and attach its `resolverContext` as `event.identity.resolverContext`.
3. Forward the incoming HTTP request headers (lowercase keys) as `event.request.headers`.

### Event shape (standard AWS, no custom flattening)

```json
{
  "arguments": { "...": "parsed GraphQL args" },
  "identity": { "resolverContext": { "...": "from the Lambda authorizer" } },
  "source": {},
  "request": { "headers": { "x-api-key": "...", "authorization": "..." } },
  "prev": null,
  "stash": {},
  "info": {
    "fieldName": "getThing",
    "parentTypeName": "Query",
    "variables": {}
  }
}
```

Notes:

* `fieldName` lives under `info.fieldName`. There is no top level `fieldName` (this matches AWS; a top level field name only appears when a request mapping template produces it).
* `identity` is omitted (or `null`) for `API_KEY` auth. Consumers detect API key auth by looking for `x-api-key` in `request.headers`.
* `identity.resolverContext` is whatever the authorizer returns. ministack should pass it through and assume no specific fields.
* `info.selectionSetList` and `info.selectionSetGraphQL` may be stubbed.

### Lambda authorizer (AWS_LAMBDA auth mode)

Invoke the authorizer once per request. The ARN comes from `lambdaAuthorizerConfig.authorizerUri` (the AWS field name). Send the standard AppSync authorizer event:

```json
{ "authorizationToken": "...", "requestContext": {}, "requestHeaders": {} }
```

Expected response:

```json
{ "isAuthorized": true, "resolverContext": { "...": "arbitrary user context" } }
```

If the authorizer returns `isAuthorized: false`, is not found, or fails to invoke, proceed with `identity = null` and do not return HTTP 500 (graceful degradation).

## Reproduction

```python
import io, json, urllib.request, zipfile

import boto3

def zip_handler(src: bytes) -> bytes:
    buf = io.BytesIO()
    with zipfile.ZipFile(buf, "w") as zf:
        zf.writestr("index.py", src.decode())
    return buf.getvalue()

resolver = b"""
import json
def handler(event, ctx):
    # A real AppSync Lambda resolver receives the field name at event.info.fieldName.
    field = (event.get('info') or {}).get('fieldName')
    return {'body': json.dumps({'seenFieldName': 'fn:' + str(field)})}
"""

lam = boto3.client('lambda', endpoint_url='http://localhost:4566', region_name='us-east-1',
                   aws_access_key_id='test', aws_secret_access_key='test')
lam.create_function(FunctionName='resolver', Runtime='python3.12',
                    Role='arn:aws:iam::000000000000:role/test-role',
                    Handler='index.handler', Code={'ZipFile': zip_handler(resolver)})
fn_arn = lam.get_function(FunctionName='resolver')['Configuration']['FunctionArn']

c = boto3.client('appsync', endpoint_url='http://localhost:4566', region_name='us-east-1',
                 aws_access_key_id='test', aws_secret_access_key='test')
api = c.create_graphql_api(name='t', authenticationType='API_KEY')['graphqlApi']
api_id = api['apiId']
api_key = c.create_api_key(apiId=api_id)['apiKey']['id']
c.create_data_source(apiId=api_id, name='LambdaDS', type='AWS_LAMBDA',
                     lambdaConfig={'lambdaFunctionArn': fn_arn})
c.create_resolver(apiId=api_id, typeName='Query', fieldName='testField',
                  dataSourceName='LambdaDS', kind='UNIT')

req = urllib.request.Request(
    'http://localhost:4566/graphql',
    data=json.dumps({'query': '{ testField { fieldName } }'}).encode(),
    headers={'Content-Type': 'application/json', 'x-api-key': api_key}, method='POST')
resp = json.loads(urllib.request.urlopen(req).read())
# Today the Lambda receives only `args` (no `info`), so the echoed value is "fn:None".
# After the fix the Lambda must receive event.info.fieldName == "testField".
assert 'fn:testField' in json.dumps(resp), resp   # currently fails
```

## Acceptance criteria

* `_resolve_lambda` builds the standard event with `info.fieldName` (no top level `fieldName`), plus `arguments`, `source`, `request.headers`, `prev`, `stash`, `info`.
* `request.headers` is forwarded with lowercase keys.
* `AWS_LAMBDA` auth invokes the authorizer at `lambdaAuthorizerConfig.authorizerUri` and sets `identity.resolverContext` from its response.
* `API_KEY` auth omits `identity`.
* A missing or failing Lambda function or authorizer does not crash the request (graceful fallback).
* Tests cover: `info.fieldName`, arguments, header forwarding, identity present in `AWS_LAMBDA` mode, identity absent in `API_KEY` mode, missing Lambda, and authorizer rejection.

AdigaAkhil:
## Problem
For `AWS_LAMBDA` data sources, ministack invoked the Lambda with only the raw GraphQL
arguments. Real AppSync sends a full `AppSyncResolverEvent` and runs the Lambda
authorizer first, so code reading `event.info.fieldName`, `event.request.headers`,
`event.identity.resolverContext`, or `event.source` broke under ministack.

## Fix (`ministack/services/appsync.py`)
- Thread incoming HTTP headers through both GraphQL routes into `_execute_graphql`.
- Build the AWS-standard event in `_resolve_lambda`: `arguments`, `source`,
  `request.headers`, `prev`, `stash`, `info` (`fieldName`/`parentTypeName`/`variables`).
- In `AWS_LAMBDA` auth mode, invoke the Lambda authorizer and pass its `resolverContext`
  as `identity.resolverContext` (omitted under `API_KEY`). Rejection (`isAuthorized:false`),
  a missing authorizer, or a failing authorizer all degrade gracefully to `identity=null`
  (HTTP 200, no crash).
- Surface an unhandled Lambda exception as a GraphQL `errors` entry instead of leaking the
  RIE error payload as `data`.

## Tests
New `tests/test_appsync_lambda_event.py` — 14 tests (registered serial in `conftest.py`)
covering `info.fieldName`, arguments, header forwarding, identity from the authorizer,
absent identity under API_KEY, empty `source`, variable substitution, missing-function
fallback, error pass-through, unhandled-exception handling, and the AWS_LAMBDA authorizer
degradation paths (rejection `isAuthorized:false`, missing authorizer, failing/raising
authorizer).

```
make stop && make run
pytest tests/test_appsync_lambda_event.py -v                  # 14 passed
pytest tests/test_appsync.py tests/test_appsync_events.py -q  # 41 passed, 1 skipped (no regressions)
```

## Out of scope
The downstream `graphql-api` `lambdaResolver` needs a `?? event.info.fieldName` fallback
(top-level `event.fieldName` comes from its AWS mapping template, which ministack doesn't
run). Tracked separately — not in this PR.

Closes #781.

### DIFF
--- a/ministack/services/appsync.py
+++ b/ministack/services/appsync.py
@@ -508,14 +508,14 @@ async def handle_request(method, path, headers, body, query_params):
         if not api_id:
             return error_response_json("UnauthorizedException", "Valid API key required", 401)
         data = json.loads(body) if body else {}
-        return _execute_graphql(api_id, data)
+        return _execute_graphql(api_id, data, request_headers=headers)
 
     if path.startswith("/v1/apis/") and path.endswith("/graphql") and method == "POST":
         parts = path.split("/")
         if len(parts) >= 5:
             api_id = parts[3]
             data = json.loads(body) if body else {}
-            return _execute_graphql(api_id, data)
+            return _execute_graphql(api_id, data, request_headers=headers)
 
     m = _PATH_RE.match(path)
     if not m:
@@ -677,7 +677,61 @@ def _resolve_api_by_key(api_key_value):
     return None
 
 
-def _execute_graphql(api_id, data):
+def _invoke_lambda_authorizer(api_id, authorizer_config, request_headers):
+    """Invoke the Lambda authorizer; return an identity dict or None on failure/rejection.
+
+    AWS stores the authorizer Lambda under ``lambdaAuthorizerConfig.authorizerUri``
+    (which ``_create_graphql_api`` persists verbatim). ministack's AppSync Events
+    authorizer reads the same key.
+    """
+    func_arn = authorizer_config.get("authorizerUri") or authorizer_config.get("authorizer_uri")
+    if not func_arn:
+        return None
+
+    import ministack.services.lambda_svc as _lambda_svc
+
+    func_name = func_arn.rsplit(":", 1)[-1]
+    func = _lambda_svc._functions.get(func_name)
+    if not func:
+        logger.warning("Lambda authorizer %s not found in ministack — skipping auth", func_name)
+        return None
+
+    authorizer_event = {
+        "authorizationToken": request_headers.get("authorization", ""),
+        "requestContext": {"operationName": "unknown"},
+        "requestHeaders": request_headers,
+    }
+
+    try:
+        result = _lambda_svc._execute_function(func, authorizer_event)
+    except Exception as e:
+        logger.warning("Lambda authorizer invocation failed: %s", e)
+        return None
+
+    if not isinstance(result, dict) or result.get("error"):
+        logger.warning("Lambda authorizer execution error")
+        return None
+
+    body = result.get("body")
+    if isinstance(body, str):
+        try:
+            body = json.loads(body)
+        except Exception:
+            return None
+    if not isinstance(body, dict):
+        return None
+
+    if not body.get("isAuthorized", False):
+        logger.warning("Lambda authorizer rejected request")
+        return None
+
+    resolver_context = body.get("resolverContext")
+    if resolver_context:
+        return {"resolverContext": resolver_context}
+    return None
+
+
+def _execute_graphql(api_id, data, request_headers=None):
     """Execute a GraphQL query/mutation against the configured resolvers."""
     query = data.get("query", "")
     variables = data.get("variables", {})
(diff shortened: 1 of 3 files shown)


Nahuel990:
@AdigaAkhil I'm merging but changing the UnauthorizedException for failure mode and also moving the test to the corresponding file

Nahuel990 (approved):

PR #826, opened 2026-06-05
fix(s3): account-scope the on-disk bucket directory (#824)

### LINKED ISSUE #824: PutBucketAsync creates folder outside of account ID
### Affected Service

S3

### What happened?

when using the AWS C# SDK calling PutBucketAsync creates a folder in the data directory, without any account ID. Any subsequent call to put data into the bucket causes a directory to be created within the account directory. i.e.

when creating a bucket named 'my-bucket' and putting in a file called 'my-file.txt' causes the following to be created in the S3 data directory

```
s3-data-directory
  -> my-bucket
  -> 000000000000
    -> my-bucket
      -> my-file.txt
```

### Reproduction steps

```python
var client = services.GetRequiredAwsService<Amazon.S3.IAmazonS3>();
await client.PutBucketAsync(new Amazon.S3.Model.PutBucketRequest { BucketName = "my-bucket" });
await client.PutObjectAsync(new new Amazon.S3.Model.PutObjectRequest { BucketName = "my-bucket", Key = "my-file" });
```

### Expected behavior

No folder is created in the root of the S3 data directory

### MiniStack version

1.3.58

### SDK / tool used

AWS SDK for .NET

AdigaAkhil:
Fixes #824.

## Problem

`PutBucket` (CreateBucket) persisted its directory at `DATA_DIR/<bucket>` — **without the account id** — while every object write goes to the account-scoped `DATA_DIR/<account>/<bucket>/<key>` (via `_object_disk_path` / `_persist_object`). The two paths never matched, so creating a bucket left a spurious empty folder at the data-dir root that no code path ever read:

```
s3-data-directory/
├── my-bucket/                  ← spurious, empty (CreateBucket)
└── 000000000000/
    └── my-bucket/
        └── my-file.txt         ← real data (PutObject)
```

It was harmless (the empty-dir skip in `_load_persisted_bucket` keeps it from resurrecting as a phantom bucket on reload), but confusing. Separately, `DeleteBucket` did **no** on-disk cleanup at all, so the bucket directory was orphaned on disk after deletion.

## Fix

- **`_create_bucket`**: account-scope the `makedirs` → `DATA_DIR/<account>/<bucket>`, matching where objects actually land. No more spurious root folder.
- **`_delete_bucket`**: add `_delete_persisted_bucket()` to remove the bucket's account-scoped directory on delete (closes the cleanup gap). The `rmtree` primitive is guarded against escaping `DATA_DIR` or wiping the account dir, and surfaces partial-failure errors via the logged `except` instead of swallowing them.

## Tests

Three new unit tests in `tests/test_s3.py`, each verified to **fail on the pre-fix code and pass after**:

- `test_s3_create_bucket_persists_account_scoped` — persists under `DATA_DIR/<account>/<bucket>`, not the root
- `test_s3_put_object_no_spurious_root_folder` — `PutBucket` + `PutObject` leaves no spurious root folder (mirrors the issue repro)
- `test_s3_delete_bucket_removes_persisted_dir` — `DeleteBucket` removes the on-disk directory

Full `test_s3.py` suite: **140 passed, 1 skipped**.

## Out of scope (noted, not changed here)

While reviewing, two pre-existing issues surfaced that are unrelated to this fix and left for follow-ups:

1. **`glue.py:812` reads the unscoped S3 path** (`os.path.join(S3_DATA_DIR, bucket, key)`) while data is written account-scoped — so Glue never finds on-disk object data (`athena.py` is correctly scoped). This is what makes the pre-existing `test_athena_mixed_glue_and_s3_uri` fail on `main`. Worth a separate issue.
2. **Legacy unscoped layout** (`DATA_DIR/<bucket>` from pre-account-scoping versions) is not cleaned up on delete — consistent with `_delete_persisted_object`, which also only touches the account-scoped path. Documented in the helper docstring; a proper fix is a migration concern.

### DIFF
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -30,6 +30,7 @@ import json
 import logging
 import os
 import re
+import shutil
 import threading
 import time
 from urllib.parse import parse_qs as _parse_qs
@@ -833,7 +834,10 @@ def _create_bucket(name: str, body: bytes, headers: dict = None):
         _bucket_versioning[name] = "Enabled"
 
     if S3_PERSIST:
-        os.makedirs(os.path.join(DATA_DIR, name), exist_ok=True)
+        # Account-scope the on-disk dir to match where objects are actually
+        # written (_object_disk_path / _persist_object). Omitting the account id
+        # here created a spurious empty folder at DATA_DIR/<bucket> (#824).
+        os.makedirs(os.path.join(DATA_DIR, get_account_id(), name), exist_ok=True)
     logger.info("S3 bucket created: %s%s", name, f" (region={region})" if region else "")
     return 200, {"Location": f"/{name}"}, b""
 
@@ -870,6 +874,8 @@ def _delete_bucket(name: str):
         del _object_retention[k]
     for k in [k for k in _object_legal_hold if k[0] == name]:
         del _object_legal_hold[k]
+    if S3_PERSIST:
+        _delete_persisted_bucket(name)
     return 204, {}, b""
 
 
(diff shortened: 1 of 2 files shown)


AdigaAkhil:
@Nahuel990  , could you please have a look at the out of scope items in this PR.
Do you want me to create a separate Issue for the same?

Nahuel990 (approved):

Nahuel990:
Awesome! Thanks @AdigaAkhil

AdigaAkhil:
> @Nahuel990 , could you please have a look at the out of scope items in this PR. Do you want me to create a separate Issue for the same?

@Nahuel990  ^^

Nahuel990:
@AdigaAkhil yes becuase Glue will break... please if you fix those check also Athena, and I'm not sure if the app.py also iterates in the folder... the second is not really an issue, they are orphan, if you add something to clean will remain forever in code... we'll reflect in changelog

AdigaAkhil:
Sure @Nahuel990 , will create a separate Issue for the same.

PR #828, opened 2026-06-05
fix(glue): account-scope on-disk script resolution and crawler completion for non-default accounts (#827)

### LINKED ISSUE #827: Glue StartJobRun cannot resolve on-disk S3 script for non-default accounts (unscoped path + lost thread context)
## Background

Follow-up to #824 / #826. While account-scoping the on-disk S3 layout (`DATA_DIR/<account>/<bucket>/<key>`), two pre-existing issues were noted as out of scope in #826. @Nahuel990 [confirmed](https://github.com/ministackorg/ministack/pull/826#issuecomment-4629922799) the Glue one should be fixed ("yes because Glue will break..."). While investigating, a second, **identical-root-cause** defect surfaced in the Glue crawler. Both are tracked and fixed here.

## Summary

Two related defects in `ministack/services/glue.py`, both because account scoping is **not honoured inside Glue's background threads**. `get_account_id()` reads a `contextvars.ContextVar` (default `000000000000`), and a bare `threading.Thread` / `threading.Timer` does **not** copy the parent context — so background work silently runs under the default account instead of the caller's. This is the same class of bug already fixed for Step Functions in #639 (`contextvars.copy_context()`).

- **Bug 1 — `StartJobRun` cannot resolve on-disk S3 scripts** (unscoped path *and* lost thread context).
- **Bug 2 — Crawlers hang in `RUNNING` forever** for non-default accounts.

---

### Problem
With `S3_PERSIST=1`, a Glue job whose `Command.ScriptLocation` is an `s3://` URI fails to resolve the script from disk. The on-disk fast path is effectively dead for persisted objects, so the job silently stubs to `SUCCEEDED` (or, for a Spark job, never runs the real script).

### Root cause (two compounding parts)
1. **Unscoped on-disk path.** `glue.py:812` (`_resolve_script`) builds:
   ```python
   local_path = os.path.join(S3_DATA_DIR, bucket, key)
   ```
   omitting the account segment. The canonical writer stores objects at `DATA_DIR/<account>/<bucket>/<key>` (`s3.py` `_object_disk_path`, and `_create_bucket` since #826), so `os.path.exists(local_path)` is always `False` for persisted objects. The in-memory fallback (`glue.py:818`) can't recover them either — persisted bodies are dropped from RAM (`s3.py`, `obj["body"] = None`).
2. **Lost request account in the worker thread.** `_resolve_script` runs inside `_execute`, launched on a bare daemon thread (`glue.py:901`, `threading.Thread(target=_execute, daemon=True)`). Because `threading.Thread` doesn't copy `contextvars`, `get_account_id()` inside the thread returns the default `000000000000`, not the caller's account — so even the scoped path (and the `AccountScopedDict` in-memory lookup) miss for any non-default account.

### Fix
- Add `import contextvars` to `glue.py`.
- `glue.py:812` → `os.path.join(S3_DATA_DIR, get_account_id(), bucket, key)` (matches `s3.py` `_object_disk_path` and `athena.py`).
- `glue.py:901` → run the worker through a snapshot of the request context:
  ```python
  ctx = contextvars.copy_context()
  thread = threading.Thread(target=ctx.run, args=(_execute,), daemon=True)
  ```

---

### Problem
`StartCrawler` flips a crawler to `RUNNING` and schedules `_finish_crawl` on a `threading.Timer` to flip it back to `READY`. For any **non-default account**, the crawler never returns to `READY` and `LastCrawl` is never recorded — it stays `RUNNING` indefinitely.

### Root cause
Same contextvar loss as Bug 1, via the timer thread. `_finish_crawl` (`glue.py:677-689`) runs on a `threading.Timer` (`glue.py:691-693`), which (like `threading.Thread`) doesn't copy `contextvars`. `_crawlers` is an `AccountScopedDict` (`glue.py:103`), so the guard at `glue.py:678`:
```python
if name in _crawlers and _crawlers[name]["State"] == "RUNNING":
```
evaluates under the default account `000000000000`, where the crawler doesn't exist → the body never runs. (Works by accident for the default account.)

### Reproduction (empirically verified)
| account | after `StartCrawler` | after timer fires | `LastCrawl` |
|---|---|---|---|
| `000000000000` (default) | `RUNNING` | `READY` | set ✅ |
| `555555555555` (non-default) | `RUNNING` | **`RUNNING`** (stuck) | never set ❌ |

A direct probe confirms the mechanism: a `threading.Timer` started in a thread where the account is `555555555555` runs its target seeing `000000000000`. Applying the `copy_context()` fix makes the non-default crawler finish (`READY`, `LastCrawl` set).

---

## Scope confirmation (full audit)

A sweep of every on-disk S3 path construction confirmed `glue.py:812` is the **only** unscoped on-disk read site:
- `athena.py:398` / `:426` — already account-scoped (and its asyncio chain propagates the contextvar correctly); **no change**.
- `app.py` `reset()` — intentional account-agnostic wipe of all top-level entries; **no change**.
- `s3.py` — canonical scoped writer/reader (`_object_disk_path`, `_persist_object`, `_read_body`, `_create_bucket`, `_delete_persisted_bucket`); **no change**.

The two background-thread context-loss sites (`glue.py:901` job runs, `glue.py:691` crawler timer) are the only places in Glue that spawn account-dependent background work without `copy_context()`.

## Out of scope (per #826 / @Nahuel990)
Cleanup of legacy unscoped on-disk dirs (`DATA_DIR/<bucket>` from pre-account-scoping versions) is **not** addressed — they are harmless orphans and adding cleanup code would have to live forever. To be reflected in the changelog instead.

## Tests
- `_resolve_script` resolves a script persisted under a non-default account's scoped path (fails before, passes after — pins Bug 1's path fix).
- `StartJobRun`'s worker thread resolves the script under the caller's account, not the default (pins Bug 1's thread fix).
- A crawler started under a non-default account returns to `READY` with `LastCrawl` set (pins Bug 2).

AdigaAkhil:
Fixes #827.

## Problem

Two related Glue bugs, both because account scoping isn't honoured in Glue's **background threads**: `get_account_id()` reads a contextvar (default `000000000000`), and a bare `threading.Thread` / `threading.Timer` doesn't copy the parent context — so background work runs under the default account. (Same root cause as #639.)

- **`StartJobRun` can't resolve on-disk S3 scripts** — `_resolve_script` built an unscoped path (`S3_DATA_DIR/<bucket>/<key>`) while objects are persisted at `DATA_DIR/<account>/<bucket>/<key>`, *and* it runs in a daemon thread where the account reverts to the default. Net: file-backed Glue scripts are never found. (The follow-up @Nahuel990 confirmed in #826.)
- **Crawlers hang in `RUNNING` forever** (non-default accounts) — `_finish_crawl` runs on a `threading.Timer`, so the account-scoped `_crawlers` guard misses and the crawler never flips back to `READY`.

## Fix

- Account-scope the path: `os.path.join(S3_DATA_DIR, get_account_id(), bucket, key)`.
- Run the job-run thread and crawler timer through `contextvars.copy_context()` so they carry the request's account — the existing `stepfunctions.py` / `rds.py` idiom.

## Tests

Four in-process tests in `tests/test_glue.py`, each verified to **fail before / pass after**: scoped resolution, worker-thread context, crawler completion, and cross-account isolation. No server / duckdb / docker needed.

## Out of scope (per #826)

Legacy unscoped on-disk dirs (`DATA_DIR/<bucket>`) are harmless orphans — left for the changelog, not cleaned up.

### DIFF
--- a/ministack/services/glue.py
+++ b/ministack/services/glue.py
@@ -808,8 +808,10 @@ def _resolve_script(script_location):
         parts = stripped.split("/", 1)
         bucket = parts[0]
         key = parts[1] if len(parts) > 1 else ""
-        # Check on-disk first
-        local_path = os.path.join(S3_DATA_DIR, bucket, key)
+        # Check on-disk first. Objects are persisted account-scoped at
+        # DATA_DIR/<account>/<bucket>/<key> (see s3._object_disk_path), so the
+        # account id MUST be part of the lookup path or it never matches.
+        local_path = os.path.join(S3_DATA_DIR, get_account_id(), bucket, key)
         if os.path.exists(local_path):
             return local_path
         # Fetch from in-memory S3
--- a/ministack/services/glue.py
+++ b/ministack/services/glue.py
@@ -11,6 +11,7 @@ scripts or when Docker is unavailable.
 Crawlers transition through RUNNING state with a configurable timer.
 """
 
+import contextvars
 import copy
 import fnmatch
 import json
@@ -900,7 +901,12 @@ def _start_job_run(data):
         run["ExecutionTime"] = int(run["CompletedOn"] - run["StartedOn"])
         run["LastModifiedOn"] = int(time.time())
 
-    thread = threading.Thread(target=_execute, daemon=True)
+    # threading.Thread does NOT copy contextvars, so without this snapshot the
+    # worker would run under the default account and fail to resolve the
+    # account-scoped on-disk script (and AccountScopedDict lookups). Carry the
+    # request's account/region into the thread. See issue #639 / stepfunctions.
+    ctx = contextvars.copy_context()
+    thread = threading.Thread(target=ctx.run, args=(_execute,), daemon=True)
     thread.start()
 
     return json_response({"JobRunId": run_id})
--- a/ministack/services/glue.py
+++ b/ministack/services/glue.py
@@ -689,7 +689,12 @@ def _start_crawler(data):
             }
             logger.info("Glue: Crawler %s finished after %ss", name, CRAWLER_RUN_SECONDS)
 
-    timer = threading.Timer(CRAWLER_RUN_SECONDS, _finish_crawl)
+    # threading.Timer (like threading.Thread) does NOT copy contextvars, so
+    # without this snapshot _finish_crawl runs under the default account and the
+    # account-scoped _crawlers guard never matches — the crawler would hang in
+    # RUNNING forever for non-default accounts. See issue #639 / stepfunctions.
+    ctx = contextvars.copy_context()
+    timer = threading.Timer(CRAWLER_RUN_SECONDS, lambda: ctx.run(_finish_crawl))
     timer.daemon = True
     timer.start()
 
--- a/tests/test_glue.py
+++ b/tests/test_glue.py
@@ -1235,3 +1235,38 @@ def test_glue_crawler_completes_for_non_default_account(monkeypatch):
     finally:
         respmod._request_account_id.reset(token)
         gluemod._crawlers._data.pop((account, name), None)
+
+
+def test_glue_resolve_script_isolated_per_account(tmp_path, monkeypatch):
+    """A script persisted under one account is not resolvable from another (#827).
+
+    Guards the core multi-tenancy property: account-scoped resolution must not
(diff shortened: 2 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #836, opened 2026-06-05
fix(cognito): merge CUSTOM_AUTH verify result into the pending challenge round (#835)

### LINKED ISSUE #835: Cognito CUSTOM_AUTH: session[] splits challengeResult and challengeMetadata into two entries per round (should be one merged ChallengeResult)
### Affected Service

Cognito

### What happened?

In the Cognito `CUSTOM_AUTH` flow, ministack appends **two** half-populated entries to the
`session` array for a **single** completed challenge round:

- one entry with `challengeMetadata` (set by `CreateAuthChallenge`) and `challengeResult: null`, and
- a second entry with `challengeResult` (from `VerifyAuthChallengeResponse`) and `challengeMetadata: null`.

AWS Cognito records exactly **one** `ChallengeResult` element per round, carrying **both**
fields together. The split breaks any multi-round custom-auth flow whose `DefineAuthChallenge` /
`CreateAuthChallenge` reads `challengeMetadata` **and** `challengeResult` from the same element to
pick the next step (e.g. magic-link → SMS-OTP, CAPTCHA → security-question).

**Actual `session` after answering round 1** (a `CUSTOM_CHALLENGE` whose `CreateAuthChallenge`
set `challengeMetadata: "ROUND_1"`):

```json
[
  { "challengeName": "CUSTOM_CHALLENGE", "challengeResult": null, "challengeMetadata": "ROUND_1" },
  { "challengeName": "CUSTOM_CHALLENGE", "challengeResult": true, "challengeMetadata": null }
]
```

The round's metadata is stranded on the pending entry, and the result lands on a separate,
metadata-less entry — so `session[-1].challengeResult && session[-1].challengeMetadata` is never
both-truthy on one element, and a metadata-driven step transition never fires.

**Root cause** — `ministack/services/cognito.py`. The InitiateAuth paths build one entry per
round and update it **in place**, but the RespondToAuthChallenge paths **append** a second entry
for the verify result with `challengeMetadata=None`:

| Path | Function | Line (v1.3.59) | Behavior |
|------|----------|----------------|----------|
| InitiateAuth | `_initiate_auth` / `_admin_initiate_auth` | 2685-2692 | updates pending entry in place — correct |
| RespondToAuthChallenge | `_admin_respond_to_auth_challenge` | **2415** | `_append_challenge_to_session(session, "CUSTOM_CHALLENGE", answer_correct, None, {}, {})` — the split |
| RespondToAuthChallenge | `_respond_to_auth_challenge` | **2766** | same line — the split |

**Proposed fix** — at both respond sites, update the pending round's result in place (mirroring
the in-place `.update()` the InitiateAuth paths already use for `challengeMetadata`), falling back
to append only when there is no pending entry:

```python
# AWS records one ChallengeResult per round, carrying BOTH the metadata (set by
# CreateAuthChallenge) and the result (from VerifyAuthChallengeResponse). Update the
# pending round's result in place rather than appending a second, metadata-less entry.
if session["challenges"] and session["challenges"][-1].get("challengeResult") is None:
    session["challenges"][-1]["challengeResult"] = answer_correct
else:
    _append_challenge_to_session(session, "CUSTOM_CHALLENGE", answer_correct, None, {}, {})
```

`answered_count = sum(1 for c in session["challenges"] if c.get("challengeResult") is not None)`
stays correct (one per round), and `issueTokens` / max-attempts logic is unchanged.

**Impact / why it slipped through** — only multi-round flows that use `challengeMetadata` for step
routing are affected. The existing `tests/test_cognito_custom_auth.py` triggers read only
`session[-1].challengeResult` and count non-null results, so they pass under both shapes.

**Related** — orthogonal to #809 (the `asyncio.to_thread` deadlock fix shipped in 1.3.59; this bug
remains after it). #725 added `CUSTOM_AUTH` with result-only test triggers, so this gap wasn't caught.

### Reproduction steps

```python
Self-contained repro against a stock ministack. The `CreateAuthChallenge` trigger echoes the
`session` it receives into `publicChallengeParameters.session_seen`, which Cognito returns to the
client — so the split is directly observable in the response, no log-diving:


import io, json, time, zipfile, boto3
from botocore.config import Config

ms = lambda svc: boto3.client(svc, endpoint_url="http://localhost:4566", region_name="us-east-1",
                              aws_access_key_id="test", aws_secret_access_key="test",
                              config=Config(retries={"max_attempts": 1}))

def zip_handler(src):
    buf = io.BytesIO()
    with zipfile.ZipFile(buf, "w") as z:
        z.writestr("index.py", src)
    return buf.getvalue()

DEFINE = """
def handler(event, context):
    sess = event['request']['session'] or []
    answered = [c for c in sess if c.get('challengeResult') is not None]
    if len(answered) >= 2:
        event['response']['issueTokens'] = True
        event['response']['failAuthentication'] = False
    else:
        event['response']['issueTokens'] = False
        event['response']['failAuthentication'] = False
        event['response']['challengeName'] = 'CUSTOM_CHALLENGE'
    return event
"""

CREATE = """
import json
def handler(event, context):
    sess = event['request']['session'] or []
    rnd = len([c for c in sess if c.get('challengeResult') is not None]) + 1
    event['response']['publicChallengeParameters']  = {'round': str(rnd), 'session_seen': json.dumps(sess)}
    event['response']['privateChallengeParameters'] = {'answer': 'ok'}
    event['response']['challengeMetadata'] = 'ROUND_%d' % rnd
    return event
"""

VERIFY = """
def handler(event, context):
    event['response']['answerCorrect'] = True
    return event
"""

lam = ms("lambda"); idp = ms("cognito-idp")
role = "arn:aws:iam::000000000000:role/lambda-role"
arns = {}
for name, src in [("def-trig", DEFINE), ("crt-trig", CREATE), ("vfy-trig", VERIFY)]:
    try: lam.delete_function(FunctionName=name)
    except Exception: pass
    arns[name] = lam.create_function(FunctionName=name, Runtime="python3.12", Role=role,
        Handler="index.handler", Code={"ZipFile": zip_handler(src)}, Timeout=10)["FunctionArn"]

pool = idp.create_user_pool(PoolName="custom-auth-repro", LambdaConfig={
    "DefineAuthChallenge": arns["def-trig"],
    "CreateAuthChallenge": arns["crt-trig"],
    "VerifyAuthChallengeResponse": arns["vfy-trig"],
})["UserPool"]["Id"]
cid = idp.create_user_pool_client(UserPoolId=pool, ClientName="c",
    ExplicitAuthFlows=["ALLOW_CUSTOM_AUTH"], GenerateSecret=False)["UserPoolClient"]["ClientId"]
idp.admin_create_user(UserPoolId=pool, Username="u@example.com", MessageAction="SUPPRESS")

init = idp.initiate_auth(ClientId=cid, AuthFlow="CUSTOM_AUTH",
                         AuthParameters={"USERNAME": "u@example.com"})
print("round 1 session_seen:", init["ChallengeParameters"].get("session_seen"))

resp = idp.respond_to_auth_challenge(ClientId=cid, ChallengeName="CUSTOM_CHALLENGE",
        Session=init["Session"], ChallengeResponses={"USERNAME": "u@example.com", "ANSWER": "ok"})
print("round 2 session_seen:", resp.get("ChallengeParameters", {}).get("session_seen"))


`round 2 session_seen` is the session the round-2 `CreateAuthChallenge` received — i.e. the
completed round-1 challenge.
```

### Expected behavior

Per the AWS docs for the
[Define auth challenge Lambda trigger](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-define-auth-challenge.html):

> **session** — An array of `ChallengeResult` elements. Each contains: **challengeName**,
> **challengeResult** (`true`/`false`), **challengeMetadata** (your name for the custom challenge).

So `round 2 session_seen` should be **one** element carrying both fields:

```json
[ { "challengeName": "CUSTOM_CHALLENGE", "challengeResult": true, "challengeMetadata": "ROUND_1" } ]
```

Actual on ministack 1.3.59 — **two** split elements:

```json
[ { "challengeName": "CUSTOM_CHALLENGE", "challengeResult": null, "challengeMetadata": "ROUND_1" },
  { "challengeName": "CUSTOM_CHALLENGE", "challengeResult": true, "challengeMetadata": null } ]
```

### MiniStack version

1.3.59

### SDK / tool used

boto3 1.42.88 (Python 3.14); Cognito CUSTOM_AUTH trigger Lambdas on the python3.12 runtime

AdigaAkhil:
Fixes #835.

The CUSTOM_AUTH `RespondToAuthChallenge` paths appended a second, metadata-less `session` entry for the verify result, splitting one round across two records. AWS records **one** `ChallengeResult` per round, carrying both `challengeMetadata` (from `CreateAuthChallenge`) and `challengeResult` (from `VerifyAuthChallengeResponse`) — so multi-round flows that read both fields from the same element (e.g. magic-link → SMS-OTP) never advanced.

The new `_update_pending_challenge_result` helper records the result on the pending round in place, with a fallback append for empty/malformed history. Applied at both `_respond_to_auth_challenge` and `_admin_respond_to_auth_challenge`.

**Tests:** added regression tests for the standard and admin paths (fail on `main`, pass with the fix). Full `tests/test_cognito_custom_auth.py` suite passes (34/34).

### DIFF
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -2334,10 +2334,10 @@ def _admin_initiate_auth(data):
                     "DefineAuthChallenge returned unexpected response — not issuing tokens, "
                     "not failing auth, and no challengeName set", 400)
 
-        # Add a pending challenge to session before checking if define/create need to happen
-        _append_challenge_to_session(session, "CUSTOM_CHALLENGE", None, None, {}, {})
-
-        # Invoke CreateAuthChallenge
+        # Invoke CreateAuthChallenge. AWS passes an EMPTY session array to the
+        # first CreateAuthChallenge (the round being created is not itself a
+        # session entry), so append the pending round AFTER Create runs —
+        # mirroring the round-2+ path in RespondToAuthChallenge.
         create_result, err = _invoke_create_auth_challenge_trigger(
             pid, cid, username, user_attrs, session, client_metadata
         )
@@ -2357,14 +2357,9 @@ def _admin_initiate_auth(data):
             private_params = {"challenge": "PROVIDE_AUTH_PARAMETERS"}
             challenge_metadata = None
 
-        # Update session with challenge parameters
-        if session["challenges"]:
-            session["challenges"][-1].update({
-                "publicChallengeParameters": public_params,
-                "privateChallengeParameters": private_params,
-                "challengeMetadata": challenge_metadata or session["challenges"][-1].get("challengeMetadata"),
-            })
-            session["last_challenge_metadata"] = challenge_metadata
+        # Append the pending round with the parameters from CreateAuthChallenge.
+        _append_challenge_to_session(session, "CUSTOM_CHALLENGE", None, challenge_metadata,
+                                    public_params, private_params)
 
         # Return challenge to client
         return json_response({
@@ -2681,10 +2676,10 @@ def _initiate_auth(data):
                     "DefineAuthChallenge returned unexpected response — not issuing tokens, "
                     "not failing auth, and no challengeName set", 400)
 
-        # Add a pending challenge to session before checking if define/create need to happen
-        _append_challenge_to_session(session, "CUSTOM_CHALLENGE", None, None, {}, {})
-
-        # Invoke CreateAuthChallenge
+        # Invoke CreateAuthChallenge. AWS passes an EMPTY session array to the
+        # first CreateAuthChallenge (the round being created is not itself a
+        # session entry), so append the pending round AFTER Create runs —
+        # mirroring the round-2+ path in RespondToAuthChallenge.
         create_result, err = _invoke_create_auth_challenge_trigger(
             pid, cid, username, user_attrs, session, client_metadata
         )
@@ -2704,14 +2699,9 @@ def _initiate_auth(data):
             private_params = {"challenge": "PROVIDE_AUTH_PARAMETERS"}
             challenge_metadata = None
 
-        # Update session with challenge parameters
-        if session["challenges"]:
-            session["challenges"][-1].update({
-                "publicChallengeParameters": public_params,
-                "privateChallengeParameters": private_params,
-                "challengeMetadata": challenge_metadata or session["challenges"][-1].get("challengeMetadata"),
-            })
-            session["last_challenge_metadata"] = challenge_metadata
+        # Append the pending round with the parameters from CreateAuthChallenge.
+        _append_challenge_to_session(session, "CUSTOM_CHALLENGE", None, challenge_metadata,
+                                    public_params, private_params)
 
         # Return challenge to client
         return json_response({
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -818,6 +818,27 @@ def _append_challenge_to_session(session: dict, challenge_name: str,
     session["last_challenge_metadata"] = challenge_metadata
 
 
+def _update_pending_challenge_result(session: dict, challenge_result: bool | None) -> None:
+    """Record a VerifyAuthChallenge result on the round it belongs to.
+
+    AWS records ONE ChallengeResult per CUSTOM_AUTH round, carrying BOTH the
+    metadata (set by CreateAuthChallenge) and the result (from
+    VerifyAuthChallengeResponse). Update the pending round's result in place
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks @AdigaAkhil !

Issue #91, opened 2026-04-03
Feature Request: Add AWS Bedrock Service Emulation


Vigneshk5:
### AWS Service name

AWS Bedrock

### Operations needed

<html><head></head><body><p><strong>Add AWS Bedrock emulation to MiniStack</strong></p>
<hr>
<p><strong>4 new services</strong></p>

Service | Purpose
-- | --
bedrock | Models, guardrails, fine-tuning, provisioned throughput
bedrock-runtime | InvokeModel, Converse, streaming
bedrock-agent | Agents, knowledge bases, flows
bedrock-agent-runtime | InvokeAgent, Retrieve, RetrieveAndGenerate

<p>No new dependencies — mock uses stdlib only.</p>
<hr>
<p><strong>Contributing</strong></p>
<p>I'd love to help build and maintain this — Bedrock emulation is something I rely on heavily and I'm happy to take ownership of these service files long-term. Happy to start with a draft PR, write the initial tests, or pick up issues as they come in. Let me know how I can best contribute.</p></body></html>

### Use case

The biggest pain point today is the Converse API, it's the main interface I use, and there's no way to test it without live credentials. Even a deterministic mock that returns the right response shape would eliminate a whole class of bugs that only surface in staging.

### Would you like to contribute this?

- [x] Yes, I'd like to implement this service and open a PR
- [x] I can help review/test an implementation

Vigneshk5:
## How inference works:

Mock mode (default): Returns deterministic, model-family-aware responses. Detects the model from modelId (e.g. anthropic.claude-*, amazon.titan-*) and returns the correct provider format. No external calls.

Proxy mode (optional): Set BEDROCK_PROXY_URL to forward requests to Ollama, Docker Model Runner, or any OpenAI-compatible API. Set BEDROCK_PROXY_TYPE to ollama, docker model runner, or openai accordingly. Falls back to mock if unreachable.

Nahuel990:
@Vigneshk5 Thanks for the detailed proposal!

We’re aiming to keep MiniStack focused on the most widely used AWS services so we can maintain a high level of quality and support. While Bedrock is an interesting addition, it’s still relatively niche for local development (most users tend to mock LLM interactions at the application layer rather than the infrastructure layer) 

We’ll keep this issue open to gauge community interest, but it’s not something we’re planning to prioritize in the near term. If demand grows, we can definitely revisit it.

Nahuel990:
Hi @mangoye,

[...]

PR #837, opened 2026-06-05
feat(bedrock,msk): Bedrock family + MSK 

Nahuel990:
Adds bedrock-runtime (8 ops), bedrock control plane (66 ops), bedrock-agent (72 ops), bedrock-agent-runtime (31 ops), bedrock-mantle (OpenAI Chat Completions), and Amazon MSK Kafka control plane. All request/response shapes, HTTP methods, status codes, and error envelopes verified against botocore service-2.json (bedrock-2023-04-20, bedrock-runtime-2023-09-30, bedrock-agent-2023-06-05, bedrock-agent-runtime-2023-07-26, kafka-2018-11-14).

bedrock-runtime:
  - Converse + ConverseStream with vnd.amazon.eventstream framing (prelude CRC + message CRC, messageStart -> contentBlockDelta+ -> contentBlockStop -> messageStop -> metadata sequence)
  - InvokeModel + InvokeModelWithResponseStream with family-specific request and response bodies (Anthropic Messages, Titan, Llama, Mistral, Cohere, AI21, Nova)
  - ApplyGuardrail, StartAsyncInvoke, GetAsyncInvoke, ListAsyncInvokes
  - Mock by default; optional MINISTACK_BEDROCK_PROXY_URL forwards to any OpenAI-compatible /v1/chat/completions endpoint (Ollama, llama.cpp, vLLM, LiteLLM) with silent fallback to mock on unreachable proxy
  - x-amzn-bedrock-input-token-count + x-amzn-bedrock-output-token-count response headers

bedrock control plane:
  - 22-model foundation catalog (Claude 3/3.5, Nova, Titan, Llama 3.1/3.2, Mistral, Cohere Command R, AI21 Jamba); region-aware ModelArns
  - System-defined + application-managed inference profiles (us/eu/apac)
  - Guardrails CRUD + versions, Custom Models, Imported Models, Provisioned Throughput, Model Customization/Import/Copy/Invocation/Evaluation Jobs, Marketplace Endpoints, Prompt Routers, Model Invocation Logging Config, Use Case for Model Access, Foundation Model Agreements, Tags

bedrock-agent:
  - Agent CRUD + Prepare + Versions + Aliases + ActionGroups + Collaborators + Knowledge Base associations
  - KnowledgeBase CRUD, DataSource CRUD, IngestionJob lifecycle, KnowledgeBaseDocuments
  - Flow CRUD + Prepare + Versions + Aliases + ValidateFlowDefinition
  - Prompt CRUD + Versions, Tags

bedrock-agent-runtime:
  - InvokeAgent + InvokeInlineAgent with eventstream chunk envelope (base64-encoded bytes)
  - Retrieve, RetrieveAndGenerate, RetrieveAndGenerateStream (eventstream), Rerank, GenerateQuery, OptimizePrompt
  - Sessions full CRUD + Invocations + InvocationSteps
  - InvokeFlow + StartFlowExecution + GetFlowExecution + ListFlowExecutions + StopFlowExecution + Events + Snapshot
  - GetAgentMemory + DeleteAgentMemory, Tags

bedrock-mantle:
  - OpenAI /v1/chat/completions wire shape
  - SSE chunked streaming with [DONE] terminator
  - Accepts string and list-of-parts content
  - Supports system, developer, tool, user, assistant roles

msk (kafka):
  - Clusters CRUD + ListClusters with name filter + GetBootstrapBrokers honoring MINISTACK_MSK_BOOTSTRAP* env vars (Glue-pattern proxy to user-supplied Redpanda / Kafka / KRaft)
  - Configurations CRUD + revisions
  - SCRAM secrets associate/disassociate, Nodes, Tags
  - Region-isolated state via AccountRegionScopedDict

routing:
  - All 4 Bedrock services share SigV4 credential scope "bedrock" (verified via botocore signing_name)
  - Disambiguated by host header, path prefix, path casing (bedrock-agent uses lowercase /agentaliases, bedrock-agent-runtime uses camelCase /agentAliases), and HTTP method for shared paths
  - POST /flows/{id}/aliases/{id} routes to runtime InvokeFlow; GET/PUT/DELETE route to control-plane FlowAlias

state:
  - All 5 new modules registered in app.py _state_map with symmetric load_state() restore at module top
  - AccountRegionScopedDict used consistently for region isolation

tests:
  - 150 new boto3 round-trip parity tests + symmetry coverage
  - Existing 164 router and persistence regression tests remain green


dcabib:
Went through the whole thing — pulled the branch, ran the 6 suites (150 pass), read the service files. Solid PR. Coverage maps to what #91 asked for: Converse (the main pain point from the issue) is tested deep — shape, multi-turn, 7 model families, token-count headers — plus ConverseStream, InvokeModel, the agent stack, and OpenAI-compatible chat completions via mantle.

A few things worth a look. Only the first is a real bug, rest are nits + a scope question.

**1. `_USE_CASE` escapes reset() and persistence — bedrock.py:1232**
It's declared and used by PutUseCaseForModelAccess, but it's not in the `reset()` loop (L210-218) and not in `get_state()`/`restore_state()`. So `/_ministack/reset` leaks it across tests, and it doesn't survive a persist/restore cycle. Every other store in the file is wired into all three. One-liner in three places.

**2. Proxy call blocks the event loop — bedrock_runtime.py:182,222**
`_proxy_to_openai_chat` is a sync `urllib.urlopen(timeout=30)` called from the async handler. Since the gateway is single-port ASGI, a slow proxy stalls every other request for up to 30s. Only fires when `MINISTACK_BEDROCK_PROXY_URL` is set, so low blast radius, but it's the same trap the API Gateway HTTP proxy already solved by offloading to a worker thread. Might want the same treatment here.

**3. MSK BrokerId serializes as float — msk.py:299**
`float(i)` -> `1.0` on the wire. Real MSK returns an int; boto3 is lenient but strict JSON clients (Go SDK) reject it. Same class of thing as the ECS timestamp fix in the changelog. `int(i)` does it.

**Scope question, not a blocker:** mantle ships Chat Completions, but `/anthropic/v1/messages` and `/v1/responses` (the other two mantle surfaces we touched on in #91) aren't here — anthropic-style is reachable via InvokeModel but not the native SDK-drop-in path. Intentional for v1.4.0 and a follow-up, or did they fall off the list? Either's fine, just want to confirm it's a choice.

Couple things I checked that turned out fine, flagging so you know they were looked at: eventstream CRC framing (correct — Py3 crc32 is already unsigned), MSK is in _state_map, agent routing disambiguation by host/path holds up.

Nice work. Non-blocking from me apart from #1.

PR #839, opened 2026-06-05
Validate AppSync tag and authorizer ARNs

Areson:
## Why
AppSync tag and Lambda authorizer paths should resolve ARNs by account and region instead of falling back to same-named local resources.

## What
- Validate AppSync tag ARNs against the stored API ARN
- Resolve Lambda authorizers with the scoped Lambda ARN helper
- Add regressions for wrong-region API tags and wrong-region authorizer ARNs

## References
- `uv run --extra dev ruff check ministack/services/appsync.py tests/test_appsync.py tests/conftest.py`
- `uv run --extra dev pytest tests/test_appsync.py -q`

### DIFF
--- a/ministack/services/appsync.py
+++ b/ministack/services/appsync.py
@@ -24,6 +24,7 @@ import os
 import re
 import time
 
+from ministack.core.arn import ArnParseError, parse_arn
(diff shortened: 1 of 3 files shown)


Nahuel990:
Thanks!

PR #854, opened 2026-06-08
feat(iceberg-rest): Apache Iceberg REST Catalog endpoint (Glue-backed)

yonatoasis:
## What

Adds the **Glue Iceberg REST Catalog data plane** to MiniStack, mirroring AWS Glue's `glue.<region>.amazonaws.com/iceberg` endpoint so DuckDB's `iceberg` extension can ATTACH:

```sql
ATTACH '000000000000' AS glue_catalog (
    TYPE iceberg,
    ENDPOINT 'localhost:4566/iceberg',
    AUTHORIZATION_TYPE 'sigv4'
);
```

Read-path subset of the [Iceberg REST OpenAPI spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml): `GET /v1/config`, ListNamespaces, GetNamespace, ListTables, LoadTable, HEAD/TableExists. Writes return `501` with the spec's error envelope.

### Shape (per review feedback — thanks @Nahuel990)

AWS serves **two distinct Iceberg REST endpoints** told apart by SigV4 signing name and prefix shape: Glue (`/v1/catalogs/{catalog}/...`, signed `glue`) and S3 Tables (`/v1/{bucket-arn}/...`, signed `s3tables`). This PR follows that:

- The Glue-backed read lives **inside `services/glue.py`** — one file per service, no new top-level module.
- `/v1/config` returns `defaults.prefix = "catalogs/{warehouse}"`, so clients build Glue-shaped URLs.
- **Zero router changes.** The existing credential-scope dispatch already lands `glue`-signed requests on `glue.py`; a path check inside `handle_request` tells the Iceberg REST protocol apart from the X-Amz-Target JSON RPC surface.
- The **s3tables `/iceberg` data plane is untouched** — requests signed `s3tables` land there exactly as before.

### Behavior notes

- A Glue table participates iff `Parameters["metadata_location"]` points at a metadata.json on MiniStack S3 (written by an external engine — Trino/Spark — then registered in Glue; MiniStack doesn't write metadata.json itself). Non-Iceberg Glue tables are hidden from the surface.
- `metadata.json` is passed through verbatim — no parsing/transforming of Iceberg metadata.
- LoadTable returns the S3 client `config` overrides (endpoint/creds/path-style/region) pointing at MiniStack, since unlike real AWS the client has no ambient creds that reach our S3.
- Missing/unparseable metadata.json → `404 NoSuchTableException`, never `200` with empty metadata (DuckDB would treat that as a real-but-empty table).
- Responses use `application/json` + the spec's `{"error": {"message", "type", "code"}}` envelope, not the AWS x-amz-json-1.0 flavor.
- DuckDB hardcodes `https://`, so DuckDB callers need `USE_SSL=1` (existing `core/tls.py` self-signed cert, no changes) and a trusted cert. A one-shot warning logs if the endpoint is hit with TLS off.

## Test plan

- [x] `pytest tests/test_glue_iceberg_rest.py` — 19 tests: routing by signing scope (glue → glue.py, s3tables → s3tables.py), config prefix shape, namespace 404s, non-Iceberg-table filtering, LoadTable verbatim passthrough, 404 on missing/unparseable metadata.json, HEAD, 501 for writes, bare-warehouse prefix rejected, JSON RPC surface unaffected
- [x] `pytest tests/test_persistence.py` — 129/129
- [ ] End-to-end DuckDB ATTACH against `USE_SSL=1` MiniStack (requires signed requests — e.g. `curl --aws-sigv4 "aws:amz:us-east-1:glue" --user test:test`)


Nahuel990 (dismissed):

Nahuel990:
Hi @yonatoasis. I almost approved this, then noticed something I want to explain rather than just merge or close.

The PR serves two different AWS surfaces from one path. On real AWS there are two separate Iceberg REST endpoints:

- `s3tables.<region>.amazonaws.com/iceberg` (signing name `s3tables`)
- `glue.<region>.amazonaws.com/iceberg` (signing name `glue`)

They are distinct services, told apart by the SigV4 signing name and by prefix shape:

- Glue uses `/v1/catalogs/{catalog}/...`
- S3 Tables uses `/v1/{url-encoded-bucket-arn}/...`

They are not one combined catalog.

This PR unions `glue._tables` and `s3tables._tables` under our single `/iceberg` handler and adds a new top-level service module, and we keep one file per service, so it does not line up with how AWS shapes it.

What you are trying to do is reasonable: let DuckDB read Glue-registered Iceberg tables. On AWS that comes from the Glue endpoint specifically. So in ministack the AWS-faithful version would be a Glue-scoped path, not a blend:

- Leave the existing `s3tables` `/iceberg` data plane alone.
- Add the Glue-backed read inside `glue.py`, served when the request is signed with the `glue` scope.

Ministack already routes by SigV4 credential scope, so a client pointed at `localhost:4566/iceberg` signing as `glue` would land on the Glue backing, and signing as `s3tables` would land on the existing one.

The pointer would come from the Glue table's `Parameters["metadata_location"]`, with the `metadata.json` served from Minstack S3.

Two things worth knowing if you want to take it further.

1 - Ministack does not currently write `metadata.json` to S3, so the read only resolves for tables whose metadata and data were written to Ministack S3 by an external engine and then registered in Glue

2 - I pulled `amazon/aws-glue-libs:5` to check, and that image is an ETL runtime with Iceberg client libraries only, no REST catalog server and no exposed ports, so it cannot serve the endpoint itself. The catalog has to be emulated by us or run as a separate open-source Iceberg REST catalog (outside Ministack)

If you want to pursue the Glue-scoped version, that is the shape we would take and I am happy to help

yonatoasis:
Thanks @Nahuel990 — you're right, and I've restructured to the shape you described. Pushed in eddf291:

- **Reverted the blend entirely**: `services/iceberg_rest.py` is gone, the s3tables `/iceberg` data plane is back untouched, and the router override is removed — `core/router.py` and `app.py` are now identical to main.
- **Glue-backed read lives inside `glue.py`**, served when the request is signed with the `glue` scope. Since the router already dispatches by credential scope, no routing changes were needed at all — a path check at the top of `glue.handle_request` separates the Iceberg REST protocol from the X-Amz-Target JSON RPC surface.
- **Glue's prefix shape**: `/v1/config` returns `defaults.prefix = "catalogs/{warehouse}"`, so clients build `/v1/catalogs/{catalog}/namespaces/...` URLs. The bare-warehouse prefix (S3 Tables' shape) is rejected with 501 rather than guessed at.
- The pointer comes from `Parameters["metadata_location"]` with metadata.json served verbatim from MiniStack S3, as you outlined. Missing/unparseable metadata.json → 404 `NoSuchTableException` rather than a 200 with empty metadata.

On your two notes: agreed on both. The read-only stance assumes an external engine wrote metadata/data to MiniStack S3 and registered the table in Glue — that's exactly our use case (Trino writes, DuckDB reads). And thanks for checking `aws-glue-libs:5`; I've kept the surface emulated in-process rather than attempting any delegation.

Tests cover the routing split explicitly (signed-as-glue → glue.py, signed-as-s3tables → s3tables.py) plus the read endpoints. Happy to adjust further.

Nahuel990 (approved):

Nahuel990:
Thanks!

PR #860, opened 2026-06-09
feat(iam): GenerateCredentialReport, GetCredentialReport

lahmish:
## What

Adds the IAM credential report generate→get handshake to the existing IAM service (`ministack/services/iam.py`):

- `GenerateCredentialReport`
- `GetCredentialReport`

## Why

Security auditing tools call `generate-credential-report` and then `get-credential-report` to retrieve a CSV snapshot of all IAM users and their credential state (password age, MFA, access key rotation). Today MiniStack returns `InvalidAction`.

## Behavior / AWS parity

- `GenerateCredentialReport` builds and caches a 22-column CSV report (matching the AWS-specified header order). Includes one row per IAM user plus a `<root_account>` row.
- `GetCredentialReport` returns the cached report base64-encoded with `ReportFormat=text/csv` and a `GeneratedTime` timestamp. Returns `ReportNotPresent` (HTTP 410) if generation hasn't been called first — matching AWS parity.
- Columns `password_enabled` and `mfa_active` are derived from live IAM state so they reflect the current account accurately.

## Tests

`tests/test_iam.py`:
- `test_iam_credential_report_get_before_generate` — 410 before generate is called
- `test_iam_credential_report_mfa_and_password` — columns reflect live user state
- `test_iam_credential_report_header` — CSV header row matches the AWS-specified 22-column order

71 total IAM tests pass.

## Notes

- Existing IAM service — no new service file, so no prior issue required per CONTRIBUTING.
- New `_credential_reports` state dict (`AccountScopedDict`) wired into `get_state` / `restore_state` / `reset`.
- README IAM action list + CHANGELOG updated.


lahmish:
**Note on the  change:** `test_elasticache_restore_state_wipes_stale_replication_group_container_ids` calls `_ensure_live_containers()` directly (bypassing HTTP), so it inherits whatever thread-local account ID xdist left from the previous test in the same worker. This caused a flaky CI failure (`111111111111` vs `000000000000`). The fix is one line: `set_request_account_id("000000000000")` at the top of the test to pin the account explicitly. Unrelated to the IAM credential report changes.

Nahuel990 (approved):

Nahuel990:
Thanks @lahmish

PR #864, opened 2026-06-09
Validate Athena tag resource ARNs

Areson:
## Why
Athena tag APIs should validate tag resource ARNs before using them as tag-state keys.

## What
- Adds parser-backed validation for Athena workgroup and data catalog tag resource ARNs.
- Adds regression coverage for accepted local resource shapes and malformed, wrong service, wrong partition, wrong account, wrong Region, and unsupported resource ARNs.

## Validation
- `uv run --offline --extra dev ruff check ministack/services/athena.py tests/test_athena.py`
- `uv run --offline --extra dev pytest tests/test_athena.py -v`

### DIFF
--- a/ministack/services/athena.py
+++ b/ministack/services/athena.py
@@ -26,6 +26,7 @@ import xml.etree.ElementTree as ET
 from datetime import date
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

PR #870, opened 2026-06-09
Validate Scheduler tag resource ARNs

Areson:
## Why
EventBridge Scheduler tag APIs should validate schedule and schedule-group ARNs before reading or mutating tags.

## What
- Adds parser-backed validation for local Scheduler schedule and schedule-group tag resource ARNs.
- Adds focused coverage for valid local resources, malformed and out-of-scope ARNs, missing resources, and same-name cross-Region no-fallback behavior.

## Validation
- `uv run --offline --extra dev ruff check ministack/services/scheduler.py tests/test_scheduler.py tests/test_scheduler_arn_adoption.py`
- `uv run --offline --extra dev pytest tests/test_scheduler.py tests/test_scheduler_arn_adoption.py -v`

### DIFF
--- a/ministack/services/scheduler.py
+++ b/ministack/services/scheduler.py
@@ -16,7 +16,9 @@ import logging
 import os
 import re
 import time
+from urllib.parse import unquote
 
+from ministack.core.arn import ArnParseError, parse_arn
 from ministack.core.persistence import load_state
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks!

PR #887, opened 2026-06-10
fix(sfn): deep-convert ecs:runTask Parameters so ContainerOverrides reach the task

### LINKED ISSUE #886: Step Functions: ecs:runTask silently drops ContainerOverrides — Parameters Pascal→camel conversion is shallow, so env never reaches the container
### Affected Service

Step Functions (the `ecs:runTask` / `ecs:runTask.sync` task-state bridge into ECS)

### What happened?

A state machine task state using `arn:aws:states:::ecs:runTask.sync` with the standard AWS ASL Parameters shape:

```json
"Overrides": {
  "ContainerOverrides": [
    { "Name": "my-container", "Environment": [ {"Name": "FOO", "Value": "bar"} ] }
  ]
}
```

launches the Docker container fine, but **none of the override environment variables reach the container** — it runs with task-definition env only. Nothing fails: the task runs, the `.sync` state succeeds, and the execution output even echoes the overrides back, so the drop is completely silent.

Root cause (read from the image, present in at least 1.3.30 and 1.3.61): `_invoke_ecs_run_task` in `ministack/services/stepfunctions.py` converts the ASL Parameters to camelCase for the ECS internals via `_pascal_to_camel`, but that function only converts **top-level** keys. So `Overrides` becomes `overrides`, while everything nested stays PascalCase (`ContainerOverrides`, `Environment`, `Name`/`Value`). `ecs._run_task` then reads `data.get("overrides", {}).get("containerOverrides", [])` and gets `[]`.

Since AWS's documented ASL shape for this integration is PascalCase (https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html), there's no input that both real AWS and MiniStack accept — every AWS-valid state machine loses its overrides. Calling ECS `RunTask` directly via the API/CLI is unaffected (boto sends camelCase natively); only the Step Functions bridge hits this. `NetworkConfiguration`'s nested `AwsvpcConfiguration` is left PascalCase by the same shallow pass.

### Reproduction steps

```shell
docker run -d --rm --name ms-repro -v /var/run/docker.sock:/var/run/docker.sock \
  -p 4566:4566 ministackorg/ministack:1.3.61

export AWS_ENDPOINT_URL=http://localhost:4566 AWS_ACCESS_KEY_ID=test \
  AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=eu-west-2

aws ecs register-task-definition --family repro-runner \
  --requires-compatibilities FARGATE --network-mode awsvpc --cpu 256 --memory 512 \
  --container-definitions '[{"name":"runner","image":"alpine:3","command":["sh","-c","env | sort; sleep 2"],"environment":[{"name":"FROM_TASKDEF","value":"yes"}]}]'

aws stepfunctions create-state-machine --name repro --role-arn arn:aws:iam::000000000000:role/x \
  --definition '{
    "StartAt": "Run",
    "States": {
      "Run": {
        "Type": "Task",
        "Resource": "arn:aws:states:::ecs:runTask.sync",
        "Parameters": {
          "Cluster": "repro-cluster",
          "TaskDefinition": "arn:aws:ecs:eu-west-2:000000000000:task-definition/repro-runner:1",
          "LaunchType": "FARGATE",
          "Overrides": {"ContainerOverrides": [{"Name": "runner", "Environment": [
            {"Name": "FROM_OVERRIDES", "Value": "yes"},
            {"Name": "RUN_ID", "Value.$": "$$.Execution.Name"}
          ]}]}
        },
        "End": true
      }
    }
  }'

aws stepfunctions start-execution --name repro-exec-1 \
  --state-machine-arn arn:aws:states:eu-west-2:000000000000:stateMachine:repro
sleep 8

# Env actually seen by the spawned container:
docker logs $(docker ps -a --filter label=ministack=ecs -q | head -1)
# FROM_TASKDEF=yes is present; FROM_OVERRIDES and RUN_ID are missing.
```

### Expected behavior

The container's environment includes `FROM_OVERRIDES=yes` and `RUN_ID=repro-exec-1`, matching real AWS behaviour for `ecs:runTask` ContainerOverrides (including `"Value.$"` JSONPath entries, which the interpreter already resolves correctly — they're dropped with the rest at the ECS hand-off).

Making `_pascal_to_camel` recurse into dicts/lists fixes it locally (we're running a patched image built on 1.3.61); happy to open a PR with that plus a regression test if useful.

### MiniStack version

1.3.61 (also reproduced on 1.3.30)

### SDK / tool used

AWS CLI 2.x against the Step Functions + ECS emulators; observed via docker logs of the spawned task container

lucasmfraser:
Fixes #886.

## What

`_pascal_to_camel` at the SFN→ECS hand-off (`_invoke_ecs_run_task`) converted only the **top-level** keys of an `ecs:runTask` state's Parameters. `Overrides` became `overrides`, but everything nested — `ContainerOverrides`, its `Environment` entries, and resolved `"Value.$"` JSONPath values — stayed PascalCase, and `ecs._run_task` (which reads camelCase) silently ignored them. The container launched with task-definition env only, the `.sync` state still succeeded, so the drop was invisible.

Since the AWS-documented ASL shape for this integration is PascalCase ([docs](https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html)), no AWS-valid state machine could get container overrides through. Direct ECS API calls were unaffected — only the Step Functions bridge.

## Fix

Make `_pascal_to_camel` recurse into dicts and lists (it remains used only by the ECS hand-off, where every nested key is an API-shape key, so deep conversion is safe). Nested `NetworkConfiguration.AwsvpcConfiguration` gets converted by the same pass. As a side effect, the stored task record's `overrides` now round-trips through `DescribeTasks` under the real camelCase ECS model instead of being dropped by SDK response parsing.

## Tests

New `test_sfn_integration_ecs_run_task_container_overrides_reach_the_task`: drives a state machine with the documented PascalCase shape (a literal env entry plus `"Value.$": "$$.Execution.Name"`) and asserts the overrides come back camelCase — with the JSONPath entry resolved to the execution name — from both the execution output and `DescribeTasks`.

Verified red on unpatched `main` (the camelCase keys are absent), green with the fix. Full `tests/test_stepfunctions.py` + `tests/test_ecs.py` pass locally against a from-source compose build (one unrelated flake, `test_ecs_service_td_update_replaces_tasks`, fails only under shared-state long runs and passes in isolation on clean `main` too). `ruff check` clean on the touched file (the one pre-existing `I001` in `ministack/services/s3.py` is on `main` already and left alone). CHANGELOG entry added under Unreleased — happy to drop it if you batch entries at release time instead.


Nahuel990:
Thanks @lucasmfraser this will be in the next release, welcome to Ministack!

PR #890, opened 2026-06-11
fix(cloudformation): emit REVIEW_IN_PROGRESS event for change-set-created stacks

maximoosemine:
### What

Creating a stack via CreateChangeSet with --change-set-type CREATE left the
placeholder stack in REVIEW_IN_PROGRESS but seeded an empty event list, so
DescribeStackEvents returned []. AWS records an initial REVIEW_IN_PROGRESS
event for the AWS::CloudFormation::Stack resource at this point. Tools that read
StackEvents[0], notably `sam deploy`, crashed with IndexError: list index out of
range.

The placeholder stack now emits the REVIEW_IN_PROGRESS event on creation,
mirroring the CreateStack path.

### Shape

The fix lives entirely in `services/cloudformation/changesets.py`; it reuses
the existing `_add_event` helper already imported from `stacks.py`.

### Behavior Notes

* `ResourceStatusReason` is left empty to stay consistent with the `CreateStack`
initial event, which also omits it (AWS sets `"User Initiated"` on both).

### Test Plan

- New `test_cfn_change_set_create_emits_review_event`: asserts the stack is in `REVIEW_IN_PROGRESS` and that `StackEvents[0]` is the `AWS::CloudFormation::Stack` / `REVIEW_IN_PROGRESS` event.
- Full `tests/test_cfn.py` passes (88 tests).
- ruff check clean on the changed file.
- End-to-end with samlocal deploy CLI: before this change it crashed with IndexError at deployer.py:396; after, the change set is created and samlocal advances into the stack-event polling loop.


Nahuel990:
@maximoosemine thanks for raising this, can you explain how are you deploying to AWS? (SAM, CDK, TF, OT, Others)

Nahuel990 (approved):

maximoosemine:
@Nahuel990 

> @maximoosemine thanks for raising this, can you explain how are you deploying to AWS? (SAM, CDK, TF, OT, Others)

I deploy to AWS with SAM: `sam build && sam deploy`.

PR #898, opened 2026-06-12
fix(cloudformation): DescribeStackResources handles LogicalResourceId filter

maximoosemine:
## What

`DescribeStackResources` accepted the optional `LogicalResourceId` parameter but never read it, so filtering by `LogicalResourceId` returned all resources in the stack rather than the single matching one. This is inconsistent with the [AWS API](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackResources.html).

The handler now reads `LogicalResourceId`, filters the resource list when present, and returns a `ValidationError` for an unknown id — matching the behavior of the singular `DescribeStackResource` call.

## Shape

The fix lives entirely in `services/cloudformation/handlers.py`; it reuses the existing `_p` helper and `_error` pattern already used by the sibling `_describe_stack_resource` function immediately above it.

## Behavior Notes

* When `LogicalResourceId` is absent the response is unchanged — all resources are returned as before.
* The `ValidationError` message mirrors the singular call: `Resource [<id>] does not exist in stack [<name>]`.

## Test Plan

* New `test_cfn_describe_stack_resources_logical_id_filter`: creates a two-resource stack (S3 + SQS), asserts the filtered call returns exactly the one named resource with the correct `ResourceType`, and a nonexistent id raises `ValidationError`.
* Full `tests/test_cfn.py` passes.
* `ruff check` clean on the changed file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- **CloudFormation — `DescribeStackResources` now honors the `LogicalResourceId` filter** — The handler now reads the optional `LogicalResourceId` parameter and, when present, returns only the matching resource or a `ValidationError` if it does not exist in the stack.
+
+---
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #900, opened 2026-06-12
fix(eventbridge): substitute reserved input-transformer variables

AbdoNile:
## Summary

EventBridge's `InputTransformer` only substituted variables declared in the target's `InputPathsMap`. AWS EventBridge **also** supports a set of *reserved* predefined input variables that require no `InputPathsMap` entry — most importantly `<aws.events.event.json>`, which inlines the full matched event as a raw JSON object.

MiniStack left these **literal**, producing invalid JSON in the delivered payload:

```json
{ "sourceEvent": <aws.events.event.json> }   // <- not valid JSON
```

Any consumer that `JSON.parse`s the target payload then fails. This breaks EventBridge→SQS/Lambda pipelines that embed the source event — notably the common **AWS CDK**-generated `InputTemplate` pattern.

## Change

`ministack/services/eventbridge.py` — in `_apply_input_transformer`, seed the reserved variables **before** the existing replace loop, using `setdefault` so an explicit `InputPathsMap` entry of the same name still wins:

| Variable | Value |
|---|---|
| `<aws.events.event.json>` | full event envelope as a **raw JSON object** (unquoted) |
| `<aws.events.event>` | full event envelope as an **escaped JSON string** |
| `<aws.events.event.ingestion-time>` | event ingestion timestamp |
| `<aws.events.rule-name>` | the rule's name |
| `<aws.events.rule-arn>` | the rule's ARN |

`_apply_input_transformer` gains an optional `rule=None` parameter (the caller `_invoke_target` already had `rule` in scope) so rule-name/rule-arn can resolve.

**Purely additive** — `InputPathsMap` behavior and all other dispatch paths are unchanged. No new dependencies.

## Tests

Added 5 integration tests to `tests/test_eventbridge.py` (matching the existing boto3-against-running-stack style):

- `<aws.events.event.json>` → delivered Body is valid JSON and the embedded event deep-equals source / detail-type / detail.
- `<aws.events.event>` → escaped-string form parses back to the event.
- `<aws.events.rule-name>` / `<aws.events.rule-arn>` resolve to the rule's name/ARN.
- `<aws.events.event.ingestion-time>` is substituted (not left literal).
- `setdefault` precedence: an explicit `InputPathsMap` entry named like a reserved var still wins.

## Verification

- `docker compose up -d --build` then `pytest tests/test_eventbridge.py -v` → **109 passed**, zero regressions.
- `ruff check ministack/` → clean for the changed file.
- TDD: new tests were watched failing against the unfixed image (literal placeholder → `JSONDecodeError`) before the fix.

## Checklist

- [x] Tests passing (`pytest tests/test_eventbridge.py`)
- [x] Linting passes for changed code (`ruff check ministack/`)
- [x] `CHANGELOG.md` entry added (under `## [Unreleased]`)
- [x] Bugfix to an existing service (no new service / `SERVICE_REGISTRY` / router / conftest changes needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- **EventBridge — input transformer reserved variables** — substitute `<aws.events.event.json>`, `<aws.events.event>`, `<aws.events.rule-name>`, `<aws.events.rule-arn>`, and `<aws.events.event.ingestion-time>` so CDK-style templates that embed the source event deliver valid JSON.
+
+---
+
 ## [1.3.62] — 2026-06-11
 
 ### Added
--- a/ministack/services/eventbridge.py
+++ b/ministack/services/eventbridge.py
@@ -969,12 +969,6 @@ def _apply_input_transformer(transformer, event, rule=None):
     except Exception:
         full = {}
 
-    raw_time = event.get("Time", "")
-    if isinstance(raw_time, (int, float)):
-        iso_time = datetime.fromtimestamp(raw_time, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
-    else:
-        iso_time = raw_time
-
     event_envelope = {
         "version": "0",
         "source": event.get("Source", ""),
@@ -982,7 +976,7 @@ def _apply_input_transformer(transformer, event, rule=None):
         "detail": full,
         "account": get_account_id(),
         "region": get_region(),
-        "time": iso_time,
+        "time": event.get("Time", ""),
         "id": event.get("EventId", ""),
         "resources": event.get("Resources", []),
     }
@@ -1001,7 +995,7 @@ def _apply_input_transformer(transformer, event, rule=None):
 
     reserved = {
         "aws.events.event.json": json.dumps(event_envelope),
-        "aws.events.event": json.dumps(json.dumps(event_envelope)),  # escaped string form
+        "aws.events.event": json.dumps({k: v for k, v in event_envelope.items() if k != "detail"}),
         "aws.events.event.ingestion-time": event_envelope.get("time", ""),
     }
     if rule:
--- a/ministack/services/eventbridge.py
+++ b/ministack/services/eventbridge.py
@@ -930,7 +930,7 @@ def _invoke_target(target, event, rule):
 
     input_transformer = target.get("InputTransformer")
     if input_transformer:
-        event_payload = _apply_input_transformer(input_transformer, event)
+        event_payload = _apply_input_transformer(input_transformer, event, rule)
     elif target.get("Input"):
         event_payload = target["Input"]
     elif target.get("InputPath"):
@@ -960,7 +960,7 @@ def _invoke_target(target, event, rule):
         logger.error("EventBridge target dispatch error for %s: %s", arn, e)
 
 
-def _apply_input_transformer(transformer, event):
+def _apply_input_transformer(transformer, event, rule=None):
     input_paths = transformer.get("InputPathsMap", {})
     template = transformer.get("InputTemplate", "")
 
@@ -969,13 +969,20 @@ def _apply_input_transformer(transformer, event):
     except Exception:
         full = {}
 
+    raw_time = event.get("Time", "")
+    if isinstance(raw_time, (int, float)):
+        iso_time = datetime.fromtimestamp(raw_time, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+    else:
+        iso_time = raw_time
+
     event_envelope = {
+        "version": "0",
         "source": event.get("Source", ""),
         "detail-type": event.get("DetailType", ""),
         "detail": full,
         "account": get_account_id(),
         "region": get_region(),
-        "time": event.get("Time", ""),
+        "time": iso_time,
         "id": event.get("EventId", ""),
         "resources": event.get("Resources", []),
     }
@@ -991,6 +998,17 @@ def _apply_input_transformer(transformer, event):
             replacements[var_name] = val if isinstance(val, str) else json.dumps(val)
         except (KeyError, TypeError, IndexError):
             replacements[var_name] = ""
+
+    reserved = {
+        "aws.events.event.json": json.dumps(event_envelope),
+        "aws.events.event": json.dumps(json.dumps(event_envelope)),  # escaped string form
+        "aws.events.event.ingestion-time": event_envelope.get("time", ""),
+    }
+    if rule:
+        reserved["aws.events.rule-name"] = rule.get("Name", "")
+        reserved["aws.events.rule-arn"] = rule.get("Arn", "")
+    for k, v in reserved.items():
+        replacements.setdefault(k, v)
 
     result = template
     for var_name, val in replacements.items():
--- a/tests/test_eventbridge.py
+++ b/tests/test_eventbridge.py
@@ -1745,3 +1745,236 @@ def test_eventbridge_anything_but_wildcard_excludes_matching(eb, sqs):
         QueueUrl=q_url, MaxNumberOfMessages=10, WaitTimeSeconds=1)["Messages"]
     ids = [json.loads(m["Body"])["detail"]["id"] for m in msgs]
     assert ids == ["prod-app-1"]
+
+
+# ---------------------------------------------------------------------------
+# Reserved input-transformer variables
+# ---------------------------------------------------------------------------
+
+def test_eventbridge_input_transformer_event_json(eb, sqs):
+    """<aws.events.event.json> embeds the full event envelope as a raw JSON object."""
+    bus_name = "qa-eb-reserved-evtjson-bus"
+    eb.create_event_bus(Name=bus_name)
+    q_url = sqs.create_queue(QueueName="qa-eb-reserved-evtjson-q")["QueueUrl"]
+    q_arn = sqs.get_queue_attributes(QueueUrl=q_url, AttributeNames=["QueueArn"])["Attributes"]["QueueArn"]
+    eb.put_rule(
+        Name="qa-eb-reserved-evtjson-rule",
+        EventBusName=bus_name,
+        EventPattern=json.dumps({"source": ["myapp.reserved"]}),
+        State="ENABLED",
+    )
+    eb.put_targets(
+        Rule="qa-eb-reserved-evtjson-rule",
+        EventBusName=bus_name,
+        Targets=[
+            {
+                "Id": "t1",
+                "Arn": q_arn,
+                "InputTransformer": {
+                    "InputPathsMap": {},
+                    "InputTemplate": '{"sourceEvent": <aws.events.event.json>}',
+                },
+            }
+        ],
+    )
+    eb.put_events(
+        Entries=[
+            {
+                "Source": "myapp.reserved",
+                "DetailType": "TestEvent",
+                "Detail": json.dumps({"foo": "bar"}),
+                "EventBusName": bus_name,
+            }
+        ]
+    )
+    msgs = sqs.receive_message(QueueUrl=q_url, MaxNumberOfMessages=1, WaitTimeSeconds=1)
+    assert len(msgs.get("Messages", [])) == 1
+    from datetime import datetime as _dt
+    body = json.loads(msgs["Messages"][0]["Body"])
+    assert body["sourceEvent"]["source"] == "myapp.reserved"
+    assert body["sourceEvent"]["detail-type"] == "TestEvent"
+    assert body["sourceEvent"]["detail"] == {"foo": "bar"}
+    assert body["sourceEvent"]["version"] == "0"
+    assert body["sourceEvent"]["time"].endswith("Z")
+    _dt.strptime(body["sourceEvent"]["time"], "%Y-%m-%dT%H:%M:%SZ")
+
+
+def test_eventbridge_input_transformer_event_escaped(eb, sqs):
+    """<aws.events.event> embeds the event envelope as an escaped JSON string."""
+    bus_name = "qa-eb-reserved-evtesc-bus"
+    eb.create_event_bus(Name=bus_name)
+    q_url = sqs.create_queue(QueueName="qa-eb-reserved-evtesc-q")["QueueUrl"]
+    q_arn = sqs.get_queue_attributes(QueueUrl=q_url, AttributeNames=["QueueArn"])["Attributes"]["QueueArn"]
+    eb.put_rule(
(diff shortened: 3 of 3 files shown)


AbdoNile (commented):
## Review — reserved input-transformer variables

Overall this is a clean, well-scoped additive bugfix. Tests pass locally (5 new + full `test_eventbridge.py` suite: **109 passed**) and `ruff check ministack/services/eventbridge.py` is clean. A few notes, none blocking.

### Substring-collision analysis (the high-risk area) — SAFE ✅
The `result.replace(f"<{var_name}>", ...)` loop substitutes placeholders that share a textual prefix (`aws.events.event` ⊂ `aws.events.event.json` ⊂ `aws.events.event.ingestion-time`). This is **collision-free** because every placeholder is wrapped in `<...>` and the replace target *includes the closing `>`*. Replacing `<aws.events.event>` searches for that literal with its trailing `>`; inside `<aws.events.event.json>` the text after the prefix is `.json>`, not `>`, so no spurious match occurs. I verified empirically (all three nested placeholders + a contrived prefix-named `InputPathsMap` key) — output is always correct regardless of dict ordering. So the order of insertion / the `setdefault` calls carry no ordering hazard here.

### `setdefault` precedence — correct ✅
User `InputPathsMap` entries are inserted into `replacements` first, then `reserved` values via `setdefault`, so an explicit map entry wins. The new `test_..._setdefault_precedence` proves it.

### `rule=None` guard — safe ✅
The `if rule:` guard means rule vars are simply absent when `rule` is falsy; the existing single caller (`_invoke_target`) always passes the real rule, so no behavior regression.

### Findings

**[minor — correctness] `time` / `ingestion-time` is an integer epoch, not an ISO-8601 string**
`event["Time"]` is stored as `int(time.time())` (see `put_events`, ~line 710/718). The envelope at line 978 uses it raw, so:
- `<aws.events.event.json>` emits `"time": 1749686400`
- `<aws.events.event.ingestion-time>` resolves to that same integer

Real AWS emits ISO-8601 (e.g. `"2026-06-12T10:00:00Z"`). Note the **non-transformer** path in `_invoke_target` (lines 913–917) already converts this via `iso_time` — consider reusing that conversion in the transformer envelope so both "full event" representations agree and match AWS. The `test_..._ingestion_time` test only asserts non-empty / not-the-placeholder, so it doesn't catch the format mismatch.

**[minor — consistency] transformer envelope diverges from the canonical envelope**
The envelope built here (lines 972–981) omits `version` ("0") and uses raw `time`, whereas the canonical event JSON in `_invoke_target` (lines 919–929) includes `version` and ISO time. Since `<aws.events.event.json>` is documented to users as "the full event," the two representations ideally should be identical. Factoring out a shared envelope builder would fix this and the `time` issue together.

**[nit — test] escaped-form test under-asserts**
`test_..._event_escaped` only checks the inner string parses and `source` matches. Asserting one more field (e.g. `detail-type`) would more fully prove the escaped JSON round-trips the whole envelope. Optional.

### CHANGELOG
The `[Unreleased] → Fixed` entry is accurate and well-written.

AbdoNile:
**Review findings addressed** in `fa888e0`:

- **#1 (correctness)** — the transformer's event envelope now renders `time` as ISO-8601 (reusing the same conversion as the non-transformer dispatch path), so `<aws.events.event.json>`'s `time` field and `<aws.events.event.ingestion-time>` match real AWS output instead of a raw epoch int. The `ingestion-time` test now asserts ISO-8601 (`endswith("Z")` + `strptime(...)`).
- **#2 (consistency)** — added `"version": "0"` to the envelope so the embedded event matches the canonical EventBridge event shape. Scoped to `_apply_input_transformer` only; the other dispatch path is untouched.
- **#3 (test nit)** — the escaped-string test now also asserts the inner `detail-type` round-trips.

`pytest tests/test_eventbridge.py` → 109 passed; `ruff check ministack/services/eventbridge.py` clean.

AbdoNile:
## Independent Verification: fix(eventbridge) reserved input-transformer variables

Verification run: **2026-06-12**, branch `fix/eventbridge-input-transformer-reserved-vars` (commit `fa888e0`), MiniStack built locally from that branch.

---

### A. Full CI-parity test suite

**Stack:** Brought up fresh from the fix branch (`docker compose up -d --build`), all services healthy.

**Parallel set** (`pytest tests/ -n 3 --dist=loadfile -m "not serial"`):
- Result: **3046 passed, 165 failed, 50 skipped** (7m 4s)
- All 165 failures classified **ENVIRONMENT** (not code): connection-refused / EndpointConnectionError under 3-worker parallel load — affects test_multitenancy, test_mwaa, test_codebuild, test_cur, test_efs, test_appsync, test_cfn_custom_resource, test_lambda_proxy. Zero overlap with eventbridge.
- **EventBridge failures: 0**

**Serial set** (`pytest tests/ -m "serial"`) — clean dedicated run on fresh stack:
- Result: **39 passed, 4 failed, 4 skipped** (13.5s)
- 3 failures in `test_apigatewayv2`: DNS — local machine cannot resolve `*.execute-api.localhost:4566` subdomain — ENVIRONMENT
- 1 failure in `test_athena::test_athena_mixed_glue_and_s3_uri`: pre-existing `mock_value != alice` assertion — ENVIRONMENT / pre-existing (test file unchanged by this PR)
- **EventBridge failures: 0**

**EventBridge test suite in isolation** (`pytest tests/test_eventbridge.py`):
```
109 passed in 5.31s
```
100% green, including all 5 new reserved-variable tests.

---

### B. Standalone runtime demonstration

Verified the **CDK-style pattern** end-to-end: custom bus → SQS queue → rule with InputTransformer → put_events → receive_message.

Synthetic event used (no PII):
```json
{"propertyId": "TEST-123", "status": "listed", "price": 450000}
```

InputTemplate:
```
{"detail": <aws.events.event.json>, "sourceEvent": <aws.events.event.json>, "evtEscaped": <aws.events.event>, "ruleName": "<aws.events.rule-name>", "ruleArn": "<aws.events.rule-arn>", "ingested": "<aws.events.event.ingestion-time>"}
```

Delivered SQS body (trimmed for readability):
```json
{
  "sourceEvent": {
    "version": "0",
    "source": "demo.app",
    "detail-type": "PropertyListed",
    "detail": {"propertyId": "TEST-123", "status": "listed", "price": 450000},
    "account": "000000000000",
    "region": "us-east-1",
    "time": "2026-06-12T19:36:11Z",
    "id": "0c8b1810-97ee-4992-b661-bca80484d963",
    "resources": []
  },
  "evtEscaped": "{\"version\": \"0\", \"source\": \"demo.app\", ...}",
  "ruleName": "demo-eb-reserved-verify-rule",
  "ruleArn": "arn:aws:events:us-east-1:000000000000:rule/demo-eb-reserved-verify-bus/demo-eb-reserved-verify-rule",
  "ingested": "2026-06-12T19:36:11Z"
}
```

**Before fix:** the template would have produced literal `<aws.events.event.json>` in the output string — invalid JSON, boto3 SDK callers would `json.loads` and raise.
**After fix:** all placeholders substituted, body is valid JSON.

---

### C. Per-criterion verdict

| # | Criterion | Verdict | Artifact |
|---|-----------|---------|----------|
| 1 | `<aws.events.event.json>` → valid JSON; `sourceEvent` deep-equals event (`source`, `detail-type`, `detail`, `version=="0"`, `time` is ISO-8601) | **PROVEN** | Standalone script output + `test_eventbridge_input_transformer_event_json` PASSED |
| 2 | `<aws.events.event>` → escaped JSON string that `json.loads` back to the event | **PROVEN** | `evtEscaped` confirmed string, inner object matches; `test_eventbridge_input_transformer_event_escaped` PASSED |
| 3 | `<aws.events.rule-name>` → rule Name | **PROVEN** | `ruleName=demo-eb-reserved-verify-rule`; `test_eventbridge_input_transformer_rule_name_and_arn` PASSED |
| 4 | `<aws.events.rule-arn>` → rule Arn | **PROVEN** | `ruleArn` matches `put_rule` response ARN; same test PASSED |
| 5 | `<aws.events.event.ingestion-time>` → substituted, non-empty, ISO-8601 | **PROVEN** | `ingested=2026-06-12T19:36:11Z` (not literal placeholder); `test_eventbridge_input_transformer_ingestion_time` PASSED |
| 6 | `setdefault` precedence: explicit `InputPathsMap` entry named like reserved var wins | **PROVEN** | Script: mapped `aws.events.rule-name → $.source` → got `"demo.prec"` (source value), not the rule name; `test_eventbridge_input_transformer_setdefault_precedence` PASSED |
| 7 | Regression: InputPathsMap-only template (no reserved vars) unchanged | **PROVEN** | `{"transformed": "<src>"}` → `"demo.compat"` as before; `test_eventbridge_input_transformer` PASSED |

**All 7 criteria: PROVEN. No failures.**

---

EventBridge test suite: 109/109. Non-eventbridge failures in the full run are environment-level only (DNS, connection drops under parallel load, pre-existing athena assertion). This PR is clean.

Nahuel990:
Thanks @AbdoNile, overall this looks good, and the `<aws.events.event.json>` support is exactly what's needed. Two things before merge:

 `aws.events.event` doesn't match AWS behavior. I will have to ask you to paste the proper outcome from real aws in order to verify it as per the AWS docs: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html#eb-transform-input-predefined (I will be able to check it myself in around 10 days)


`aws.events.event` should be a raw JSON object without the `detail` field and without additional escaping.

Right now it's a double-`json.dumps` escaped string that still includes `detail`.

Could you change it to `json.dumps({k: v for k, v in event_envelope.items() if k != "detail"})` and update the test to expect an object rather than a string?

Please drop the `time` epoch to ISO change... The integer-epoch `Time` value is intentional and documented. It was introduced in v1.3.25 to prevent a float-parsing issue that was crashing Go and Java SDK parsers.

Changing it only on the transformer path would make delivery behavior inconsistent. If we decide to move to ISO timestamps, that should be a separate, deliberate change applied across both delivery paths, along with documentation updates (but it's not planned) 

Please keep `time` as-is.

Keeping `version: "0"` is correct though; nice catch :)

AbdoNile:
Thanks for the review — addressed in `006259a`.

**`<aws.events.event>` — fixed.** It now renders a raw JSON object with the `detail` field removed (was a double-escaped string that also included `detail`), per the [predefined-variables docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html#eb-transform-input-predefined). `<aws.events.event.json>` stays the full event *including* `detail`. The escaped-form test now asserts an object (`isinstance(dict)`, `version=="0"`, `"detail" not in evt`). I can't run real AWS here to paste live output, so the test encodes the documented shape — flagging so you can confirm it against real AWS on your end.

**`time` — reverted to the stored integer epoch.** `version: "0"` retained.

One thing for your awareness (no change implied unless you want one): the default, non-transformer delivery path in `_invoke_target` currently converts `Time` to ISO-8601 before delivery —

```python
raw_time = event["Time"]
if isinstance(raw_time, (int, float)):
    iso_time = datetime.fromtimestamp(raw_time, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
event_payload = json.dumps({... "time": iso_time ...})
```

So with this revert, the transformer path emits the epoch while the default path emits ISO. Happy to leave it exactly as-is for this PR — just surfacing it in case aligning the two paths is something you'd want to fold into the separate, deliberate change you mentioned.

`pytest tests/test_eventbridge.py` → 109 passed; `ruff check ministack/services/eventbridge.py` clean.

AbdoNile:
## Re-verification after review changes (`d31f1fe`)

Re-ran the full runtime verification against the updated behavior (`<aws.events.event>` → object with `detail` removed; `time` reverted to integer epoch). Stack rebuilt from the branch; synthetic data only.

### Verification Methodology

Behavior-first: the feature is exercised in a real running MiniStack and the delivered output is observed — not inferred from the code. Two complementary methods:

1. **Integration test suite (live stack).** MiniStack Docker image rebuilt from the branch (`docker compose up -d --build`) so the container runs the modified code, then `pytest tests/test_eventbridge.py` — boto3 tests that hit `http://localhost:4566`, each creating a bus + SQS queue + rule with an `InputTransformer`, calling `PutEvents`, then `ReceiveMessage` and asserting on the delivered payload. Plus `ruff check`.
2. **Standalone runtime demonstration (independent of the test file).** A separate boto3 script reproducing the real-world CDK pattern — one rule whose `InputTemplate` embeds all five reserved variables at once — sends a synthetic event and inspects the actual SQS message body. The decisive check is that `json.loads(body)` succeeds (proving valid JSON; before the fix the literal `<aws.events.event.json>` made it invalid), followed by per-criterion structural assertions.

Each of the 7 acceptance criteria is mapped to a concrete artifact (a named passing test + a line of demo output) and classified PROVEN / FAILED / UNVERIFIED.

**Tests:** `pytest tests/test_eventbridge.py` → **109 passed**; `ruff check ministack/services/eventbridge.py` → clean.

**Standalone runtime demo** — one rule whose `InputTemplate` exercises all five reserved variables; delivered SQS body (verbatim, synthetic):

```json
{
  "full": {
    "version": "0", "source": "demo.app", "detail-type": "PropertyListed",
    "detail": {"propertyId": "TEST-123", "status": "listed", "price": 450000},
    "account": "000000000000", "region": "us-east-1",
    "time": 1781354802, "id": "1e44a381-...-40888405f915", "resources": []
  },
  "eventNoDetail": {
    "version": "0", "source": "demo.app", "detail-type": "PropertyListed",
    "account": "000000000000", "region": "us-east-1",
    "time": 1781354802, "id": "1e44a381-...-40888405f915", "resources": []
  },
  "ruleName": "demo-eb-reserved-verify-rule",
  "ruleArn": "arn:aws:events:us-east-1:000000000000:rule/demo-eb-reserved-verify-bus/demo-eb-reserved-verify-rule",
  "ingested": "1781354802"
}
```

| # | Criterion | Verdict |
|---|-----------|---------|
| 1 | `<aws.events.event.json>` → valid JSON, full event **including `detail`**, `version=="0"` | **PROVEN** |
| 2 | `<aws.events.event>` → JSON **object** (not escaped string) with **`detail` removed** | **PROVEN** (`detail` absent; keys: account, detail-type, id, region, resources, source, time, version) |
| 3 | `<aws.events.rule-name>` → rule Name | **PROVEN** |
| 4 | `<aws.events.rule-arn>` → rule Arn | **PROVEN** |
| 5 | `<aws.events.event.ingestion-time>` → substituted (integer epoch, as stored) | **PROVEN** (`"1781354802"`, not literal) |
| 6 | `setdefault` precedence: explicit `InputPathsMap` entry wins | **PROVEN** (`rn="demo.prec"`, the source value) |
| 7 | Regression: `InputPathsMap`-only template unchanged | **PROVEN** (`transformed="demo.compat"`) |

Note `time` is the integer epoch (`1781354802`) per your guidance — both `<aws.events.event.json>` and `<aws.events.event.ingestion-time>` reflect the stored value unchanged.

Nahuel990:
Thanks @AbdoNile this looks better! We'll release it in the next version, welcome to Ministack!

Nahuel990 (approved):

PR #905, opened 2026-06-12
fix(ecs): Fix ECS container command overrides

noynoy83:
## Why
ECS `RunTask` already applied containerOverrides.environment, but containerOverrides.command was ignored, i.e., the task definition's command was passed through unchanged.

## What
- Look up each container's override by name and, when command is present, apply it to a copy of the container definition before starting Docker. 
- Added tests for command overrides.

## Validation
- `uv run pytest tests/test_ecs.py -v`
  - `32 passed`
- `uv run ruff check ministack/services/ecs.py`
  - `All checks passed!`
- `uv run ruff check tests/test_ecs.py`
  - `All checks passed!`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


Nahuel990:
@noynoy83 thanks for this, looks overal ok but please remove the uv.lock file

noynoy83:
@Nahuel990 
Sorry for the confusion. I forgot to exclude the uv.lock file.
I’ve now removed it and rebased the branch onto the latest `ministackorg/ministack:main`. Could you please take another look?

*Note
One CI test is failed, but it seems not to be related to changes on my PR.
I think the failed test is probably a flaky test.

Nahuel990 (approved):

PR #913, opened 2026-06-13
fix(cloudformation): GetTemplateSummary now returns capabilities info

maximoosemine:
## What

The response from `GetTemplateSummary` was missing the `Capabilities` and `CapabilitiesReason` fields. This meant the `aws cloudformation deploy` CLI and any SDK caller could not determine what capabilities a template requires before creating a change set, inconsistent with the [AWS API](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_GetTemplateSummary.html).

The handler now computes and returns `Capabilities` and `CapabilitiesReason` matching AWS behavior:

- `CAPABILITY_NAMED_IAM` — any `AWS::IAM::*` resource with an explicit name property.
- `CAPABILITY_IAM` — any other `AWS::IAM::*` resource.
- `CAPABILITY_AUTO_EXPAND` — template contains a `Transform` key.

`CapabilitiesReason` uses the format derived from the AWS API.

## Shape

The fix is contained to `_get_template_summary` in `services/cloudformation/handlers.py`.

## Behavior Notes

* Templates with no IAM resources and no `Transform` receive no `Capabilities` or `CapabilitiesReason` in the response, unchanged from before.

## Test Plan

* New `test_cfn_get_template_summary`: four sub-cases — no capabilities (S3-only template), `CAPABILITY_NAMED_IAM` (role with explicit `RoleName`), `CAPABILITY_IAM` (role without explicit name), and `CAPABILITY_AUTO_EXPAND` (template with `Transform`). Asserts exact `CapabilitiesReason` string for the named IAM case.
* Full `tests/test_cfn.py` passes.
* `ruff check` clean on the changed file.
* Tested against `aws cloudformation deploy`


maximoosemine:
~~Looks like the CI failure is the result of a flaky test unrelated to these changes.~~

This got sorted out after pushing an unrelated change.

Nahuel990 (approved):

Nahuel990:
Thanks @maximoosemine !

PR #914, opened 2026-06-14
fix(lambda): CreateEventSourceMapping persists FilterCriteria

maximoosemine:
### What

`CreateEventSourceMapping` accepted a `FilterCriteria` parameter but was silently discarding it, so any filter specified at ESM creation time was never applied. Only filters set via `UpdateEventSourceMapping` were persisted. This change fixes that.

### Test plan

- Added tests to `tests/test_lambda.py`.
- Full `tests/test_cfn.py` passes.
- `ruff check` clean on the changed file.


Nahuel990:
Thanks @maximoosemine !

PR #917, opened 2026-06-14
ci: try to improve test performance

jgrumboe:


jgrumboe:
@Nahuel990 I'd like you to review my changes here.

```text
scripts/select_test_shard.py
.github/workflows/ci.yml
```

## TL;DR

- Scan files → count tests → weight = `tests*10 + size`.  
- Sort by weight, greedily dump each file into the lightest shard.  
- CLI lets CI ask for “shard N of M” in either parallel or serial mode.  
- CI runs three parallel shards + one serial shard, keeping the flaky state‑sharing tests out of the parallel pool.
- The new sharding cuts the CI wall‑clock time from roughly five minutes to about three minutes, shaving off around 30 % of the total test‑run duration.

## How the sharding script works (informal walk‑through)

### 1. What the script is for
We want to split the whole test suite into a few chunks (shards) so CI can run them in parallel.  
At the same time we need a *serial* bucket for the tests that mess with shared state (they’re marked `@pytest.mark.serial` or listed in `_SERIAL_TESTS`).

### 2. How we decide how “heavy” a file is
For each `tests/test_*.py` file we compute a **weight**:

```
weight = (number_of_tests * 10) + file_size_in_kib
```

* Test count dominates (×10) – more tests = heavier.  
* File size is just a tiny tie‑breaker.

### 3. The actual sharding algorithm (simple greedy bin‑packing)

1. **Collect** all test files.  
2. **Compute** their weight (using the mode `parallel` or `serial`).  
3. **Sort** the files from heaviest to lightest.  
4. **Loop** over the sorted list:  
   * Find the shard that currently has the **lowest total weight**.  
   * Drop the file into that shard and add its weight to the shard’s total.  

That’s it – a classic “keep the piles as even as possible” approach.

### 4. What I fixed additionally

* Added the two flaky tests (`test_sfn_mock_config_jsonata_assign_applied` and `test_publish_with_appsync_sigv4_scope_on_events_vhost`) to `_SERIAL_TESTS` so they now run in the serial phase.  Reported in #920 

That’s the whole story in a nutshell.

Nahuel990:
Sorry for the slow reply @jgrumboe  on leave with limited internet

LGTM but ran it through the AI for a review. It flagged this:

1. Why the PR is stuck on **"Expected - Waiting for status to be reported"**: the matrix renames the checks to `test (parallel, 0, 3)` etc., so the required `test` check never resolves. Needs a gate job to require instead:

```yaml
test-result:
  name: test
  needs: test
  if: always()
  runs-on: ubuntu-latest
  steps:
    - run: '[ "${{ needs.test.result }}" = "success" ]'
```

2. `fail-fast: false` on the matrix - otherwise one failed shard cancels the rest.

3. Drift guard - the AST selector reimplements pytest collection; correct now, but could silently drop tests if marking changes. A `pytest --collect-only` vs selector-union check makes it loud.

Coverage itself checks out - verified zero tests dropped. **1 and 2 are must-haves.**

jgrumboe:
@Nahuel990 
Thanks for your input. I've addressed everything.

I've now added a "plan" step to get some insights into how sharding is done. Wdyt? Is it worth it?
On the one hand, it provides insight into where test files go and makes shard configuration easier; on the other hand, it adds 30-40 seconds again.

Nahuel990 (approved):

PR #919, opened 2026-06-15
fix(lambda): CloudFormation-created ESMs never poll DynamoDB streams

maximoosemine:
### What

- `_poll_dynamodb_streams` used `if not stream_records: return` as an early-exit condition. Because `_stream_records` is an `AccountScopedDict`, it is falsy when the table has not yet received any writes — causing the poller loop to skip entirely and never initialise the per-ESM stream position. The fix changes the condition to `if stream_records is None: return`, which only exits when the attribute does not exist on the DynamoDB module.
- Stream-position initialization was inside the `if not table_records: continue` block, so it only ran on the first tick that had records. By then `len(table_records)` was already ≥ 1, and those records were skipped as "already seen". Initialization moved before the `continue` block, so position is set eagerly (to 0 for `TRIM_HORIZON`, to `len(table_records)` for `LATEST`) even when the stream is still empty.
- `_lambda_esm_create` in `provisioners.py` used `func_name.rsplit(":", 1)[-1]` to extract the function name from a versioned ARN. Replaced the ad-hoc split with the existing `_resolve_name_and_qualifier` helper and stores `Qualifier` in the ESM dict (field parity with the API path).

### Test plan

- `test_lambda_dynamodb_stream_esm_latest_processes_first_record` — integration test: creates an ESM on an empty DynamoDB stream with `StartingPosition=LATEST`, waits for the poller to tick, then writes the first item and asserts it is delivered (was failing before the fix).
- Full `tests/test_cfn.py` passes.
- `ruff check` clean on the changed file.


Nahuel990 (approved):

Nahuel990:
Thanks @maximoosemine !

PR #922, opened 2026-06-15
fix(ecs): inject containerDefinitions[].secrets (Secrets Manager) into RunTask containers

### LINKED ISSUE #918: ECS RunTask: containerDefinitions[].secrets (Secrets Manager) is not injected into the running container
## Summary

When an ECS task definition's `containerDefinitions[].secrets` references a
Secrets Manager secret ARN (`valueFrom`), MiniStack's ECS RunTask
implementation does not resolve these secrets and inject them as environment
variables into the docker container it starts. Only `containerDefinitions[].environment`
is used.

## Where

In `ministack/services/ecs.py`, both `_build_task_containers` and the
RunTask container-launch code build the container env from
`cdef.get("environment", [])` only:

```python
env = {e["name"]: e["value"] for e in cdef.get("environment", [])}
env.update(env_override)
```

`cdef.get("secrets", [])` is never read, so any `valueFrom` entries are
silently dropped.

## Reproduction

1. Create a Secrets Manager secret and a secret version with some value.
2. Register an ECS task definition with a container that has:
   ```json
   "environment": [{"name": "FOO", "value": "bar"}],
   "secrets": [{"name": "SECRET_VAL", "valueFrom": "<secret arn>"}]
   ```
3. Run the task (with the Docker socket mounted so MiniStack actually starts
   a container).
4. `docker inspect <container> --format '{{range .Config.Env}}{{println .}}{{end}}'`

## Expected

The container's environment includes both `FOO=bar` and
`SECRET_VAL=<resolved secret value>`.

## Actual

Only `FOO=bar` is present; `SECRET_VAL` is missing entirely.

## Environment

- MiniStack 1.3.63 (docker image `ministackorg/ministack:1.3.63`)

## Context

Found while validating a Terraform-managed ECS service (task-canvas backend)
that relies on `secrets` to inject `JWT_SECRET`/DB password from Secrets
Manager. As a workaround we now resolve the secret values via Terraform's
`aws_secretsmanager_secret_version` data source and pass them as
`environment` instead, for the MiniStack-only env. A native fix would let us
use the same task definition shape as real AWS.

kurok:
## What

ECS RunTask built each container's environment from `containerDefinitions[].environment` only. `containerDefinitions[].secrets` (Secrets Manager `valueFrom`) were silently dropped, so task definitions that rely on Secrets Manager started without those variables. This forced users to resolve secrets out-of-band and pass them as `environment`, diverging from their real-AWS task definition shape.

## Fix

- Add a reusable `secretsmanager.resolve_secret_string(secret_id, version_stage="AWSCURRENT")` helper (built on the existing `_resolve`/`_find_stage_version`) so other services can read secret values in-process.
- In `_run_task`, resolve each container's `secrets` and merge them into the env before container overrides are applied. Supports the plain secret ARN (whole `SecretString`) and the `:json-key:` suffix form (selects one field from a JSON secret).

## Scope / limitations

- Only Secrets Manager `valueFrom` is resolved. SSM Parameter Store references and unresolvable secrets are skipped with a warning (no behavior change from today for those).

## Test

Added `test_ecs_run_task_injects_secrets_manager_secrets` (Docker-free, mirrors the existing monkeypatched-docker run_task tests) covering both the whole-secret and json-key forms. New test plus the existing run_task override tests pass locally.

Fixes #918


kurok:
@Nahuel990 would you be able to take a look when you have a moment? Small, self-contained fix for #918 — ECS RunTask now resolves `containerDefinitions[].secrets` (Secrets Manager `valueFrom`, incl. the `:json-key:` form) and injects them into the container env. Added a Docker-free test mirroring the existing run_task tests; it passes locally. CI just needs the workflow approved for this fork PR (currently `action_required`). Thanks!

Nahuel990 (approved):

Nahuel990:
Thanks!!

PR #925, opened 2026-06-15
fix(s3) Preserve WebsiteRedirectLocation

murlock:
When uploading an object to S3 with x-amz-website-redirect-location was discarded.
A test has been added.

I will check if CloudFront Distribution will honor this information.


Nahuel990 (approved):

Nahuel990:
Thank you @murlock ! This will go in next release!

PR #926, opened 2026-06-15
fix(iam): implement instance-profile tagging actions

c-julin:
`TagInstanceProfile`, `UntagInstanceProfile`, and `ListInstanceProfileTags` were not registered in the IAM handler table, so every call failed with `InvalidAction: Unknown IAM action`. Tags were also never stored on the instance-profile object or emitted by `GetInstanceProfile` / `ListInstanceProfiles`.

Changes in `iam.py`:

- Add handlers for `TagInstanceProfile`, `UntagInstanceProfile`, and `ListInstanceProfileTags`, following the existing `TagRole` / `TagUser` / `TagPolicy` pattern.
- Store tags on the instance-profile object, including tags supplied at `CreateInstanceProfile` time (the AWS API accepts `Tags` on create).
- Emit a `Tags` member from `_instance_profile_xml` so `GetInstanceProfile` and `ListInstanceProfiles` read them back, mirroring `_user_xml` / `_role_xml`.

Added `test_iam_instance_profile_tags` (tag -> list -> untag round-trip) and `test_iam_instance_profile_tags_serialized_in_get` (create-time and after-the-fact tags read back via both `GetInstanceProfile` and `ListInstanceProfiles`). All existing IAM tests still pass.

Ran into this using ministack as the AWS endpoint for a Terraform setup: `aws_iam_instance_profile` calls `TagInstanceProfile` on create and reads the tags back to detect drift, so without the read-back a second `terraform apply` never converged to "No changes".


Nahuel990:
Thank you @c-julin !!

PR #929, opened 2026-06-16
fix(ec2): accept legacy source-group params and return ReferencedGroupInfo (#916)

### LINKED ISSUE #916: [Bug] AuthorizeSecurityGroupIngress with source security group (UserIdGroupPairs) not returned by DescribeSecurityGroupRules
## Summary
`ec2:AuthorizeSecurityGroupIngress` with a `source_security_group_id`
(`UserIdGroupPairs`) appears to create the rule — it shows up in
`DescribeSecurityGroups` under `IpPermissions` — but the rule is not returned
(or not returned with a matching `SecurityGroupRuleId`) by
`DescribeSecurityGroupRules`. This breaks Terraform's `aws_security_group_rule`
create waiter, which polls `DescribeSecurityGroupRules` for the new rule ID and
times out after ~5 minutes with:

```
Error: waiting for Security Group Rule (sgrule-XXXXXXXXXX) create: couldn't find resource
```

## Reproduction
```sh
docker run -d -p 4566:4566 ministackorg/ministack:1.3.63
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1

VPC_ID=$(aws --endpoint-url=http://localhost:4566 ec2 create-vpc --cidr-block 10.0.0.0/16 --query 'Vpc.VpcId' --output text)
SG_A=$(aws --endpoint-url=http://localhost:4566 ec2 create-security-group --group-name sg-a --description sg-a --vpc-id $VPC_ID --query 'GroupId' --output text)
SG_B=$(aws --endpoint-url=http://localhost:4566 ec2 create-security-group --group-name sg-b --description sg-b --vpc-id $VPC_ID --query 'GroupId' --output text)

aws --endpoint-url=http://localhost:4566 ec2 authorize-security-group-ingress \
  --group-id $SG_B --protocol tcp --port 8080 --source-group $SG_A

# Rule shows up here:
aws --endpoint-url=http://localhost:4566 ec2 describe-security-groups --group-ids $SG_B \
  --query 'SecurityGroups[0].IpPermissions'
# -> [{"IpProtocol":"tcp","FromPort":8080,"ToPort":8080,"UserIdGroupPairs":[],...}]
# Note UserIdGroupPairs is empty even though --source-group was used.

# But not here (empty list, or no rule with the expected ID):
aws --endpoint-url=http://localhost:4566 ec2 describe-security-group-rules \
  --filters Name=group-id,Values=$SG_B
```

## Expected
- `DescribeSecurityGroups` should show the rule's `UserIdGroupPairs` referencing
  `$SG_A`.
- `DescribeSecurityGroupRules` should return a rule for `$SG_B` with
  `ReferencedGroupInfo.GroupId == $SG_A` and a stable `SecurityGroupRuleId`,
  matching what `AuthorizeSecurityGroupIngress` returns in its response.

## Impact on Terraform
```hcl
resource "aws_security_group_rule" "ecs_from_alb" {
  type                     = "ingress"
  from_port                = 8080
  to_port                  = 8080
  protocol                 = "tcp"
  security_group_id        = aws_security_group.ecs_service.id
  source_security_group_id = aws_security_group.alb.id
}
```
`terraform apply` hangs for ~5 minutes then fails with:
```
Error: waiting for Security Group (sg-XXXX) Rule (sgrule-XXXXXXXXXX) create: couldn't find resource
```

Workaround: use an inline `ingress` block with `security_groups = [...]` on
`aws_security_group` instead of a separate `aws_security_group_rule` resource
(the inline form does not poll `DescribeSecurityGroupRules`).

## Environment
- MiniStack: `ministackorg/ministack:1.3.63` (Docker)
- Terraform: 1.15.6
- AWS provider: hashicorp/aws ~> 6.50

kurok:
## Summary

Fixes #916. `AuthorizeSecurityGroupIngress`/`AuthorizeSecurityGroupEgress` silently
dropped a **source security group** when it was sent via the legacy flat top-level
parameter form (the AWS CLI's `--source-group`, i.e. `SourceSecurityGroupId` /
`SourceSecurityGroupName`, also emitted by older SDKs). `_parse_ip_permissions`
only handled the nested `IpPermissions.N.*` structure, so the rule was never
stored — it didn't appear in `DescribeSecurityGroups` (`UserIdGroupPairs` came
back empty) nor in `DescribeSecurityGroupRules`.

A secondary parity gap: even for the nested form, the **Authorize response**
omitted `ReferencedGroupInfo` for group-pair rules (real AWS includes it).

## Root cause

- `_parse_ip_permissions(params, prefix)` loops only over `IpPermissions.{i}.*`
  keys. Legacy callers send `IpProtocol` / `FromPort` / `ToPort` /
  `CidrIp` / `SourceSecurityGroupId` at the top level, so the loop body never
  runs and it returns `[]` → the rule is dropped on create.
- `_sg_rule_xml` (used to build the Authorize response) iterated `IpRanges` /
  `Ipv6Ranges` only; a group-pair rule fell through to a bare `<item>` with no
  `<referencedGroupInfo>`, so the Authorize response disagreed with
  `DescribeSecurityGroupRules` (which already rendered it correctly).

## Fix

- Add `_parse_legacy_ip_permission()` and fall back to it in
  `_parse_ip_permissions()` when no nested `IpPermissions.N.*` keys are present.
  Because all four authorize/revoke ingress+egress handlers share that parser,
  both directions are fixed by the one change. The nested form is untouched
  (the fallback only triggers when the nested parse yields nothing).
- Extend `_sg_rule_xml()` to emit a `<referencedGroupInfo>` item per
  `UserIdGroupPairs` entry, with a `SecurityGroupRuleId` derived identically to
  `DescribeSecurityGroupRules` (`sgr-{sg}-{direction}-{idx}`), so the three APIs
  agree.

Net effect: Terraform's `aws_security_group_rule` create waiter (which polls
`DescribeSecurityGroupRules` for the new rule ID) resolves instead of timing out.

## Tests

Two regression tests in `tests/test_ec2.py`, both verified to fail on `main` and
pass with this change:

- `test_ec2_sg_authorize_ingress_legacy_source_group` — the legacy CLI wire form
  (the fully-dropped path), asserting the rule is visible and ID-consistent across
  `Authorize`, `DescribeSecurityGroups`, and `DescribeSecurityGroupRules`.
- `test_ec2_sg_authorize_ingress_referenced_group_nested` — the nested
  `IpPermissions` form that boto3 / the Terraform provider actually send,
  asserting the Authorize response now includes `ReferencedGroupInfo` with a
  matching `SecurityGroupRuleId`.

`pytest tests/test_ec2.py` → 124 passed. `ruff check ministack/services/ec2.py` → clean.

## Checklist

- [x] Tests added and passing (`pytest tests/test_ec2.py`)
- [x] N/A — bug fix to an existing service (no new service / registry / router / fixture changes)


kurok:
Hi @Nahuel990 🙏 — would you mind taking a look at this when you have a moment?

I ran into it while standing up a Terraform stack against MiniStack: an `aws_security_group_rule` with `source_security_group_id` would hang its create waiter for ~5 min and then fail, because the source-group reference sent via the legacy `--source-group` form was dropped — so the rule never showed up in `DescribeSecurityGroups`/`DescribeSecurityGroupRules`. This parses that form and makes the Authorize response return `ReferencedGroupInfo` consistently, which unblocks the waiter.

Kept it minimal and scoped to the existing parser; added regression tests for both the CLI and nested (boto3/Terraform) wire forms. Happy to adjust anything. Thanks for the great project!

kurok:
CI failure here is **not caused by this PR** — it's a pre-existing flaky test-isolation bug that this PR just happened to expose.

### What failed
`tests/test_ecs.py::test_ecs_run_task_injects_secrets_manager_secrets` → `KeyError: 'SECRET_VAL'`, with:
```
WARNING ecs:ecs.py:1023 ECS: could not resolve secret arn:aws:secretsmanager:...:secret:ecs-secret-plain-... for env var SECRET_VAL
```
This PR only touches `ec2.py`, `test_ec2.py`, and `CHANGELOG.md`.

### Root cause
`test_persistence.py::test_module_cold_import_with_typical_snapshot_does_not_log_restore_failure` pops a service module from `sys.modules`, re-imports it fresh, and never restores the original. Parametrized over every service, cold-importing `ecs` then `secretsmanager` leaves the leaked `ecs` bound to a **stale** `secretsmanager` — so any later test that creates a secret via the live module and resolves it through ECS RunTask fails. It only surfaces when `--dist=loadfile` schedules `test_persistence.py` and `test_ecs.py` onto the same worker with the cold-imports first. Adding this PR's `test_ec2.py` reshuffled that distribution; `main` is green only by scheduling luck.

### Fix
Tracked separately in #934 (restores the original module in both `sys.modules` and the `ministack.services` package attribute). Reproduced and verified locally: the failing sequence goes 1-failed → all-pass.

### To get this PR green
Once #934 merges to `main`, merge `main` into this branch and re-run CI.

Nahuel990:
Sorry for the delay @kurok was away for a few days... This looks good, thanks!

PR #931, opened 2026-06-16
fix(autoscaling): implement instance refresh actions

c-julin:
`StartInstanceRefresh`, `DescribeInstanceRefreshes`, and `CancelInstanceRefresh` were not registered in the AutoScaling handler table, so every call failed with `InvalidAction: Unknown AutoScaling action`. Instance refreshes were never recorded on the ASG, so a poller had nothing to read back.

Changes in `autoscaling.py`:

- Add handlers for `StartInstanceRefresh`, `DescribeInstanceRefreshes`, and `CancelInstanceRefresh`, registered in `_ACTION_MAP` and following the existing query-API handler pattern.
- Record instance refreshes on the ASG object (in a new `InstanceRefreshes` list), so they ride the existing persistence / `reset` / `restore_state` plumbing. `DescribeInstanceRefreshes` returns them most-recent-first and supports filtering by `InstanceRefreshIds`.
- Since ministack runs no real instances, a refresh has nothing to roll, so it is recorded as `Successful` (`PercentageComplete: 100`) immediately. `CancelInstanceRefresh` returns `ActiveInstanceRefreshNotFound` once nothing is in progress, matching real AWS.

Added `test_start_and_describe_instance_refresh` (start -> describe round-trip, including id filtering), `test_start_instance_refresh_unknown_asg_fails`, and `test_cancel_instance_refresh_when_none_active`. All existing AutoScaling tests still pass.

Ran into this using ministack as the AWS endpoint for a Terraform setup: when an `aws_autoscaling_group`'s launch template changes and an `instance_refresh` block is present, the provider calls `StartInstanceRefresh` and then polls `DescribeInstanceRefreshes` until the refresh reaches a terminal state. Without these actions the apply failed outright; without the immediate `Successful` status the provider's wait would never complete.


Nahuel990:
Thanks @c-julin !

PR #933, opened 2026-06-16
fix(s3): GetBucketOwnershipControls 404s after delete, default otherwise

c-julin:
`GetBucketOwnershipControls` returned a default `BucketOwnerEnforced` block (HTTP 200) whenever no controls were stored — including immediately after `DeleteBucketOwnershipControls`. So during `terraform destroy` the delete was accepted but the resource still appeared to exist, and Terraform looped on its delete waiter until it failed:

```
Error: waiting for S3 Bucket Ownership Controls (<bucket>) delete: found resource
```

The fix can't simply 404 on "no stored controls", because real S3 distinguishes two cases: a bucket that was never configured reports the default Object Ownership (`BucketOwnerEnforced`), whereas a bucket whose controls were explicitly deleted returns `404 OwnershipControlsNotFoundError`.

Changes in `s3.py`:

- `_delete_bucket_ownership_controls` records a deleted tombstone on the bucket; `_put_bucket_ownership_controls` clears it (so re-creating works).
- `_get_bucket_ownership_controls` now returns: the stored controls if present; `404 OwnershipControlsNotFoundError` if explicitly deleted; otherwise the default `BucketOwnerEnforced` block as before. Only the post-delete case changes, which is what `DeletePublicAccessBlock` already does via `404 NoSuchPublicAccessBlockConfiguration`.

Extended `test_s3_ownership_controls` to assert all three states: never-configured returns the default, put-then-get returns the stored value, and delete-then-get returns `404 OwnershipControlsNotFoundError`. Targeted ownership/public-access tests pass; unrelated full-suite failures (Lambda event-notification delivery, s3control host routing, POST-object) reproduce identically on `main` and are artifacts of running the suite against a single from-source process rather than the full image.

Ran into this using ministack as the AWS endpoint for a Terraform setup that creates `aws_s3_bucket_ownership_controls` on a couple of buckets: `terraform destroy` could not converge because the GET kept reporting the controls as present after the delete.


Nahuel990 (approved):

Nahuel990:
Thanks @c-julin !

PR #934, opened 2026-06-16
test(persistence): restore cold-imported module to fix cross-module desync flake

kurok:
A flaky test that fails depending on pytest-xdist file→worker distribution under `--dist=loadfile`.

## Symptom
`tests/test_ecs.py::test_ecs_run_task_injects_secrets_manager_secrets` intermittently fails with `KeyError: 'SECRET_VAL'` and warnings:
```
WARNING ecs:ecs.py:1023 ECS: could not resolve secret arn:aws:secretsmanager:...:secret:ecs-secret-plain-... for env var SECRET_VAL
```

## Root cause
`test_module_cold_import_with_typical_snapshot_does_not_log_restore_failure` pops a service module from `sys.modules` and re-imports it fresh to catch the NameError-at-import regression — but never restores the original module. Since the test is parametrized over every persisted service:

1. Cold-importing `ecs` (alphabetically first) creates a fresh `ecs` whose `from ministack.services import secretsmanager` binds the then-current `secretsmanager`.
2. Cold-importing `secretsmanager` later swaps that module out, leaving the leaked `ecs` pointing at a **stale** `secretsmanager`.
3. A later test that creates a secret via the live `secretsmanager` and resolves it through `ecs.resolve_secret_string` reads the stale module → secret not found → `KeyError`.

The bug only surfaces when `--dist=loadfile` schedules `test_persistence.py` and `test_ecs.py` onto the same worker with the cold-imports running first — which is why `main` is green by scheduling luck rather than correctness.

## Fix
Restore the original module object in **both** `sys.modules` and the `ministack.services` package attribute (the latter is what `from ministack.services import <mod>` actually reads) once the cold-import assertion has run.

## Verification
Reproduced locally:
```
pytest \
  "tests/test_persistence.py::...[ecs-ecs]" \
  "tests/test_persistence.py::...[secretsmanager-secretsmanager]" \
  "tests/test_ecs.py::test_ecs_run_task_injects_secrets_manager_secrets"
# before: 1 failed, 2 passed   after: 3 passed
```
Full `tests/test_persistence.py` (133 params) passes, and `test_persistence.py` + the ECS secrets test together pass (134).

### DIFF
--- a/tests/test_persistence.py
+++ b/tests/test_persistence.py
@@ -674,22 +674,43 @@ def test_module_cold_import_with_typical_snapshot_does_not_log_restore_failure(
     # importlib.reload() won't work because it merges into the
     # existing namespace; the late-declared symbol stays bound from
     # the prior import.
+    import ministack.services as _services_pkg
+
     full_name = f"ministack.services.{mod_name}"
+    # The cold-import swaps a brand-new module object into BOTH sys.modules and
+    # the `ministack.services` package attribute. Other already-imported modules
+    # that did `from ministack.services import <mod>` keep a reference to the
+    # ORIGINAL object, so we must restore it afterwards. Otherwise the fresh
+    # module (with empty, reset state) leaks into later tests on the same xdist
+    # worker and desyncs cross-module references — e.g. cold-importing `ecs`
+    # then `secretsmanager` leaves the fresh `ecs` pointing at a stale
+    # `secretsmanager`, so ECS RunTask can no longer resolve Secrets Manager
+    # secrets created via the live module. Both the sys.modules entry and the
+    # package attribute must be restored: `from ministack.services import <mod>`
+    # reads the package attribute, not sys.modules directly.
+    original_mod = sys.modules.get(full_name)
     sys.modules.pop(full_name, None)
 
-    caplog.clear()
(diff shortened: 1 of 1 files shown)


kurok:
Hi @Nahuel990, could you please take a look at this PR and review it for merge when you have a moment? Thanks in advance!

jgrumboe:
@kurok Would it also help (or be better) to add this test to the serial tests? Then it's executed not in parallel with others.

kurok:
Thanks @jgrumboe. `serial` wouldn't fix this one — it's a state leak, not a parallel-pressure issue.

The `serial` set is for tests sensitive to concurrent load (cold-start timeouts, account-global state). Here the cold-import test pops the module from `sys.modules` and the `ministack.services` package attribute and never restores it, leaking a stale module into whatever runs next *in the same process*. Since the serial phase also runs in one process, the leak would still corrupt later tests there — serial would just reshuffle scheduling and hide it, which is the "green by luck" state this PR is removing.

The `finally`-restore makes the test hermetic, so it's correct under any scheduling without adding to the slower serial phase. Glad to also mark it `serial` if you'd prefer, but it shouldn't be needed.

Nahuel990 (approved):

PR #924, opened 2026-06-15
fix(glue): GetUserDefinedFunctions treats Pattern as RE2 regex, not glob

yonatoasis:
## Summary

The Glue stub in `ministack/services/glue.py` matched the `GetUserDefinedFunctions` `Pattern` argument as a **glob** (`fnmatch`), but real AWS Glue treats `Pattern` as an **RE2 regular expression** applied to the function name.

This broke Trino: Trino's Hive/Glue connector resolves a UDF by calling `GetUserDefinedFunctions` with a regex like `trino__<name>__.*`. Against MiniStack that returned 0 matches, so Trino reported `Function 'x' not registered` on every call even though `CREATE FUNCTION` had persisted the object correctly.

This PR:

- Compiles `Pattern` as a regex (`re.compile`) and matches it (`re.search`) against `FunctionName` in `_get_user_defined_functions`. `.*` matches all; `trino__dw_clean_text__.*` matches that function.
- Raises `InvalidInputException` with a Glue-like message (`Invalid pattern syntax: error parsing regexp: ...`) when `Pattern` is not a valid regex (e.g. a bare `*`), instead of silently matching everything.
- Leaves the `DatabaseName` omitted/`*` "search all databases" behaviour and pagination untouched, and does **not** touch Create/Get/Update/Delete (already correct).
- `GetTables` keeps its own `_simple_glob_match` helper — its glob semantics are correct and unchanged.
- Also clears `_user_defined_functions` in `reset()`, which the original UDF PR (#817) omitted — UDF state leaked across resets.

Fixes #923.

## Behaviour parity

Reproduced against MiniStack from this branch with a UDF named `trino__dw_clean_text__abc`:

| Pattern                       | Real AWS Glue          | This branch |
|-------------------------------|------------------------|-------------|
| `.*`                          | matches (1)            | 1 ✅        |
| `trino__dw_clean_text__.*`    | matches (1)            | 1 ✅        |
| `*`                           | `InvalidInputException`| `InvalidInputException` ✅ |

Invalid-pattern response message: `Invalid pattern syntax: error parsing regexp: nothing to repeat at position 0` (RE2's inner text differs from Go's, but the `Invalid pattern syntax: error parsing regexp:` prefix matches Glue).

## Test plan

Verified locally against MiniStack running on a fresh `.venv` (Python 3.13), endpoint on a private port.

- **`pytest tests/test_glue.py`** — 56 passed.
[...]

PR #936, opened 2026-06-16
fix(glue): GetUserDefinedFunctions accepts java.util.regex \Q...\E patterns

### LINKED ISSUE #935: Glue stub: GetUserDefinedFunctions rejects java.util.regex \Q...\E patterns (Trino UDF resolution fails with "bad escape \Q")
## Summary

MiniStack's Glue stub rejects the `Pattern` that Trino's Hive/Glue connector sends to `GetUserDefinedFunctions`, so Trino cannot resolve any UDF (CREATE works, every call fails).

PR #924 correctly changed `GetUserDefinedFunctions` to treat `Pattern` as a regular expression (RE2) instead of a glob. But Trino's connector resolves a UDF by sending a **java.util.regex** pattern that wraps the function name with `\Q...\E` (java.util.regex literal-quote), e.g. `trino__\Qdw_clean_text\E__.*`. Real AWS Glue compiles `Pattern` with `java.util.regex` and accepts it.

Python's `re` (like Go's RE2) does not support `\Q...\E`, so MiniStack 1.3.64 returns:

```
InvalidInputException: Invalid pattern syntax: error parsing regexp: bad escape \Q at position 7
```

**Affected versions:** 1.3.64 (the regex behaviour was introduced by #924).

## Reproduction (raw AWS Glue, against MiniStack)

```bash
aws glue create-database --database-input '{"Name":"pq"}'
aws glue create-user-defined-function --database-name pq \
  --function-input '{"FunctionName":"trino__dw_clean_text__abc","ClassName":"TrinoFunction","OwnerType":"USER","ResourceUris":[{"ResourceType":"FILE","Uri":"x"}]}'

# Fails today with: Invalid pattern syntax: error parsing regexp: bad escape \Q at position 7
aws glue get-user-defined-functions --database-name pq --pattern 'trino__\Qdw_clean_text\E__.*'

# Still works (plain regex)
aws glue get-user-defined-functions --database-name pq --pattern '.*'
```

End-to-end, every Trino UDF call fails:

```sql
CREATE OR REPLACE FUNCTION hive.<db>.aggregate_key(...) LANGUAGE PYTHON ...;
SELECT hive.<db>.aggregate_key(ARRAY['"a b!"']);
-- errors with: bad escape \Q
```

## Root cause

In `_get_user_defined_functions` (`ministack/services/glue.py`), `Pattern` is compiled directly with `re.compile(pattern)`. Python's `re` engine has no `\Q...\E` literal-quote support, whereas the `java.util.regex` engine that real Glue uses does, so any pattern containing `\Q` raises `re.error: bad escape \Q`.

## Suggested fix

Before compiling with `re`, emulate `java.util.regex` literal-quote semantics for `Pattern`: replace each `\Q...\E` span with `re.escape(<literal between \Q and \E>)`, and a trailing `\Q` with no closing `\E` quotes to end-of-string (per Java). Other Java-vs-RE2 escape differences are not needed — `\Q...\E` is the only construct Trino emits.

## Impact

Any client that resolves a Glue UDF via a `\Q...\E` pattern fails — most notably Trino's Hive/Glue connector, which makes UDFs entirely unusable: `CREATE FUNCTION` succeeds but every invocation errors with `bad escape \Q`.

yonatoasis:
## Summary

PR #924 made `GetUserDefinedFunctions` match `Pattern` as a regex (RE2) instead of a glob — correct. But Trino's Hive/Glue connector resolves a UDF by sending a **java.util.regex** pattern that wraps the function name with `\Q...\E` (java.util.regex literal-quote), e.g. `trino__\Qdw_clean_text\E__.*`. Real AWS Glue compiles `Pattern` with `java.util.regex` and accepts it. Python's `re` (like Go's RE2) does not support `\Q...\E`, so MiniStack 1.3.64 returned:

```
InvalidInputException: Invalid pattern syntax: error parsing regexp: bad escape \Q at position 7
```

→ Trino couldn't resolve any UDF (CREATE worked, every call failed).

This PR adds `_translate_java_regex_quote()`, which rewrites each `\Q...\E` span as `re.escape(<literal between \Q and \E>)` before compiling, and quotes a trailing `\Q` with no closing `\E` to end-of-string (per Java semantics). The literal between the markers is matched literally — regex metacharacters inside it no longer apply. Everything outside `\Q...\E` is still compiled as a normal regex, so `.*` and friends are unaffected, and a genuinely invalid pattern (e.g. a bare `*`) still raises `InvalidInputException`. `\Q...\E` is the only Java-vs-RE2 construct Trino emits, so no other escape differences are handled.

Fixes #935.

## Test plan

All steps verified locally against MiniStack running on port 4566 from this branch (Python 3.13, fresh `.venv`).

* **`pytest tests/test_glue.py`** — 57 passed (56 pre-existing + 1 new). The new `test_glue_get_user_defined_functions_pattern_java_quote` covers:
  - the exact pattern Trino emits (`trino__\Qdw_clean_text\E__.*`) resolving the function instead of raising `bad escape \Q`
  - `.*` still matching everything (regex semantics unaffected)
  - the quoted span being matched literally — `^\Qa.b\E$` matches only `a.b`, not `axb`
  - a trailing `\Q` with no closing `\E` quoting to end-of-string

* **AWS CLI round-trip** against the running emulator, reproducing the issue's acceptance steps:
  - `get-user-defined-functions --pattern 'trino__\Qdw_clean_text\E__.*'` → returns `trino__dw_clean_text__abc` (before this PR: `Invalid pattern syntax: error parsing regexp: bad escape \Q at position 7`)
  - `get-user-defined-functions --pattern '.*'` → still matches
  - `get-user-defined-functions --pattern '*'` → still `InvalidInputException` (`nothing to repeat`)

* **Persistence/`reset()` untouched.** The change is confined to compiling `Pattern`; the `_user_defined_functions` store and its `reset()`/persistence wiring (added in #924) are unchanged, so a coordinator restart still preserves created UDFs.

---

## Also in this PR: fix a flaky DynamoDB-stream ESM test (unrelated to Glue)

CI's **serial global-state set** was failing on `test_lambda_dynamodb_stream_esm_latest_processes_first_record` ("LATEST ESM skipped the first record on an initially-empty table"), independent of the Glue change. Root cause: the `LATEST` starting position for a DynamoDB-stream event source mapping was initialized **lazily on the poller's first tick**. Under the serial set's load, a slow poll cycle misses the test's 2s window, so the first poll lands *after* the first record is written — the position anchors at `len == 1` and the record is skipped permanently (the position never rewinds).

Fix: anchor the position **eagerly in `CreateEventSourceMapping`** (`_init_stream_position`), so `LATEST` is independent of poll-thread timing. The poller's lazy init is kept as a fallback for ESMs created out-of-band (e.g. restored from persisted state).

Verified locally against a fresh MiniStack: the eager init runs end-to-end (poller reads the subscription-time anchor), the target test passes repeatedly on cold starts and within the full `-m serial` set, and the rest of `tests/test_lambda.py` is unchanged (the only other failure locally is `test_lambda_docker_flags_applied_to_run_kwargs`, a pre-existing Docker-unavailable env issue that also fails without this change).


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #941, opened 2026-06-18
Validate CloudFront tag ARNs

Areson:
## Summary
- Add parser-backed validation for CloudFront TagResource, UntagResource, and ListTagsForResource.
- Validate global CloudFront ARN shape for distributions, functions, and key-value stores without request-Region matching.
- Reject malformed, wrong-service, wrong-account, and missing-resource ARNs before reading or mutating tags.

## Validation
- `uv run --extra dev pytest tests/test_cloudfront.py -q` (54 passed)

### DIFF
--- a/ministack/services/cloudfront.py
+++ b/ministack/services/cloudfront.py
@@ -30,6 +30,7 @@ from xml.etree.ElementTree import Element, SubElement, tostring
 
 from defusedxml.ElementTree import fromstring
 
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks for this @Areson I'll be modifying 2 codes in OpenSearch and CloudFront NoSuchResource -> EntityNotFound nand ValidationException/BaseException

PR #946, opened 2026-06-18
fix(cfn): ApiGatewayV2 CFN provisioner honours ms-custom-id tag (fixes #400)

### LINKED ISSUE #400: Feature Request: Custom/predictable IDs for API Gateway resources (LocalStack ls-custom-id parity)
### Feature summary

Add a mechanism to pin MiniStack-generated API Gateway IDs to caller-supplied values (tag-based, mirroring LocalStack's [custom IDs for API Gateway resources via tags](https://docs.localstack.cloud/aws/services/apigateway/#custom-ids-for-api-gateway-resources-via-tags)).

### Resources in scope

- **`aws_apigatewayv2_api`** (HTTP + WebSocket) — the api-id appears in every data-plane URL (`{apiId}.execute-api.localhost`). Today that id is random-per-apply, so any client that hardcodes the URL has to be re-pointed after every `terraform apply`.
- **`aws_apigateway_rest_api`** (v1) — same problem.

Matches LocalStack's documented scope for this feature (API Gateway only).

### Proposed UX

Recognize a tag (e.g. `ms-custom-id=<value>`) on `CreateApi` / `CreateRestApi`. When present, MiniStack uses the tag value as the resource ID instead of generating a random one. Example in Terraform:

```hcl
resource "aws_apigatewayv2_api" "example" {
  name          = "example"
  protocol_type = "HTTP"
  tags = {
    ms-custom-id = "example"  # → URL becomes http://example.execute-api.localhost:4566
  }
}
```

Error behavior: if the requested id is already in use, fail the Create call with a clear error (rather than silently falling back to a random id) so misconfigs surface fast.

### Use case

Stable, predictable API Gateway IDs matter whenever the id leaks into a URL that something else hardcodes:

- **Frontend apps in local dev** bake API URLs into an `.env` at build time. A rotating api-id means regenerating env and rebuilding the app after every `terraform apply`.
- **External tools that forward to a local URL** (webhook listeners, browser dev proxies, curl-based smoke tests) have to be re-pointed after each apply.
- **Documentation / onboarding snippets** go stale if the URL they show is only valid for the session that produced the doc.

Today the standard workaround is a reverse-proxy sidecar that reads the rotating ids out of terraform outputs at container start and rewrites `Host` headers — working but extra moving parts. Combined with a path-based data-plane (filing separately), custom IDs make the proxy unnecessary.

### LocalStack reference

https://docs.localstack.cloud/aws/services/apigateway/#custom-ids-for-api-gateway-resources-via-tags — LocalStack uses an `ls-custom-id` tag on API Gateway resources. Existing `ls-custom-id` users would transition transparently if MiniStack honored either `ls-custom-id` or a `ms-custom-id` alias.

- [ ] Yes, I'd like to implement this and open a PR
- [x] I can help review/test an implementation

hiddengearz:
## Summary

- The `_apigw_v2_api_create` CloudFormation provisioner always generated a random UUID for the API id, ignoring the `ms-custom-id` tag even when present in the template's `Tags` field
- The direct `CreateApi` API path already called `_resolve_custom_api_id()` correctly (`apigateway.py:1163`); the v1 REST API CFN provisioner also handles it via `_create_rest_api()`. Only the v2 CFN provisioner was missing the call
- Fix: call `_apigw_v2._resolve_custom_api_id(props.get("Tags", {}), _apigw_v2._apis)` before falling back to `new_uuid()[:8]` — mirrors the existing v1 pattern (one-line change at `provisioners.py:3527`)
- Adds `test_cfn_apigwv2_ms_custom_id`: deploys a CFN stack with `Tags: {ms-custom-id: cfn-pinned-api}` and asserts both the physical resource id and the live `ApiId` equal `cfn-pinned-api`

## Test plan

- [x] New test passes: `pytest tests/test_cfn.py::test_cfn_apigwv2_ms_custom_id -v` — 1 passed in 0.21s
- [x] Verified end-to-end against a running Ministack instance — `mcp-api`, `iac-private-api`, and `iac-public-api` all resolve to deterministic IDs on fresh deploys


Nahuel990 (approved):

Nahuel990:
Thanks @hiddengearz

PR #947, opened 2026-06-18
perf(lambda): store code_zip bytes as content-addressed blob files

mattwang44:
## Summary

`lambda_svc.get_state` currently base64-encodes every function's `code_zip` (and every per-version `code_zip`) inline into `lambda.json`. For deployments with many or large function zips, the resulting state file grows large and the warm-boot path has to `json.load` the whole thing in a single shot, peaking at roughly 2× the encoded size during decode.

Concrete workload that motivated this: 26 functions, ~30 MB zip each (all-deps-inlined Lambda packaging), produces a ~1 GB `lambda.json`. Restart on a 2 Gi-limited pod OOMs within ~30 seconds during the `base64.b64decode` walk.

## Change

`code_zip` bytes are written to `${STATE_DIR}/lambda-blobs/{sha256}.zip` on save. `lambda.json` keeps only a `{"code_blob_ref": "<sha256>"}` marker per function and per version.

Properties:

- **Content-addressed.** Two functions with byte-identical zips share one file on disk. Comes up naturally with shared-layer Lambda packaging.
- **Orphan sweep at save time.** `get_state` collects the set of currently referenced sha256s and removes any blob file not in that set. Without this, repeated `UpdateFunctionCode` would leak files across restarts.
- **In-memory shape unchanged.** `_functions[name]["code_zip"]` is still `bytes` after restore. Invoke / update / publish-version / delete paths are untouched.
- **Atomic blob writes.** `tempfile + os.replace`, never leaves partial files.
- **Backward compatible restore.** Pre-existing `lambda.json` files that store `code_zip` as inline base64 still load. No migration step required when upgrading in place.
- **Missing blob = degrade, not crash.** If a blob file is missing at restore (corrupted volume, partial mount), the affected function's `code_zip` is set to `None` with an error log; the rest of the stack still starts.

## Measurement

Apple M1 Pro / 10 cores / 32 GB / APFS SSD / macOS 25.5. Each row is the mean of 2 runs.

| Workload                | Metric              | main (base64 inline) | this PR (blob) |   Δ      |
| ----------------------- | ------------------- | -------------------- | -------------- | -------- |
| **26 × 5 MB**           | `lambda.json`       | 130 MB               | 36 KB          | −99.97%  |
|                         | restore (ms)        | 577                  | 261            | −55%     |
|                         | save (ms)           | 597                  | 160            | −73%     |
|                         | cold invoke p50 (ms)| 95                   | 95             | flat     |
|                         | cold invoke max (ms)| 440                  | 137            | −69%     |
|                         | warm invoke p50 (ms)| 54                   | 55             | flat     |
| **100 × 2 MB**          | `lambda.json`       | 200 MB               | 137 KB         | −99.93%  |
|                         | restore (ms)        | 640                  | 326            | −49%     |
|                         | save (ms)           | 720                  | 226            | −69%     |
|                         | cold invoke p50 (ms)| 97                   | 98             | flat     |
| **1 × 200 MB** (via S3) | `lambda.json`       | 279 MB               | 1.7 KB         | −99.999% |
|                         | restore (ms)        | 742                  | 388            | −48%     |
|                         | save (ms)           | 1327                 | 436            | −67%     |
|                         | cold invoke (ms)    | 880                  | 306            | −65%     |
|                         | warm invoke (ms)    | 53                   | 53             | flat     |

Warm invoke is unchanged at every shape because restore still hands `code_zip` back into `_functions._data` as bytes — the worker pool reads from memory, never disk.

Save / restore / cold-invoke all drop because the inline path pays `base64.b64encode/b64decode` plus a single-shot `json.dump/load` over a state file that scales with total code bytes. The blob path replaces that with content-addressed file I/O; the JSON file itself becomes trivially small. Cold-invoke gains come from removing the ~2× memory peak during decode, which lowers GC pressure during the warm-up window.

## Trade-offs not measured

These are situations where blob storage could plausibly cost more than the inline shape; I didn't have a clean way to reproduce them on the test box and they may be worth confirming on the maintainer side:

- **Slow or networked persistent volume** (NFS, gp2-class EBS, in-cluster PV). Each `open(blob_file)` has a round-trip cost; the inline shape amortizes that into one big sequential read. APFS / NVMe makes this invisible. The 100 × 2 MB row above shows the blob path stays cheaper even at 100 file opens, but it doesn't speak to network-storage latency.
- **Cold filesystem page cache** (first boot on a freshly-started pod). The inline shape benefits from a single sequential read; the blob shape reads many small files. Steady-state numbers above reflect a warm cache.

## Tests

`tests/test_lambda.py` gains seven cases next to `test_lambda_durable_persistence_round_trip`, following the same in-process `get_state`/`restore_state` pattern with `try/finally` snapshot + per-test `CODE_BLOB_DIR` redirect:

- `test_lambda_code_zip_round_trip_through_blob_storage`
- `test_lambda_get_state_replaces_code_zip_with_blob_ref`
- `test_lambda_per_version_code_zip_also_externalized`
- `test_lambda_identical_code_dedups_to_single_blob`
- `test_lambda_legacy_inline_base64_persistence_still_loads`
- `test_lambda_missing_blob_degrades_without_aborting_restore`
- `test_lambda_get_state_prunes_orphan_blobs`

Full `tests/test_lambda.py` suite passes on this branch (201 cases) against a running MiniStack.


mattwang44:
**Update:** CI passed after retrying via force-push (5th attempt landed on a clean run). The intermediate failures were spread across multiple unrelated tests, each transient:

| Attempt | Failing test                                                                       | Mode              |
| ------- | ---------------------------------------------------------------------------------- | ----------------- |
| 1       | `test_ecs.py::test_ecs_run_task_injects_secrets_manager_secrets`                   | `KeyError: 'SECRET_VAL'` |
| 2       | `test_stepfunctions.py::test_sfn_mock_config_jsonata_assign_applied`               | `TimeoutError`    |
| 3       | `test_ecs.py::test_ecs_run_task_injects_secrets_manager_secrets`                   | `KeyError: 'SECRET_VAL'` |
| 4       | `test_stepfunctions.py::...jsonata_assign_applied` + `test_ecs.py::...secrets_manager` | both              |
| 5       | (none)                                                                             | ✅ passed         |

None of those tests are touched by this PR (only `tests/test_lambda.py` and `tests/test_ministack.py::test_ministack_persist_lambda_roundtrip` are). I ran them locally to confirm:

- `test_lambda_dynamodb_stream_esm_latest_processes_first_record` (which also failed once during the retries): 10/10 pass isolated on both `main` and this branch.
- Full `tests/test_lambda.py` parallel suite (`-n auto -m "not serial"`): 3/3 clean runs on both branches (199/199 here, 192/192 on main).
- Recent CI on `main` itself shows the same flake pattern (e.g. run 27587732317 with `test_sfn_mock_config_jsonata_assign_applied` timeout on 2026-06-16) — so this is pre-existing, runner-resource-dependent.

Architecturally the explanation lines up: this PR's code only runs in `get_state` / `restore_state`, both gated by `PERSIST_STATE=1`. CI uses the default `PERSIST_STATE=0`, so the new helpers are dormant in every test.

Leaving this here for posterity in case the same flake patterns show up on future PRs.

Nahuel990:
This is brilliant @mattwang44 ! I run a few times and works amazing

PR #956, opened 2026-06-19
feat(iam): seed AWSXRayDaemonWriteAccess + AWSXrayReadOnlyAccess + AWSLambdaRole

mattwang44:
## What

Pre-seed three more AWS-managed policies with their canonical documents in `ministack/services/iam.py::_seed_aws_managed_policies`:

- `AWSXRayDaemonWriteAccess`
- `AWSXrayReadOnlyAccess`
- `AWSLambdaRole`

Documents mirrored from the [AWS Managed Policy Reference](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/), matching the convention this file already uses for the 24 pre-seeded policies.

## Why

`AWSXRayDaemonWriteAccess` is the most actively-referenced AWS-managed policy by terraform-aws-modules/lambda — every module call with `attach_tracing_policy = true` does:

```hcl
data "aws_iam_policy" "tracing" {
  count = var.attach_tracing_policy ? 1 : 0
  arn   = "arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess"
}
```

So any Terraform stack that turns on Lambda tracing (very common — both upstream module defaults and most production patterns) does this lookup. Without the seed, the data source 404s and the operator either has to set `MINISTACK_AUTOCREATE_AWS_MANAGED=1` (and accept the permissive wildcard fallback that doesn't match real AWS behaviour) or invent a downstream workaround.

`AWSXrayReadOnlyAccess` and `AWSLambdaRole` are cheap companions — same shape, frequently bundled.

## Tests

Three new tests in `tests/test_iam.py`, mirroring the existing
`test_iam_seeded_amazon_eks_cluster_policy` shape:

- Verify the policy is gettable
- Verify a sample real action key is present (e.g. `xray:PutTraceSegments`)
- Verify the wildcard `*` is NOT present (proves the canonical document is in use, not the autoviv fallback)

```
tests/test_iam.py::test_iam_seeded_amazon_eks_cluster_policy PASSED
tests/test_iam.py::test_iam_seeded_aws_xray_daemon_write_access PASSED
tests/test_iam.py::test_iam_seeded_aws_xray_readonly PASSED
tests/test_iam.py::test_iam_seeded_aws_lambda_role PASSED
```

## Notes

Submitting as draft to flag for review before promotion. Happy to add more policies in a follow-up PR if maintainers want to expand the seed catalog further — the criterion I used here was "what does the most popular upstream Terraform module (terraform-aws-modules/lambda) reference today".

Filed as a contributor noticing this gap while debugging a downstream sandbox that targets Ministack via the upstream Lambda module.


Nahuel990 (approved):

Nahuel990:
Thanks @mattwang44 !

PR #960, opened 2026-06-21
fix(sns): accept non-string Message values in SNS's Publish

noynoy83:
## Why
Step Functions' `arn:aws:states:::sns:publish` integration does not always pass `Parameters.Message` as a plain string; inline object values (e.g. in YAML or ASL) arrive as structured data. 
MiniStack's SNS `Publish` only accepted strings, so publish failed when a dict was passed.

## What
- JSON-serialize dict `Message` values in SNS `Publish` before delivery
- Add SFN integration test for object payload publish via SQS subscription

## Validation
- [x] `pytest tests/test_stepfunctions.py`
  - `tests/test_stepfunctions.py::test_sfn_integration_sns_publish_structured_payload PASSED`
  - `138 passed`


Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
Thanks @noynoy83 !

Issue #897, opened 2026-06-12
Docs: IaC page does not mention AWS SAM or its limitations


ankitaabad:
The [IaC documentation](https://ministack.org/docs/iac) covers Terraform, CDK, Pulumi, and CloudFormation — but there is no mention of AWS SAM.

Since MiniStack's CloudFormation engine does not process the `AWS::Serverless-2016-10-31` transform, SAM templates with resources like `AWS::Serverless::Function` fail with `Unsupported resource type`. This is a common stumbling block for users coming from AWS, where the SAM transform is handled server-side automatically.

## Suggestion

Add a SAM section to the IaC page. Many users trying SAM will hit the unsupported resource error and may abandon MiniStack thinking it simply doesn't work with SAM — when in fact there is a straightforward workaround.

### Workaround: Convert SAM to CloudFormation before deploying

Given a SAM template at `template.yaml`:

1. Run `sam build` — this compiles Lambda code and outputs a SAM template at `.aws-sam/build/template.yaml` where `CodeUri` still points to local build artifacts.

2. Run `sam package` on `.aws-sam/build/template.yaml` — this uploads the compiled code to S3 and outputs `.aws-sam/build/packaged.yaml`, a SAM template where `CodeUri` now points to S3 locations.

3. Run `sam validate --debug` on `.aws-sam/build/packaged.yaml` — this outputs the translated CloudFormation template (with `AWS::Lambda::Function`, `AWS::IAM::Role`, etc. instead of `AWS::Serverless::Function`) which preserves the S3 code references.

4. Deploy the extracted CloudFormation template with `aws cloudformation deploy` — this creates the stack using raw CloudFormation resource types that MiniStack supports.

It is important to run `sam validate --debug` on the packaged template (`.aws-sam/build/packaged.yaml`), not the build template (`.aws-sam/build/template.yaml`). The packaged template contains S3 references; the build template still points to local file paths which CloudFormation cannot access.

Documenting this workflow would help users understand that SAM is usable with MiniStack, just with an extra conversion step.

ankitaabad:
I found that aws cloudformation deploy doesn't update Lambda code in MiniStack, but aws cloudformation update-stack does. The root cause appears to be that change sets report Changes: [] (empty) even when the S3 key in AWS::Lambda::Function Code property has changed. Since deploy uses change sets internally, it sees "no changes" and reports success without updating anything. If you could look into this as well, that would be helpful.

Nahuel990:
[...]

PR #964, opened 2026-06-22
docs: clarify SAM IaC limitations

dashitongzhi:
## Summary
- Document MiniStack's current AWS SAM / CloudFormation transform limitation
- Explain why `AWS::Serverless::*` resources fail without pre-translation
- Add local workflow guidance for building, packaging, translating, and deploying SAM templates

References #897

## Validation
- Ran `git diff --check` for markdown whitespace issues


Nahuel990:
@dashitongzhi thanks for this but we added it in our docs https://ministack.org/docs/iac#sam and also in our MCP https://github.com/ministackorg/ministack-mcp. Since this is something we don't support I'd just add a liner in the ReadMe pointing those sources rather than the full explanation

Nahuel990 (approved):

PR #966, opened 2026-06-22
fix(eks): resolve DescribeCluster endpoint for in-cluster-network callers

b-rajesh:
The failure-fallback paths in cluster creation and OIDC-config restart hardcoded https://localhost:{port}, which is reachable from the ministack host but not from inside the nested k3s container. An in-cluster controller (e.g. Karpenter) reading cluster.endpoint could not reach the kube-apiserver and needed a --cluster-endpoint=https://kubernetes.default.svc override.

Endpoint resolution is now a single helper, _resolve_cluster_endpoint, used by both the create and OIDC-restart background paths on success and on failure. It prefers the k3s container's IP on the ministack Docker network (https://{ip}:6443) and falls back to the host-published endpoint only when no container IP is reachable. The host-side fallback now honours MINISTACK_HOST (matching the OIDC issuer URL) so remote-host deployments report a reachable endpoint instead of a hardcoded localhost.

Restored clusters (persistence) normalize their endpoint to the stable host form instead of carrying over a dead container IP, keeping the ACTIVE-cluster shape consistent for aws eks update-kubeconfig and Terraform.

The failure path skips the readiness probe (probe=False) so it returns immediately instead of blocking the background thread on a port the failed container will never open.


b-rajesh:
Hi, 
The failure seems to be on rds file changes.  would rerunning the job help ?

Nahuel990:
@b-rajesh  the restore normalization looks good (resetting to the host form is right)

But `_resolve_cluster_endpoint` prefers `https://{container_ip}:6443`, and `DescribeCluster` returns that to every caller, so again I don't folllow what is the use case for this. A docker-internal IP isn't routable from the host, so `aws eks update-kubeconfig` + `kubectl` from the host gets an unreachable endpoint. We already publish the port (`ports={"6443/tcp": port}`), so `https://{MINISTACK_HOST}:{port}` works from the host and from same-network containers. It covers everything the container IP does...plus the host

The except path `cluster["endpoint"] = f"https://localhost:{port}"` with no `_wait_for_port`, so there is no block to fix (or can you point where?) `probe=False` only avoids the refactor adding one

keep the restore fix, default to `https://{MINISTACK_HOST}:{port}` on all paths, make the container IP an explicit opt-in if a container-only deployment needs it, review the comments that are in the code which does not reflect what the code is actually doing (and gets confusing to review)

b-rajesh:
Thank you @Nahuel990  
You are right, worked on it to simplify .. 

**Container IP / who can reach it.** Agreed. `DescribeCluster` returns one endpoint to every caller, and a docker-internal IP isn't host-routable, so it broke the common `aws eks update-kubeconfig` + kubectl path. The published
  `https://{MINISTACK_HOST}:{port}` is a superset (host + same-network containers that can route to `MINISTACK_HOST`), so that's now the default on **every** path — create, OIDC restart, and restore. Container IP is an explicit opt-in
  (`MINISTACK_EKS_ENDPOINT_MODE=container-ip`) for the same-network-container-only case.

  **The `probe` / except path.** Correct — the original except path was an instant `https://localhost:{port}` with no `_wait_for_port`, so there was nothing to fix; `probe=False` only undid a wait my own refactor introduced. Dropped the
  probe machinery entirely, and `_wait_for_port` is now unused so it's removed.

  **Comments.** Fixed it -->  Rewritten to describe what the code does (host-published default; container-ip opt-in, not host-routable) instead of the old container-IP-preference rationale.

  Kept the restore normalization. Net **−73 lines** (simpler than before). Tests reworked accordingly; `ruff` clean and the full `tests/test_eks.py` suite (parallel + serial, incl. the Docker k3s create/restart/restore) is green.

jgrumboe:
@b-rajesh Please resolve the conflicts so that CI tests can be rerun. Thank you

Nahuel990:
@b-rajesh the host-default change and the restore fix are good the one thing left is MINISTACK_EKS_ENDPOINT_MODE variable... It is unnecessary and has no equivalent in AWS, the host-published endpoint https://{MINISTACK_HOST}:{port} is already reachable from the host and from same-network containers, so the container-ip mode doesn't cover any case the default doesn't... please next time try to open an Issue first stating what are you aiming to do to avoid the back and forth over PRs

b-rajesh:
thanks for the thorough reviews @Nahuel990  , and sorry for the back-and-forth. I should have opened issue describing the use case and let you weigh in on scope *before* writing code, exactly as you suggested.

Having said , cost of your time your reviews made the emulator's boundaries concrete for me (including the other 2 PR's): it's for validating config / RBAC / API-call shape locally, not running real infra; it shouldn't carry maintained reference catalogs (instance types) that drift; and in-cluster controller config like IMDS endpoint belongs in the user's PodSpec, not the emulator. That's a much sharper picture than I started with.

Going forward I'll lead with an issue and only open a PR once there's appetite. Appreciate the patience getting #966 to something mergeable.

Nahuel990:
This looks good now! Thanks!

Nahuel990 (approved):

PR #967, opened 2026-06-23
feat(eks): inject IMDS endpoint into k3s containers for in-cluster AWS SDKs

b-rajesh:
In-cluster AWS SDKs (Karpenter, AWS Load Balancer Controller, EBS CSI driver) now reach MiniStack's IMDS handler on the gateway instead of the unreachable EC2 link-local `169.254.169.254` — eliminating the
  `AWS_EC2_METADATA_DISABLED=true` + dummy-credentials workaround.
  
  ## Problem
  
  The IMDS link-local address `169.254.169.254` is hypervisor-provided on real EC2 and absent in plain Docker networking, so an AWS SDK running inside the nested k3s container hangs on IMDS and eventually fails. The documented workaround
  was to disable IMDS and supply dummy creds + region.

  ## Changes
  
  - The k3s container now receives `AWS_EC2_METADATA_SERVICE_ENDPOINT=http://<host>:<gateway-port>` (and `AWS_REGION` for the cluster's region) so in-cluster SDKs hit MiniStack's IMDS handler directly.
  - `<host>` is MiniStack's own container IP when k3s shares its Docker network, otherwise `host.docker.internal` — now also resolvable on plain Linux Docker via an `extra_hosts` `host-gateway` mapping.
  - `region` is resolved at the request boundary and passed into the background start/restart threads; the new `_resolve_ministack_host` helper is docker-based and contextvar-free, so the threads never read request-scoped state.

  ## Tests
  
  4 new deterministic tests (no Docker / no boto3): env injected, env omitted when unresolved (back-compat), host → container IP on shared network, host → `host.docker.internal` fallback. `ruff` clean; `pytest tests/test_eks.py` green
  (parallel + serial).

  ## Notes

  `AWS_EC2_METADATA_SERVICE_ENDPOINT` + `AWS_REGION` are standard AWS SDK env vars, not MiniStack-specific knobs. The iptables/DNAT path (for SDKs that ignore the env var) is intentionally not implemented — AWS SDK for Go v1.27+/v2, boto3,
  and AWS SDK for Java v2 all honour it.

  Note: independent of #966 (both branch off main) — if both merge, a trivial rebase in _bg_start/_bg_restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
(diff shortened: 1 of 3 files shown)


Nahuel990:
@b-rajesh thanks for the work here, but we can't merge this... the env var lands on the k3s server container, but pod env comes from the PodSpec, so the controllers running as pods (Karpenter, ALB, EBS-CSI) never actually pick it up. And even with creds, those controllers call `RunInstances`/`ELB`/`EBS`, which are control-plane stubs in MiniStack, so nothing real boots and the flow dead-ends regardless. I'd suggest you open an issue with your real use case because it seems you are testing a different scenario on your end

b-rajesh:
You're right — thanks for catching this, I'll close it.

  The mechanism is wrong: I set the env on the k3s **server container**, but pods get their env from the PodSpec, not the node/container environment — so Karpenter / ALB controller / EBS-CSI (all pods) never see
  `AWS_EC2_METADATA_SERVICE_ENDPOINT`. Container env ≠ pod env; I should have validated where the var actually lands before building this. The right place for it is the user's Helm values / PodSpec, outside MiniStack.

  And as you note, even past that the controllers call `RunInstances`/ELB/EBS which don't boot anything real, so the flow dead-ends at the scope ceiling regardless.

  Closing this. If a concrete scenario comes up that genuinely needs emulator-side support, I'll open an issue first per your suggestion.

PR #969, opened 2026-06-23
fix(ec2): real DescribeInstanceTypes data + DescribeInstanceTypeOfferings

b-rajesh:
## Summary

`DescribeInstanceTypes` now returns real vCPU / memory / architecture for the common current-generation families, and the previously-missing `DescribeInstanceTypeOfferings` is implemented — so any consumer that filters or scores instances by capacity (IaC tools, the AWS CLI, cluster autoscalers such as Karpenter) gets accurate data.

## Problem

`DescribeInstanceTypes` synthesized values from a name heuristic — `vcpus = 2/4/8` by substring, `mem = 1024/2048/4096` — so it reported, e.g., `m5.large` as **8 vCPU / 4096 MiB** (real: 2 vCPU / 8192 MiB), always advertised `x86_64` (so Graviton `m6g`/`c6g`/`r6g` came back as the wrong architecture), and covered only ~12 types. A consumer's capacity filter silently scores a wrong-shaped type as zero-capacity (or the wrong arch) and drops it with no error — a wrong value is worse than a missing one. `DescribeInstanceTypeOfferings` was unimplemented.

## Changes

- New `ministack/services/_ec2_instance_types.py` — a curated table of published AWS specs (vCPU, memory, architecture, network) for general purpose (m5, m5n, m6i, m6a, m6g), compute (c5, c5n, c6i, c6a, c6g), memory (r5, r6i, r6g), and burstable (t3, t3a). `cores`/`threadsPerCore` are derived from architecture (x86_64 = SMT-2, Graviton = 1).
- `DescribeInstanceTypes` renders from the table (existing response shape preserved; supports `InstanceType` lists and the `processor-info.supported-architecture` filter). Uncurated types fall back to a generic shape rather than erroring, so callers that probe arbitrary types are unaffected.
- `DescribeInstanceTypeOfferings` returns the instance-type × location cross product (AZs by default; honours `LocationType` and `instance-type`/`location` filters), with locations aligned to `DescribeAvailabilityZones`.

## Tests

- Direct, in-process unit tests of the data module: a self-consistency check across all ~104 entries (memory = vCPUs × family ratio, architecture per family, sanity), `cores_and_threads` (both branches), and representative spec locks — 100% statement coverage of the new file.
- Integration tests: accurate specs (m5.large, c6i.xlarge, t3.medium, r6g.large/arm64), architecture filter, offerings × AZs, offerings by region, lenient unknown fallback.
- `ruff` clean; `pytest tests/test_ec2.py` green.

## Notes

vCPU / memory / architecture are the published AWS specifications (stable, documented); network strings follow AWS's per-size tiers and are approximate at the top end. The module header documents how to refresh against `aws ec2 describe-instance-types`.


Nahuel990:
@b-rajesh can you explain what is your use case for having this in an emulator?

b-rajesh:
Hi @Nahuel990 

Use case: **locally validating AWS-coupled controllers and IaC that read instance-type shapes** — primarily cluster autoscalers (Karpenter, Cluster Autoscaler) and Terraform's `aws_ec2_instance_type`/`aws_ec2_instance_type_offerings` data sources.

  Why it needs to be *accurate*, not just present: these consumers call `DescribeInstanceTypes` on every reconcile and **filter/score candidates by `VCpuInfo.DefaultVCpus`, `MemoryInfo.SizeInMiB`, and `ProcessorInfo.SupportedArchitectures`**. The previous handler synthesized those from a name heuristic — `m5.large` came back as 8 vCPU / 4096 MiB (real: 2 / 8192), and *every* family reported `x86_64`, so Graviton (`m6g`/`c6g`/`r6g`)
  looked like the wrong architecture. The effect on a consumer is silent: it scores the mis-shaped type as zero-capacity (or wrong-arch) and drops it, with no error — "controller is running, no nodes ever launch, nothing in the logs." A
  wrong value is worse than a missing one. `DescribeInstanceTypeOfferings` was also unimplemented, so the AZ-availability lookup those tools do had nothing to hit.

  This also completes an enablement the repo already advertises — the EC2 row notes CreateFleet "unblocks Karpenter / Cluster Autoscaler local validation," but the candidate-type catalog is queried *before* `CreateFleet`; with a wrong
  catalog the controller never reaches the fleet call. So a user iterating on a NodePool (instance-type requirements, `arch: arm64`, resource limits) can now confirm locally that the expected types are selected, instead of being misled.

  Scope kept deliberately small: a curated table of the common current-generation families (m/c/r/t + Graviton) with **real published AWS specs** — not all ~850 types. Uncurated types fall back to a generic shape (lenient, so arbitrary
  probes don't break), GPU/metal are out, and the module header documents how to refresh against `aws ec2 describe-instance-types`.

Hope this helps ?

Nahuel990:
@b-rajesh not really. You shouldn't have in a IaC a describe-instance command... you describe-instance to explore resources. Doing so in an emulator you return a hardcoed set which is not accurate, AWS adds ec2 instances on a monthly basis, keeping up the list implies reviewing this permanently for a list we don't return. If you have your tf (or whichever IaC you use) you should have something like

resource "aws_instance" "something" {
  ami           = "ami-1234567890abcdef"
  instance_type = "t3.micro"}
  
  with the proper instance already selected. We support that.

b-rajesh:
Fair — closing it.

You're right on both counts: IaC declares the type directly (`instance_type = "t3.micro"`), which you already support via RunInstances — it doesn't call DescribeInstanceTypes; and a hardcoded catalog is a permanent maintenance/accuracy liability (AWS adds types continuously, any static list drifts) for a return that doesn't unblock a real local flow.

And to be straight about my original driver (Karpenter): even with accurate types it then calls RunInstances/CreateFleet → stubs → nothing boots, so it dead-ends at the same scope ceiling regardless. Doesn't clear the bar. Closing.

PR #975, opened 2026-06-24
fix(sns): deliver to lambda subscribers asynchronously

ryan-bennett:
## Problem

SNS fanout invoked a `lambda`-protocol subscriber **synchronously** inside `Publish`:

```python
elif protocol == "lambda":
    _deliver_to_lambda(endpoint, envelope, ...)   # blocks Publish
```

So a slow (or hung) subscriber Lambda blocked the `Publish` call — and its upstream caller. This is especially painful when a Step Functions task publishes a notification mid-execution: a slow notification subscriber stalls the whole state machine.

AWS delivers SNS → Lambda **asynchronously**: `Publish` returns as soon as the notification is accepted, the subscriber runs independently, and failures are handled out-of-band (retries / DLQ) rather than surfacing to the publisher. The `http(s)` path already does this on a background thread; only `lambda` was synchronous.

This surfaced migrating a CDK app from LocalStack to MiniStack: a Step Functions error-handler task published to an SNS topic whose Lambda subscribers were slow, and the synchronous fanout blocked the task until it timed out.

## Fix

Deliver to `lambda` subscribers on a daemon background thread, mirroring the existing `http(s)` path, so `Publish` returns immediately.

## Test

Adds `test_sns_to_lambda_delivery_is_async`: subscribes a deliberately slow Lambda (sleeps, then signals receipt out-of-band via SQS) and asserts `Publish` returns well before the subscriber completes, while the subscriber is still delivered to (the SQS signal arrives).

- On `main` (before the fix) `Publish` blocks on the subscriber and the timing assertion fails.
- With the fix it passes.

## Checklist
- [x] Tests added and passing (`pytest tests/test_sns.py -k delivery_is_async`)
- [x] Linting passes (`ruff check ministack/ tests/`)

(New-service checklist items don't apply — this is a bug fix to an existing service.)


ethan-dyas438:
Just wanted to bump this PR, since I ran into this issue locally and am using a local patch to override the `sns.py` file.

Nahuel990:
Thanks for this @ryan-bennett please be careful where you place Changelog entries

Nahuel990 (approved):

PR #981, opened 2026-06-24
feat(eks/lambda): instance-scoped container cleanup for concurrent MiniStack instances

mpaquettePax8:
## Summary

Make MiniStack's container cleanup **instance-scoped** via an optional `MINISTACK_INSTANCE_ID` environment variable, so that **multiple MiniStack instances can run concurrently on the same Docker host** without evicting each other's containers.

## The problem

MiniStack reaps containers by their `ministack=<service>` label in two places:

- `_stop_docker_containers()` at boot (`app.py`) — reaps `ministack=rds|ecs|elasticache|eks|lambda`
- `_stop_all_k3s()` (`services/eks.py`)

Both filter **only** on the `ministack=<service>` label, which is global across the host. That's correct for the common single-instance case, but it means that **starting a second MiniStack instance tears down the first one's containers** — the boot-time cleanup of instance B matches and removes instance A's k3s clusters, RDS, Lambda containers, etc.

This makes it impossible to run more than one isolated MiniStack at a time on a shared host — e.g. several feature branches / worktrees side by side, a CI matrix with parallel jobs on one runner, or multiple independent test environments.

## The change

Introduce an **opt-in** `MINISTACK_INSTANCE_ID` env var:

- When set, every container MiniStack spawns (k3s, Lambda) gets an extra label `ministack_instance=<id>` in addition to the existing `ministack=<service>` label.
- The boot-time and teardown cleanup paths add `ministack_instance=<id>` to their label filter, so each instance only reaps the containers **it** spawned.
- When **unset**, behaviour is byte-for-byte identical to today (global cleanup) — fully backward compatible, no migration needed.

A small `_instance_labels()` helper centralizes reading the env var and producing the label dict, kept next to the other module-level docker helpers in each service.

Combined with the existing per-instance port knobs (e.g. `EKS_BASE_PORT` for the k3s API port, distinct published ports), this is enough to run N fully isolated MiniStack instances on one host: each gets a unique `MINISTACK_INSTANCE_ID` and unique ports, and they no longer collide at boot.

## Files

- `ministack/app.py` — `_stop_docker_containers()`: scope boot cleanup to the instance when the env is set.
- `ministack/services/eks.py` — label spawned k3s containers; scope `_stop_all_k3s()`.
- `ministack/services/lambda_svc.py` — label spawned Lambda containers.

30 insertions, 4 deletions across 3 files. No public API or default-behaviour changes.

## Reproduce / verify

```bash
# Instance A — its own gateway + k3s API port + instance id
docker run -d --name ministack-a -p 4566:4566 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -e MINISTACK_INSTANCE_ID=a -e EKS_BASE_PORT=16443 \
  ministackorg/ministack
# create a cluster in A: aws --endpoint http://localhost:4566 eks create-cluster ...

# Instance B — distinct gateway port, k3s API port, instance id
docker run -d --name ministack-b -p 4577:4566 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -e MINISTACK_INSTANCE_ID=b -e EKS_BASE_PORT=26443 \
  ministackorg/ministack
# create a cluster in B as well

# Booting B must NOT remove A's k3s container; each is scoped to its instance:
docker ps --filter label=ministack=eks         --format '{{.Names}}'   # both A's and B's
docker ps --filter label=ministack_instance=a   --format '{{.Names}}'   # only A's
docker ps --filter label=ministack_instance=b   --format '{{.Names}}'   # only B's
```

With the env **unset**, both instances share the global label space and the second boot reaps the first's containers — the behaviour this PR fixes.

## Testing

- **Single-instance (env unset):** boot cleanup still reaps all `ministack=*` containers as before — confirmed behaviour is unchanged.
- **Two instances, distinct `MINISTACK_INSTANCE_ID` + ports:** brought both up concurrently, each creating its own k3s cluster + workloads; confirmed the second instance's boot does **not** evict the first, that each spawned container carries the correct `ministack_instance` label, and that both stayed healthy across repeated start/stop cycles.
- **Unit tests** (`tests/test_instance_scoped_cleanup.py`, no Docker required) cover all three change sites with the env set vs. unset: the `_instance_labels()` helper, the `_k3s_run_kwargs` spawn-label builder, and `_stop_docker_containers` (the boot cleanup filter is asserted via a fake docker client). 6 tests, all passing.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -1844,10 +1844,16 @@ def _stop_docker_containers():
         client = docker.from_env()
     except Exception:
         return
+    # Instance-scoped cleanup: when MINISTACK_INSTANCE_ID is set, only reap
+    # containers spawned by THIS instance (tagged ministack_instance=<id>), so
+    # multiple MiniStack instances can share one Docker host without evicting
+    # each other's k3s/RDS/etc. at boot. Unset → original global behaviour.
+    _instance = os.environ.get("MINISTACK_INSTANCE_ID")
     for label in ("ministack=rds", "ministack=ecs", "ministack=elasticache", "ministack=eks", "ministack=lambda"):
         try:
+            _flt = [label] + ([f"ministack_instance={_instance}"] if _instance else [])
             # all=True so exited-but-not-removed orphans get cleaned at boot.
-            for c in client.containers.list(all=True, filters={"label": label}):
+            for c in client.containers.list(all=True, filters={"label": _flt}):
                 try:
                     c.stop(timeout=5)
                     c.remove(v=True)
--- a/ministack/services/eks.py
+++ b/ministack/services/eks.py
@@ -233,6 +233,14 @@ def _get_docker():
         return None
 
 
+def _instance_labels():
+    """Per-instance label for spawned containers so concurrent MiniStack
+    instances don't reap each other's containers at boot (see
+    MINISTACK_INSTANCE_ID and _stop_all_k3s). Empty when the env is unset."""
+    iid = os.environ.get("MINISTACK_INSTANCE_ID")
+    return {"ministack_instance": iid} if iid else {}
+
+
 def _get_ministack_network(client):
     """Detect the Docker network MiniStack is running on."""
     if DOCKER_NETWORK:
@@ -343,7 +351,7 @@ def _k3s_run_kwargs(name: str, port: int, ms_network: str | None = None, oidc_ar
         devices=["/dev/fuse"],
         ports={"6443/tcp": port},
         name=f"ministack-eks-{name}",
-        labels={"ministack": "eks", "cluster_name": name},
+        labels={"ministack": "eks", "cluster_name": name, **_instance_labels()},
         environment={"K3S_KUBECONFIG_MODE": "644"},
         volumes={"/lib/modules": {"bind": "/lib/modules", "mode": "ro"}},
         tmpfs={"/run": "", "/var/run": "", "/tmp": ""},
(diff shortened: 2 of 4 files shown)


Nahuel990:
@mpaquettePax8 this is already covered in Ministack without adding extra variables that set us apart from AWS.

We support multi-tenant in the same way AWS does (you need to set accountID).. Different access keys map to different account IDs, and all resource state is isolated per account through AccountScopedDict. That is how AWS isolates, by account, so running multiple environments side by side means one instance with a distinct account per environment, not several instances.

If you need separate Ministack processes (different versions, a CI matrix, strict process isolation), isolate at the Docker boundary, not in Ministack code: give each instance its own Docker daemon (Docker-in-Docker per job, or a dedicated daemon per instance). Each Ministack then only sees its own containers, so the existing global ministack=* cleanup is already scoped, no labels needed. Ephemeral container-per-CI-job does the same.

MINISTACK_INSTANCE_ID adds a permanent env var plus label and cleanup surface for a host-orchestration concern that has no AWS equivalent, to solve something the account model or the Docker layer already handle. Ministack stays a single in-process emulator, with isolation kept at the account level or the infra layer below it

mpaquettePax8:
Thanks for the thoughtful response, @Nahuel990 — that makes sense. You're right that this is a host-orchestration concern with no AWS equivalent, and that the account model (`AccountScopedDict` per access key) and Docker-daemon isolation already cover it without adding a permanent label/env surface to MiniStack.

I verified both on our side: one MiniStack instance happily hosts multiple independent EKS clusters under different accounts (distinct k3s containers on auto-incremented ports, account-isolated state, no cross-eviction), and a per-instance Docker daemon (DinD) scopes the existing `ministack=*` cleanup correctly. We'll go with the account model and keep isolation at the infra layer rather than in MiniStack.

Closing this — appreciate keeping MiniStack lean. 🙏

PR #797, opened 2026-06-03
Replace all hard coded localhost with environment variable

neriyaco:
Replaces hard coded `localhost` with `MINISTACK_HOST` environment variable

Related to #795

PR #982, opened 2026-06-24
fix(cognito): allow public base URL for SAML ACS / OIDC federation callback

usatenko:
## Problem

The SAML ACS URL and the OIDC federation callback are built with a hardcoded
`http://{MINISTACK_HOST}:{GATEWAY_PORT}/...`:

```python
def _acs_url() -> str:
    return f"http://{_MINISTACK_HOST}:{_MINISTACK_PORT}/saml2/idpresponse"

def _oidc_callback_url() -> str:
    return f"http://{_MINISTACK_HOST}:{_MINISTACK_PORT}/oauth2/idpresponse"
```

`_oidc_callback_url()` is sent to the external IdP as the `redirect_uri` (and reused on the
back-channel token exchange). External IdPs (Google, OIDC, SAML) validate it as an exact
string, so when MiniStack runs behind a reverse proxy that terminates TLS on a different
host/port (e.g. nginx on `:443`), the value never round-trips:

* public IdPs reject `http://` and non-standard ports for non-localhost hosts, and
* the browser cannot reach `:4566` from outside.

Only the host is configurable today (`MINISTACK_HOST`, #797); the scheme and port are not,
so setting `MINISTACK_HOST=account.example.com` still yields
`http://account.example.com:4566/oauth2/idpresponse` — still rejected.

### Repro
1. `cognito-idp create-identity-provider --provider-type Google ...` on a pool + app client.
2. Put MiniStack behind nginx on https, proxying `/oauth2/authorize` and `/oauth2/idpresponse`
   to `127.0.0.1:4566`.
3. Hit `/oauth2/authorize?identity_provider=Google&...`.
4. The redirect to `accounts.google.com` carries
   `redirect_uri=http://localhost:4566/oauth2/idpresponse`.

## Solution

Introduce `_external_base_url()`, honored by both `_acs_url()` and `_oidc_callback_url()`,
following the precedent of `MINISTACK_COGNITO_ISSUER_BASE_URL` (#815):

- `MINISTACK_COGNITO_HOSTED_UI_URL` — externally reachable base (e.g.
  `https://account.example.com`) to override; when unset, fall back to
  `http(s)://{MINISTACK_HOST}:{GATEWAY_PORT}`, scheme honoring `USE_SSL`.

This lets the federation callback be `https://account.example.com/oauth2/idpresponse` (no
forced port, correct scheme) so it round-trips through a public IdP.

## Backward compatibility

Fully backward compatible: with no env set and `USE_SSL` unset the output is byte-for-byte the
previous `http://localhost:4566/...`. `USE_SSL=1` (already used for the listener) makes the
internal default `https://...` consistently.

Closes the gap left by #797 for federated sign-in behind an HTTPS proxy.

### DIFF
--- a/ministack/services/cognito.py
+++ b/ministack/services/cognito.py
@@ -1138,14 +1138,42 @@ def _invoke_verify_auth_challenge_trigger(pool_id: str, client_id: str, username
 # SAML / OAuth2 helpers
 # ---------------------------------------------------------------------------
 
+def _external_base_url() -> str:
+    """
+    Public base URL the OUTSIDE world uses to reach this MiniStack.
+
+    The SAML ACS URL and the OIDC federation callback are handed to an *external* IdP
+    (Google, Okta, a SAML provider, ...) which validates them as an exact string and then
+    redirects the user's browser back to them. When MiniStack runs behind a reverse proxy
+    that terminates TLS on a different host/port (e.g. nginx on :443), the internal
+    ``http://{host}:{port}`` form never round-trips:
+
+      * public IdPs reject ``http://`` (and non-standard ports) for non-localhost hosts, and
+      * the browser cannot reach ``:4566`` from outside.
+
+    Set ``MINISTACK_COGNITO_HOSTED_UI_URL`` to the externally reachable base (e.g.
+    ``https://account.example.com``) to override. When unset, falls back to
+    ``http(s)://{MINISTACK_HOST}:{GATEWAY_PORT}``, where the scheme honors ``USE_SSL`` so it
+    stays consistent with the listener.
+
+    With no env set and ``USE_SSL`` unset this returns ``http://localhost:4566`` - identical
+    to the previous hardcoded behaviour, so existing setups are unaffected.
+    """
+    base = os.environ.get("MINISTACK_COGNITO_HOSTED_UI_URL")
+    if base:
+        return base.rstrip("/")
+    scheme = "https" if os.environ.get("USE_SSL", "").strip().lower() in ("1", "true", "yes") else "http"
+    return f"{scheme}://{_MINISTACK_HOST}:{_MINISTACK_PORT}"
+
+
 def _acs_url() -> str:
     """Assertion Consumer Service URL for SAML responses."""
-    return f"http://{_MINISTACK_HOST}:{_MINISTACK_PORT}/saml2/idpresponse"
+    return f"{_external_base_url()}/saml2/idpresponse"
 
 
 def _oidc_callback_url() -> str:
     """OIDC callback URL — external OIDC IdPs redirect back here with `code`+`state`."""
-    return f"http://{_MINISTACK_HOST}:{_MINISTACK_PORT}/oauth2/idpresponse"
+    return f"{_external_base_url()}/oauth2/idpresponse"
 
 
 def _build_saml_authn_request(pool_id: str, destination: str) -> str:


Nahuel990:
@usatenko can you describe your specific use case? If this PR was generated mostly by AI, please get familiar with the codebase first.

The precedent you point to (MINISTACK_COGNITO_ISSUER_BASE_URL) was a rejected PR, so it is not something we build on. We do not add config knobs or shortcuts... the aim is to return exactly what you would get from AWS

For a reverse-proxy setup, MiniStack already derives its URLs from the incoming request (the OIDC discovery document builds its base from the Host header), which is the direction we would take here rather than a new env var

You can share the exact scenario and we will point you at the right approach

usatenko:
Sure - concrete use case:

We run MiniStack as a drop-in AWS infra for our staging app environment. The app uses a Cognito user pool with a real Google IdP for sign-in. Staging is served on a HTTPS domain (https://account.example.com) by an nginx reverse proxy that terminates TLS and forwards /oauth2/* to MiniStack on :4566.

The flow breaks at the hop where MiniStack redirects the browser to Google:

- GET /oauth2/authorize?identity_provider=Google&... → fine.
- Redirect to accounts.google.com/o/oauth2/v2/auth?...&redirect_uri=http://localhost:4566/oauth2/idpresponse.

Google rejects that redirect_uri (http + an internal host/port unreachable from the browser), so federation never completes. It comes from _oidc_callback_url(), built from _MINISTACK_HOST:_MINISTACK_PORT.

You right, host-derivation is the more AWS-faithful direction, does it make sense to rework the pull request to use this approach? I mean, is there a chance you will accept it? Or you are not going to make changes for such scenario?

usatenko:
@Nahuel990 please let me know if it makes sense to rework the pull request to use host-derivation approach
thanks

Nahuel990:
@usatenko  Sorry for the delay, I wanted to verify the AWS behaviour against a real AWS today, AWS doesnt do host-derivation either... It builds these from the user pool domain

The Cognito SAML docs give the ACS URL as exactly https://{your user pool domain}/saml2/idpresponse, and the OIDC callback follows the same pattern at `/oauth2/idpresponse` It is the configured domain, not request headers, and not an env var

Minisack already stores that domain (`CreateUserPoolDomain` sets `pool["Domain"]`) 

The two helpers are global today and don't know the pool, so they need the pool (or its `Domain`) passed in from the federation handlers... that way you whatever you do in Ministack is the same as you would do in AWS, so the rework would be to build `_acs_url()` anfd  `_oidc_callback_url()` from the configured domain (from its pool) 
Configure the domain as your user pool domain, the same as you would on AWS, and the callback becomes `https://account.example.com/oauth2/idpresponse` on its own, with no knob and no headers

usatenko:
@Nahuel990 

After testing against the current build: CreateUserPoolDomain appends .auth.{region}.amazoncognito.com: Domain=account.example.com comes back as account.example.com.auth.us-east-1.amazoncognito.com. So there's no custom-domain path today, and "just configure the domain" would point Google at a host my proxy doesn't serve.

So building the callback from pool["Domain"] needs custom-domain support too.
My plan would be: _acs_url()/_oidc_callback_url() take the pool, and if Domain has a dot use https://{Domain}/... form (AWS custom-domain behavior), else expand to {Domain}.auth.{region}.amazoncognito.com, falling back to the internal host when no domain is set.
Both federation handlers already have the pool, so authorize and the token exchange match with no extra bookkeeping. 

Sound right?

Nahuel990:
@usatenko the dot count isn't how AWS decides, and it doesn't even map to the rule, AWS keys off `CustomDomainConfig` not the string

From the `CreateUserPoolDomain` docs: "Provide this parameter only if you want to use a custom domain... Otherwise, you can omit this parameter and use a prefix domain instead." https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPoolDomain.html

A custom domain is a full FQDN like `auth.example.com` (already two dots) a prefix is a bare label like `my-prefix` that AWS expands to `my-prefix.auth.{region}.amazoncognito.com`

So the fix is to capture `CustomDomainConfig` in `CreateUserPoolDomain` and mark the domain as custom vs prefix, then build `_acs_url()` and `_oidc_callback_url()` from that:

-custom gives `https://{Domain}/...`
-prefix gives `https://{Domain}.auth.{region}.amazoncognito.com/...`
-no domain set uses the internal host

PR #987, opened 2026-06-25
feat(sqs): reject XML 1.0 forbidden characters in message bodies

yamachu:
## Problem

AWS SQS only accepts message bodies containing characters valid in XML 1.0.
The following are forbidden...

- `#x00`–`#x08` (NUL and other C0 control characters)
- `#x0B` (vertical tab), `#x0C` (form feed)
- `#x0E`–`#x1F` (remaining C0 control characters)
- `#xFFFE`, `#xFFFF` (non-characters)

Tab (`#x9`), LF (`#xA`), and CR (`#xD`) remain allowed.
See: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html

MiniStack was silently accepting messages containing these characters instead of returning `InvalidMessageContents` (400). Code that sends such payloads passes locally against MiniStack but fails against real AWS, masking a class of client-side bugs.
See: https://docs.aws.amazon.com/ja_jp/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html#API_SendMessage_Errors

## Fix

Added a forbidden-character check in `_act_send_message`, immediately after the empty-body guard. Messages containing any disallowed character now raise `InvalidMessageContents` (400), matching real AWS behavior.

The regex is compiled once at module level as a deny-list (`[\x00-\x08\x0b\x0c\x0e-\x1f\ufffe\uffff]`) to avoid per-request recompilation.

### DIFF
--- a/ministack/services/sqs.py
+++ b/ministack/services/sqs.py
@@ -28,6 +28,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 import struct
 import threading
 import time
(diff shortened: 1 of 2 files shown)


Nahuel990:
Thanks for this @yamachu  it will be available in the next version

Nahuel990 (approved):

PR #991, opened 2026-06-26
updated feature: glue schema registry

daniel-schnurpfeil:


Nahuel990:
@daniel-schnurpfeil @danschnurp I see you are only delivering the tests now, the previous PR was great (with the code) the only problem we had is that you protected your branch so we couldn't resolve the conflict at the moment of merging into main Ministack. If you bring those changes with these tests we can merge it

PR #1007, opened 2026-06-28
Negative-cache failed OIDC discovery to prevent 2hr auth outage on transient failures

### LINKED ISSUE #999: API Gateway v2 JWT authorizer: OIDC JWKS discovery failure gets negative-cached for 2hrs, causing valid tokens to fail auth
### Affected Service

API Gateway (v2)

### What happened?

_fetch_oidc_jwks_uri(issuer) performs OIDC discovery (GET {issuer}/.well-known/openid-configuration) to resolve jwks_uri, then writes the result into the module-level _oidc_config_cache with a 7200s (2hr) TTL — unconditionally, even when discovery fails. On a transient network error, jwks_uri resolves to None and that None gets cached. Every subsequent JWT validation for that issuer then falls back to an incorrect default path ({issuer}/.well-known/jwks.json), which 404s for issuers (e.g. Salesforce) that serve keys at a different path (/id/keys). The result: every authenticated request against that issuer returns a generic 401, for up to 2 hours, even though the JWT is valid and the issuer/audience config is correct. The only workaround is restarting the container, which clears the in-memory cache.

### Reproduction steps

```python
- Stand up MiniStack with an API Gateway v2 HTTP API + JWT authorizer pointed at a real OIDC issuer (e.g. https://login.salesforce.com or any IdP with /.well-known/openid-configuration).
- Trigger a transient failure during OIDC discovery for that issuer — e.g. block/drop the discovery request once (network blip, DNS hiccup, or temporarily firewall the discovery endpoint) so _fetch_oidc_jwks_uri gets a failed fetch and resolves jwks_uri = None.
- Let the request retry/succeed naturally afterward, then immediately send a request with a valid, unexpired JWT from that issuer to any route behind the JWT authorizer.
- Observe: request fails with 401, even though the token is valid and issuer/audience match.
- Retry the same valid request repeatedly for up to 2 hours — it keeps failing with 401 until either the cache TTL (7200s) expires or the container is restarted.
```

### Expected behavior

A transient OIDC discovery failure should not be cached. _fetch_oidc_jwks_uri should only write to _oidc_config_cache when jwks_uri was successfully resolved (non-None); a failed discovery attempt should just be retried fresh on the next request, instead of poisoning auth for every request to that issuer for the full TTL window.

### MiniStack version

1.3.56

### SDK / tool used

AWS CLI v2, Terraform

Pratham2703005:
_fetch_oidc_jwks_uri previously cached the result of OIDC discovery unconditionally, including failures. A single transient network blip during discovery would resolve jwks_uri = None and cache that for the full 7200s TTL, causing every subsequent JWT validation for that issuer to fall back to the wrong default path ({issuer}/.well-known/jwks.json) and 404/401 for up to 2 hours — even though the token, issuer, and audience were all valid. The only workaround was restarting the container.

this closes #999 

Changes:
- _fetch_oidc_jwks_uri now only writes the 7200s success cache when discovery actually resolves a jwks_uri.
- On failure, it writes a short 60s negative cache instead of caching nothing — this avoids re-running discovery on every single request during an outage, while letting auth recover within a minute once the issuer is reachable again.
- Updated test_apigw_oidc_discovery_failure_is_not_cached to cover both halves of the new behavior: a retry within the 60s window reuses the cached failure (no extra discovery call), and a retry after the TTL elapses re-attempts discovery and succeeds.

### DIFF
--- a/ministack/services/apigateway.py
+++ b/ministack/services/apigateway.py
@@ -516,6 +516,8 @@ async def _fetch_oidc_jwks_uri(issuer: str) -> str | None:
         jwks_uri = (json.loads(body or b"{}") or {}).get("jwks_uri")
     except (OSError, ValueError, json.JSONDecodeError):
         jwks_uri = None
+    if not isinstance(jwks_uri, str) or not jwks_uri.strip():
+        jwks_uri = None
     if jwks_uri is not None:
         # Mirror _fetch_jwks (line ~543) which hardcodes `now + 7200`; no
         # separate env knob for OIDC discovery — it shares the JWT-auth
--- a/ministack/services/apigateway.py
+++ b/ministack/services/apigateway.py
@@ -516,9 +516,13 @@ async def _fetch_oidc_jwks_uri(issuer: str) -> str | None:
         jwks_uri = (json.loads(body or b"{}") or {}).get("jwks_uri")
     except (OSError, ValueError, json.JSONDecodeError):
         jwks_uri = None
-    # Mirror _fetch_jwks (line ~543) which hardcodes `now + 7200`; no separate
-    # env knob for OIDC discovery — it shares the JWT-auth external-fetch path.
-    _oidc_config_cache[issuer] = {"jwks_uri": jwks_uri, "expiresAt": now + 7200}
+    if jwks_uri is not None:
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks @Pratham2703005 !

Issue #990, opened 2026-06-26
Organizations: add `ListParents` (OU/account parent read-back for Terraform)


b-rajesh:
### AWS Service name

Organizations

### Operations needed

- `ListParents(ChildId)` — `ChildId` is an `ou-*` id or an account id. Read the stored parent and return the single parent AWS returns:
  `{ "Parents": [ { "Id": "<parent_id>", "Type": "ROOT" | "ORGANIZATIONAL_UNIT" } ] }`
  (`Type = ROOT` when the parent id starts with `r-`, else `ORGANIZATIONAL_UNIT`). Return `ChildNotFoundException` for an unknown `ChildId`.

### Use case

A multi-OU Terraform/OpenTofu build (`aws_organizations_organizational_unit`) creates OUs via `CreateOrganizationalUnit` (already supported), but the provider's Read then calls `ListParents` to populate `parent_id` — real-AWS `DescribeOrganizationalUnit` doesn't return the parent, so the provider must `ListParents` the child, and in SDKv2 this fires on every create and refresh. `ListParents` isn't implemented, so `apply` errors on the read-back (`InvalidAction: Operation 'ListParents' not implemented`): nested OUs never get created, outputs are empty, idempotency is unreachable. This single read op unblocks the whole OU-hierarchy flow locally. (It also covers the account read-back needed if `CreateAccount`/`MoveAccount` are added later.)

### Would you like to contribute this?

- [x] Yes, I'd like to implement this service and open a PR
- [ ] I can help review/test an implementation

b-rajesh:
## Proof

Verified against a clean **MiniStack v1.3.69** (Terraform 1.5.7 + `hashicorp/aws ~> 5.0`, provider `endpoints { organizations = "http://localhost:4566" }`).

`CreateOrganizationalUnit` is already supported, so the OU is created in MiniStack — but the provider's Read immediately calls `ListParents` (real-AWS `DescribeOrganizationalUnit` doesn't return the parent), which is absent, so `apply` errors on the read-back.

```hcl
# parent_id from: aws --endpoint-url http://localhost:4566 organizations list-roots --query 'Roots[0].Id'
resource "aws_organizations_organizational_unit" "team" {
  name      = "team-ou"
  parent_id = var.root_id
}

resource "aws_organizations_organizational_unit" "team_dev" {
  name      = "team-dev-ou"
  parent_id = aws_organizations_organizational_unit.team.id
}
```

```
[...]

Issue #989, opened 2026-06-26
Organizations: account vending + SCP attachment for local Terraform `apply` (store + list, not enforce)


b-rajesh:
### AWS Service name

Organizations

### Operations needed

_Account lifecycle — for `aws_organizations_account`:_
- `CreateAccount` (return `CreateAccountStatus`; immediate `SUCCEEDED` is fine for an emulator, or `IN_PROGRESS`→poll to mirror AWS)
- `DescribeCreateAccountStatus`
- `MoveAccount` (root → target OU)
- `ListParents` — needed for the round-trip: the provider refreshes `parent_id` via `ListParents`, so without it the resource shows perpetual drift after a move (`DescribeAccount` doesn't return the parent, same as real AWS)

_SCP control-plane — store + list, **not** enforce — for `aws_organizations_policy` + `_policy_attachment`:_
- `CreatePolicy` / `UpdatePolicy` / `DeletePolicy` / `DescribePolicy` / `ListPolicies` (`SERVICE_CONTROL_POLICY`)
- `TagResource` / `UntagResource` / `ListTagsForResource` — `aws_organizations_policy` Read calls ListTagsForResource, so once CreatePolicy lands, omitting these moves the hard-stop to the post-create read-back (a Create without its paired Read op fails apply). 
- `AttachPolicy` / `DetachPolicy`
- `ListPoliciesForTarget` / `ListTargetsForPolicy`
- `EnablePolicyType` / `DisablePolicyType`
- seed AWS-managed `FullAWSAccess` attached to root/OUs (real-AWS default, so attach/detach diffs match)

**_Out of scope:**_ request-time SCP **enforcement** (per-request deny). This proves the SCP is authored + attached, not that a disallowed call is blocked.

### Use case

Validate an AWS Organizations IaC flow **locally** (Terraform/OpenTofu `apply`, not just `plan`): vend a sub-account → place it under an OU → author an SCP → attach it to the OU. Today both `aws_organizations_account` and `aws_organizations_policy(_attachment)` fail at `apply` against MiniStack because these mutation ops are missing, so the flow is `plan`-only.

[...]

PR #1009, opened 2026-06-29
feat(organizations): add account membership + OU placement

lahmish:
## What

Adds the missing account-membership and OU-placement write path to the existing
Organizations service (`ministack/services/organizations.py`). No new service, no
infra changes.

New actions:

- **Core:** `CreateAccount`, `DescribeCreateAccountStatus`, `MoveAccount`, `ListParents`, `ListChildren`
- **Invite path:** `InviteAccountToOrganization`, `AcceptHandshake`, `ListHandshakesForAccount`

## Why

The service modelled the OU tree and the read paths, but there was no way to add a
member account or place it under an OU — so `ListAccountsForParent` only ever returned
the management account at the root, and `MoveAccount` / `ListParents` / `ListChildren`
fell through to the generic `InvalidAction ... not implemented` error.

## Implementation notes

- **Placement reuses the existing internal `_ParentId` field** already stored on every
  account/OU record, so move/list operations are just reads and updates of that field —
  no new parent-tracking structure.
- **`CreateAccount` is async, like real AWS:** it returns `State=IN_PROGRESS` with no
  `AccountId`, then materialises the account at the root and flips to `State=SUCCEEDED`
  (populating `AccountId` / `CompletedTimestamp`) on the first
  `DescribeCreateAccountStatus`.
- **Invite path** lets a caller-chosen 12-digit id be pinned onto a member account
  (`CreateAccount` assigns a random id, so the invite is the only way to pin a specific
  one — and in MiniStack the access key *is* the account, so that id is meaningful).
  Because an invite spans two accounts (the master creates it, the invited account
  accepts it) and module state is namespaced per caller via `AccountScopedDict`,
  `_handshakes` is kept as a small global dict keyed by handshake id; `AcceptHandshake`
  derives the caller from the signing access key, rejects anyone but the invited account
  (`AccountOwnerNotVerifiedException`), and materialises the member into the master's org
  graph. Visibility is enforced per-caller in `ListHandshakesForAccount`.
- AWS-shaped error codes throughout: `AccountNotFoundException`,
  `SourceParentNotFoundException`, `DestinationParentNotFoundException`,
  `ChildNotFoundException`, `ParentNotFoundException`,
  `CreateAccountStatusNotFoundException`, `DuplicateAccountException`,
  `DuplicateHandshakeException`, `HandshakeNotFoundException`,
  `InvalidHandshakeTransitionException`.
- `reset()`, `get_state()` and `restore_state()` updated for the new state.

## Testing

`tests/test_organizations.py` gains 7 cases (async create + describe-status, move +
placement, move error cases, missing create-status, invite/accept pinning, wrong-caller
rejection, list-handshakes). They use isolated 12-digit accounts and never call
`/_ministack/reset`, so no `_SERIAL_TESTS` entry is needed.

```
$ pytest tests/test_organizations.py -v
13 passed
```

Also verified end-to-end against a Docker image built from this branch using the bundled
`awslocal` CLI (create-account → describe-create-account-status → move-account →
list-accounts-for-parent / list-parents / list-children).

## Checklist

- [x] New actions added to `_DISPATCH` in `organizations.py`
- [x] New module state cleared in `reset()` and handled in `get_state()` / `restore_state()`
- [x] Tests added and passing (`pytest tests/test_organizations.py -v`)
- [x] Linting passes (`ruff check` on the changed files)
- [x] `README.md` service table updated
- [x] `CHANGELOG.md` entry added


Nahuel990:
@lahmish This overlaps work that's already open and claimed: #990 (ListParents) and the account-vending half of #989 (CreateAccount/MoveAccount). Please check open issues and PRs before starting so we don't duplicate effort.

lahmish:
Thanks @Nahuel990 - my mistake, I should've checked first. I've read #989 and #990 and will close this rather than duplicate @b-rajesh 's claimed work.
I have some enforcement-layer ideas that look out of scope for #989; I'll raise those separately as their own issue to discuss. Sorry for the overlap.

PR #1011, opened 2026-06-29
fix(router): access key extraction from query params

### LINKED ISSUE #1010: S3 doesn't support presigned URLs
### Affected Service

S3

### What happened?

After creating a pre-signed URL with a different account than the default one on ministack (`MINISTACK_ACCOUNT_ID`) the server will respond with 404 since the bucket does not exist under the default account.

I was thinking about adding support for getting the account id from query params, but it's supported only in specific services (primarily S3) so we wouldn't want to support something AWS isn't.
For example, if I send HTTP request to IAM service with the credentials in the URL, real AWS wouldn't honor the credentials, but in my implementation, it will extract the account id.

Here is the implementation I currently have:

```py
def extract_access_key_id(headers: dict, query_params: dict) -> str:
    """Extract the AWS access key ID from the Authorization header or query parameters."""

    if "X-Amz-Credential" in query_params:
        credential = query_params["X-Amz-Credential"][0]
        access_key_id = credential.split("/")[0]
        return access_key_id

    auth = headers.get("authorization", "")
    if auth.startswith("AWS4-HMAC"):
        match = re.search(r"Credential=([^/]+)/", auth)
        if match:
            return match.group(1)

    if "AWSAccessKeyId" in query_params:
        return query_params["AWSAccessKeyId"][0]

    if auth.startswith("AWS "):
        match = re.search(r"AWS ([^:]+):", auth)
        if match:
            return match.group(1)

    return ""
```

I can open a PR for this, or scope the extraction from query params only to services that support this behavior.

### Reproduction steps

```python
curl $(aws s3 presign s3://my-bucket/my-file.txt --expires-in 3600)
```

### Expected behavior

File downloaded to CWD

### MiniStack version

1.3.69

### SDK / tool used

AWS CLI v2.34.6

neriyaco:
Fixes #1010


Nahuel990:
@neriyaco this PR is breaking several tests

neriyaco:
I'll check it

Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1012, opened 2026-06-29
fix(lambda): clean up orphaned tmpdir and terminate process on worker respawn

hiddengearz:
## Why

When a Lambda worker process dies between invocations (handler error, crash, or
timeout), the `Worker` class has two resource leaks:

1. **tmpdir leak**: `_spawn()` creates a new `self._tmpdir` via `tempfile.mkdtemp()`
   without removing the previous directory. The old directory (containing the
   extracted function code, handler script, and any layers) is abandoned on disk
   and never cleaned up. On a long-running instance with frequently-invoked
   Lambdas that occasionally error, this accumulates indefinitely.

2. **process leak**: When the handler returns `{"status": "error"}`, `invoke()`
   sets `self._proc = None` without calling `terminate()` first. The worker
   subprocess stays alive in its stdin readline loop consuming memory (~68 MB for
   a Node.js worker), but the Python reference is lost so it can never be
   reaped. Each subsequent invocation spawns a new process while the previous
   orphan remains running.

Both leaks compound together: every error response loses the process reference
and every subsequent respawn leaves the previous tmpdir behind.

## What

- `_spawn()`: clean up the previous `self._tmpdir` with `shutil.rmtree` before
  creating a new one. Uses `ignore_errors=True` consistent with `kill()`.
- `invoke()` error path: call `self._proc.terminate()` before setting
  `self._proc = None`. Uses `SIGTERM` (graceful) rather than `SIGKILL` since
  the process successfully returned a response and is still responsive — consistent
  with the existing `kill()` method; the timeout path correctly uses `SIGKILL`
  for unresponsive processes.

## Validation

```
$ python3 -m ruff check ministack/core/lambda_runtime.py tests/test_lambda_runtime_worker.py
All checks passed!

$ python3 -m pytest tests/test_lambda_runtime_worker.py -v
============================= test session starts ==============================
platform darwin -- Python 3.11.5, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/michaelsmith/Documents/ministack
plugins: langsmith-0.6.4, xdist-3.8.0, anyio-4.7.0, time-machine-2.14.1, dash-2.16.1, docker-3.1.1
collected 2 items

tests/test_lambda_runtime_worker.py::test_tmpdir_cleaned_before_respawn PASSED [ 50%]
tests/test_lambda_runtime_worker.py::test_process_terminated_on_error_response PASSED [100%]

============================== 2 passed in 0.07s ===============================
```

Both tests run without a Docker instance via `pytest tests/test_lambda_runtime_worker.py -q`.


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1016, opened 2026-06-30
ci: reduce buildx cache pressure and clean up PR caches

jgrumboe:
## Summary
Reduces GitHub Actions cache pressure for PR preview image builds and adds automatic cleanup of stale PR caches on close.

## Changes
- change Docker Buildx GHA cache export from `mode=max` to `mode=min` in `.github/workflows/docker-publish-on-pr.yml`
- add `.github/workflows/cleanup-pr-caches.yml` to delete caches associated with closed pull requests
- keep existing shared cache scopes (`regular-pr` / `full-pr`) while lowering cache export size

## Why
A recent PR preview build failed during Buildx cache export with:

- `error writing layer blob: failed to reserve cache`

The failure happened while exporting to the GitHub Actions cache, which suggests cache reservation/quota contention rather than an application build problem.

Using `mode=min` should reduce cache export size and lower the chance of reservation failures. Cleaning up PR-specific caches on PR close should also reduce stale cache accumulation over time.

## Expected impact
- fewer cache reservation failures in Docker preview builds
- less stale PR cache buildup
- slightly less aggressive cache reuse than `mode=max`, but better reliability

## Notes
- this does not change image contents or publish behavior
- if cache pressure continues, a follow-up could move from shared scopes to PR-specific scopes or further tune cache strategy

### DIFF
--- a/.github/workflows/cache-cleanup-on-pr-close.yml
+++ b/.github/workflows/cache-cleanup-on-pr-close.yml
@@ -1,0 +1,41 @@
+name: Cleanup PR caches
+
+on:
+  pull_request_target:
+    types: [closed]
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: read
+
+    steps:
+      - name: Delete caches for closed PR
+        env:
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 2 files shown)


jgrumboe:
To free up some cache space for now, I've cleaned up old caches (not being used the last 2 weeks).

Nahuel990 (approved):

PR #1019, opened 2026-06-30
Add region-scoped DynamoDB metadata state

Areson:
## Why
Continue the multi-region state migration by keeping DynamoDB supporting metadata and stream reads isolated per account and region.

## What
- Move DynamoDB table-supporting metadata stores onto account+region scoped storage
- Restore legacy table-name keyed metadata into the table regions, using ARN-bearing metadata values to resolve ambiguous same-name tables
- Scope DynamoDB Streams iterators and record reads to the stream ARN account+Region
- Scope EventBridge Pipes DynamoDB stream polling to the source stream ARN account+Region
- Add coverage for same-name table metadata, stream records, regional listing isolation, legacy restore, and Pipes DynamoDB stream polling

## Risk Assessment
Medium — isolated to DynamoDB supporting metadata, DynamoDB Streams reads, and the existing DynamoDB Streams to SNS Pipes polling path on the multi-region feature branch.

## References
- `uv run --extra dev ruff check ministack/services/dynamodb.py ministack/services/dynamodb_streams.py ministack/services/pipes.py tests/test_pipes.py`
- `uv run --extra dev pytest tests/test_pipes.py tests/test_dynamodb.py tests/test_dynamodb_streams.py -q`

### DIFF
--- a/ministack/services/dynamodb.py
+++ b/ministack/services/dynamodb.py
@@ -74,27 +74,27 @@ from ministack.core.persistence import PERSIST_STATE, load_state
 # validate spec.region == request region) rejected them — a self-contradiction
 # (B7). Legacy account-scoped persistence migrates via the table's TableArn.
 _tables = AccountRegionScopedDict()
-_tags = AccountScopedDict()
-_ttl_settings = AccountScopedDict()
-_pitr_settings = AccountScopedDict()
+_tags = AccountRegionScopedDict()
+_ttl_settings = AccountRegionScopedDict()
+_pitr_settings = AccountRegionScopedDict()
 # Kinesis streaming destinations — TableName -> list of
 # {"StreamArn": str, "DestinationStatus": "ACTIVE"|"DISABLED",
 #  "ApproximateCreationDateTimePrecision": "MILLISECOND"|"MICROSECOND"}.
 # ACTIVE entries get each _emit_stream_event record fanned out via
 # kinesis.put_record_internal; DISABLED entries stay on the describe
 # response (matching the ~24 h AWS retention window for readability).
-_kinesis_destinations = AccountScopedDict()
+_kinesis_destinations = AccountRegionScopedDict()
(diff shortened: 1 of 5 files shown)


kurok:
**Major: Pipes DynamoDB→SNS poller not updated for region-scoped `_stream_records`**

This PR moves `_stream_records` to `AccountRegionScopedDict` (keyed `(account, region, table)`). The Lambda ESM poller compensates — `_poll_dynamodb_streams` sets `_request_account_id`/`_request_region` per ESM from the source ARN before reading (`lambda_svc.py:5712-5714`). The **Pipes** poller was not updated: `_poll_once` runs in a daemon thread and calls `stream_records.get(table_name, [])` (`pipes.py:141`) with no context set, so `get_region()` returns the default (`us-east-1`).

Failure scenario: a Pipe with a DynamoDB stream source on a table in `us-west-2`. Writes store records under `(acct, us-west-2, table)`; the poller reads `(acct, us-east-1, table)` → always `[]` → records never fan out to SNS. Before this change (`_stream_records` was account-scoped, region-agnostic) these records were delivered. `_initial_position` (`pipes.py:212`) runs in request context and *does* see the correct region, so position bookkeeping is inconsistent with what the poller later reads.

Suggested fix — read with the source ARN's account+region (mirroring the Lambda poller), at `pipes.py:141` and `:212`:

```python
        table_name = _table_name_from_stream_arn(source_arn)
        if not table_name:
            continue

        source_spec = parse_arn(source_arn)
        records = stream_records.get_scoped(
            source_spec.account_id, source_spec.region, table_name, []
        )
```

`parse_arn`/`ArnParseError` are already imported. Setting the request context around the loop body (as Lambda does) is the more thorough option, since `_pipes` and the SNS publish path are also account-scoped.

Areson:
I've made adjustments, but I'm deferring a larger adjustments to pipes the referenced PR, which I will retarget after this merges.

PR #1020, opened 2026-06-30
Add region-scoped SSM parameter state

Areson:
## Why
Continue the multi-region state migration by making SSM parameter state behave independently per account and region while preserving legacy persisted state.

## What
- Move SSM parameters, history, and tags onto account+region scoped storage
- Parse SSM parameter ARNs for lookups, tagging, and CloudFormation parameter refs
- Preserve legacy restore behavior and slash/no-slash compatibility with focused tests

## Risk Assessment
Low to medium — isolated to SSM parameter handling and CloudFormation SSM parameter outputs on the multi-region feature branch.

## References
- `uv run --extra dev ruff check ministack/services/ssm.py ministack/services/cloudformation/provisioners.py tests/test_ssm.py`
- `uv run --extra dev pytest tests/test_ssm.py -q`

### DIFF
--- a/ministack/services/cloudformation/provisioners.py
+++ b/ministack/services/cloudformation/provisioners.py
@@ -726,8 +726,7 @@ def _ssm_create(logical_id, props, stack_name):
     ptype = props.get("Type", "String")
     value = props.get("Value", "")
     description = props.get("Description", "")
-    # ARN: no extra slash if name starts with /
-    param_arn = f"arn:aws:ssm:{get_region()}:{get_account_id()}:parameter{name}"
+    param_arn = _ssm._param_arn(name)
 
     _ssm._parameters[name] = {
         "Name": name,
--- a/ministack/services/ssm.py
+++ b/ministack/services/ssm.py
@@ -15,7 +15,9 @@ import os
 import time
 from datetime import datetime, timezone
 
+from ministack.core.arn import ArnParseError, parse_arn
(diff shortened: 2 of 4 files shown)


kurok:
Added regression coverage for the two region-scoping behaviors that this migration changes but that weren't yet tested — opened against this PR's branch so it can ride along: Areson/ministack#2

- `test_cloudformation_ssm_parameter_is_region_scoped` — a CFN `AWS::SSM::Parameter` is stored under the deploy region and returns `ParameterNotFound` from another region.
- `test_ecs_container_secret_ssm_valuefrom_is_region_scoped` — ECS `secrets[].valueFrom`: bare SSM names resolve only in the task's request region (same-name params in other regions stay isolated; a missing one fails the launch), while full ARNs resolve by the ARN's embedded region.

Both are unit-style (no server/Docker); `pytest` + `ruff` green locally.

Nahuel990:
@Areson I'm prepping the 1.4 release, which is why I'm holding these two PRs. I'll merge them directly into `main` once I finish, I think we're in good shape to release without breaking backward compatibility... I'm just running a batch of tests first

Areson:
@Nahuel990 Makes sense. Most of the work up to this point was refactoring around ARN parsing, which is mostly wrapped. There is _just_ enough multi-region around Aurora and a few services for our use-case, which is the only real multi-region in the feature branch at this point.

PR #1032, opened 2026-07-02
feat(persistence): on-demand state save, named checkpoints with hot restore, and shutdown durability fixes

amays-fwd:
## Motivation

  State kept resetting for API Gateway → Lambda setups across restarts. Two root causes in the persistence lifecycle, plus a missing capability:

  1. State is only saved on graceful ASGI lifespan shutdown — but the SIGTERM handler installed by `python -m ministack` calls `sys.exit(0)` directly, which can bypass hypercorn's lifespan
  shutdown entirely. `docker stop` could silently drop all state; crash/`SIGKILL` always did.
  2. There was no way to save or snapshot state programmatically.

  ## Added

  **`/_ministack/state/*` admin endpoints** (work regardless of `PERSIST_STATE` — that flag keeps governing only the automatic boot-restore/shutdown-save cycle):

  - `POST /_ministack/state/save` — flush all in-memory service state to `STATE_DIR` now
  - `POST /_ministack/state/checkpoint {"name": "...", "overwrite"?}` — named snapshot under `${STATE_DIR}/checkpoints/<name>/`: per-service state files, Lambda code blobs, S3 object data when
  `S3_PERSIST=1`, and a manifest. 409 on duplicate, 400 on invalid name (single path segment, traversal-safe)
  - `POST /_ministack/state/restore {"name": "..."}` — **hot restore into the running instance**: in-memory reset → file swap → per-service restore → pollers/schedulers restarted. Same
  semantics as a process restart (RUNNING Step Functions executions → `FAILED`/`States.ServiceRestart`, open API GW WebSocket connections closed). Existence is checked before anything is wiped,
  so a typo'd name 404s without touching live state
  - `GET /_ministack/state/checkpoints` / `DELETE /_ministack/state/checkpoints/<name>`

  **`PERSIST_INTERVAL`** (seconds, default `0` = off, requires `PERSIST_STATE=1`) — periodic autosave so crash/OOM/`SIGKILL`, which no shutdown hook can catch, loses at most one interval of
  state.

  ## Fixed

  - **SIGTERM/`docker stop` flush**: an `atexit` safety net saves state when `PERSIST_STATE=1` and the lifespan shutdown save didn't run. No change to the existing signal handler or lifespan
  path — the net only fires when they didn't.
  - **`/_ministack/reset` file wipe**: was gated on `PERSIST_STATE`, so files written by the explicit save endpoint with the gate off survived reset and could be resurrected into a later
  checkpoint. Now wipes `${STATE_DIR}/*.json` unconditionally, and also removes `lambda-blobs/` (previously orphaned blobs leaked after every reset). Named checkpoints survive reset by design —
  snapshot → reset → restore is the intended workflow.

  ## Design notes

  - Checkpoints **hardlink** Lambda code blobs (copy on cross-device): `lambda_svc.get_state()` prunes orphan blobs on every call, so a checkpoint must own its blob files or a later save
  corrupts it. Content-addressed sha256 names make hardlinks safe.
  - All mutating endpoints serialize behind the same reset lock as `/_ministack/reset`, which also gates request admission — snapshots are near-quiescent. A new `save_all_resilient` retries
  `get_state()` snapshots that race in-flight requests.
  - Fully additive to existing persistence: `save_state`/`load_state`/`save_all`, the shutdown save, and every service's `get_state`/`restore_state` contract are unchanged. New primitives
  (`write_state_file`/`read_state_file`) are ungated and path-parameterized; `core/checkpoints.py` is a new module.
  - Checkpoints capture state files of modules not loaded this run (prior-boot state) by copying their live `STATE_DIR` files, instead of force-importing all ~55 service modules.

  ## Tests

  `tests/test_checkpoints.py` (39 tests): checkpoint CRUD, name validation (traversal, length), blob-prune survival, unloaded-module capture, hot restore round-trips (incl. Lambda code bytes
  and SFN RUNNING→FAILED), gate independence, reset interaction (files wiped without gate, blobs cleaned, checkpoints survive), autosave loop/gating, atexit once-only semantics, and a serial
  SIGTERM integration test (`python -m ministack` + boto3 + `kill -TERM` → `sqs.json` present).

  Verified end-to-end against a live server: REST API + `AWS_PROXY` Lambda integration → checkpoint → delete both → hot restore → invoke returns the Lambda response again, with `PERSIST_STATE`
  unset.

  Checked:
  - ruff clean (one pre-existing `I001` in `services/s3.py`, untouched here)
  - `pytest tests/test_checkpoints.py tests/test_persistence.py` green (172)
  - live-server regressions green: `test_ministack`, `test_apigatewayv1`, `test_lambda`, `test_ses`, `test_account`, `test_tagging` (621 total)
  - README (env-var table + persistence section) and CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+- **State — on-demand save and named checkpoints via `/_ministack/state/*` admin endpoints** — `POST /_ministack/state/save` flushes all in-memory service state to `STATE_DIR` immediately; `POST /_ministack/state/checkpoint {"name": ...}` saves a named snapshot under `${STATE_DIR}/checkpoints/<name>/` (per-service state files, Lambda code blobs as hardlinked copies so Lambda's orphan-blob prune can't corrupt a checkpoint, and S3 object data when `S3_PERSIST=1`); `POST /_ministack/state/restore {"name": ...}` hot-restores a snapshot into the running instance without a restart (in-memory reset, file swap, per-service restore, pollers/schedulers restarted — same semantics as a process restart: RUNNING Step Functions executions become `FAILED`/`States.ServiceRestart`, open API Gateway WebSocket connections are closed); `GET /_ministack/state/checkpoints` lists and `DELETE /_ministack/state/checkpoints/<name>` removes them. The endpoints work regardless of `PERSIST_STATE` — that flag keeps governing only the automatic boot-restore/shutdown-save cycle — and named checkpoints survive `/_ministack/reset`, so snapshot → reset → restore works as a test-isolation workflow.
+- **Persistence — `PERSIST_INTERVAL` periodic autosave** — opt-in (seconds, default `0` = off, requires `PERSIST_STATE=1`): a background task saves all service state every N seconds, so a crash, OOM kill, or `SIGKILL` — which no shutdown hook can catch — loses at most one interval of state instead of everything since the last clean shutdown.
+
+### Fixed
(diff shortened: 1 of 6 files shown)


Nahuel990:
@amays-fwd can you explain "how" is docker/podman or any container loosing state when persist_state=1? atexit handlers DO run on sys.exit(0)

Issue #1038, opened 2026-07-05
DynamoDB update-item does not validate input values


ChinCJ:
### Affected Service

DynamoDB

### What happened?

Calling update-item on DynamoDB does not validate the item's values, allowing invalid values to be inserted into the database.

### Reproduction steps

```python
aws --endpoint-url http://localhost:4566 dynamodb update-item \
    --table-name invalid-state \
    --key "{\"id\": {\"S\": \"1\"}}" \
    --update-expression "SET #A = :a" \
    --expression-attribute-names "{\"#A\": \"A\"}" \
    --expression-attribute-values "{\":a\": {\"SS\": []}}"

aws --endpoint-url http://localhost:4566 dynamodb scan --table-name invalid-state
```

### Expected behavior

This should trigger the following error:
aws: [ERROR]: An error occurred (ValidationException) when calling the UpdateItem operation: 1 validation error detected: One or more parameter values were invalid: An string set  may not be empty

### MiniStack version

v1.3.70

### SDK / tool used

AWS CLI v2.35.0

Issue #1040, opened 2026-07-05
DynamoDB return-values incorrectly returned


chincharjuin:
### Affected Service

DynamoDB

### What happened?

UPDATED_NEW and UPDATED_OLD should always return Attributes if the values are not None, regardless of whether they were changed from their previous values.

### Reproduction steps

```python
aws dynamodb create-table \   
   --endpoint-url http://localhost:4566                                                       
   --table-name ministack-demo \                                                                                  
   --attribute-definitions AttributeName=pk,AttributeType=S \                                                     
   --key-schema AttributeName=pk,KeyType=HASH \                                                                   
   --billing-mode PAY_PER_REQUEST  

aws dynamodb put-item \          
   --endpoint-url http://localhost:4566                                                                                     
   --table-name ministack-demo \                                                                                  
   --item '{"pk": {"S": "k1"}, "a": {"S": "same"}}'

aws dynamodb update-item \    
   --endpoint-url http://localhost:4566                         
   --table-name ministack-demo \                                                                                  
   --key '{"pk": {"S": "k1"}}' \ 
   --update-expression "SET a = :v" \                                                                             
   --expression-attribute-values '{\":v\": {"S": "same"}}' \                                                       
   --return-values UPDATED_NEW
```

### Expected behavior

{"Attributes": {"a": {"S": "same"}} should be the response according to AWS CLI, but instead nothing is returned

### MiniStack version

v1.3.70

### SDK / tool used

AWS CLI v2.35.0

Issue #1039, opened 2026-07-05
DynamoDB invalid string set validation logic


ChinCJ:
### Affected Service

DynamoDB

### What happened?

Ministack validates the following string set value as invalid: [""]

AWS allows for empty strings to be part of a valid string set.

### Reproduction steps

```python
aws dynamodb --endpoint-url http://localhost:4566 put-item \
    --table-name ministack-testing \
    --item "{\"id\": {\"S\": \"5\"}, \"A\": {\"SS\": [\"\"]}}"
```

### Expected behavior

This should pass validation, however it hits the following error:
aws: [ERROR]: An error occurred (ValidationException) when calling the PutItem operation: One or more parameter values were invalid: An string set may not be empty

### MiniStack version

v1.3.70

### SDK / tool used

AWS CLI v2.35.0

Nahuel990:
@ChinCJ did you run this against real AWS?

chincharjuin:
Hello! Yes, the commands were all run against real AWS to ensure that its behaviour on Ministack differed.

Nahuel990:
@ChinCJ we merged the PR but haven't released them. I checked myself against AWS because this was behaving different a few weeks ago but it seems it went back as it was. We'll release v1.3.72 today with this fix (and the other 2) in it

Nahuel990:
This is released in v1.3.72 thanks!

PR #1041, opened 2026-07-05
Fixes to dynamodb validation inconsistencies

chincharjuin:
Fixes the following dynamodb validation inconsistencies:
1. Update item does not validate item values
2. String set validation does not accept empty strings
3. UPDATED_NEW and UPDATED_OLD return values not equivalent to AWS behaviour

Regression tests and minimal code changes were applied for these issues.


chincharjuin:
Fixes #1038, #1039, #1040

Nahuel990:
Thanks for this @chincharjuin

Nahuel990 (approved):

PR #1044, opened 2026-07-06
fix(ec2): implement placement group actions

c-julin:
`CreatePlacementGroup`, `DeletePlacementGroup`, and `DescribePlacementGroups` were not registered in the EC2 handler table, so every call failed with `InvalidAction: Unknown EC2 action: CreatePlacementGroup`. Placement groups were never stored, so nothing could read one back.

Changes in `ec2.py`:

- Add handlers for `CreatePlacementGroup`, `DeletePlacementGroup`, and `DescribePlacementGroups`, registered in `_ACTION_MAP` and following the existing `CreateKeyPair` / `CreateSecurityGroup` query-API handler pattern (query params in, `_xml` / `_error` responses out).
- Store placement groups in a new `_placement_groups` `AccountScopedDict`, so they ride the existing `get_state` / `restore_state` / `reset` persistence plumbing exactly like `_key_pairs`.
- Parse `TagSpecification.N` via the shared `_parse_tag_specs` helper keyed on the `pg-` group id, and register the `pg-` prefix in `_guess_resource_type`, so `CreateTags` / `DescribeTags` and `tag:`/`tag-key`/`tag-value` filters (`_resource_matches_tag_filters`) treat placement groups like every other tagged EC2 resource. `DescribePlacementGroups` also supports `GroupName.N`, `GroupId.N`, and `group-name` / `state` / `strategy` filters.
- Duplicate `GroupName` returns `InvalidPlacementGroup.Duplicate`; an unknown name on delete or a named describe returns `InvalidPlacementGroup.Unknown`, mirroring the existing `InvalidKeyPair.*` error shape. `partitionCount` is emitted only for the `partition` strategy, matching real EC2.

Added `test_ec2_placement_group_crud` (create -> describe -> delete -> describe-unknown round-trip), `test_ec2_placement_group_partition_reports_count`, `test_ec2_placement_group_duplicate`, `test_ec2_placement_group_delete_unknown`, and `test_ec2_placement_group_tag_round_trip` (create-with-tags -> describe read-back -> `tag:` filter). All existing EC2 tests still pass.

Ran into this using ministack as the AWS endpoint for a Terraform setup: `aws_placement_group` calls `CreatePlacementGroup` on apply, then `DescribePlacementGroups` (by name) to read the resource back and detect drift, and `DeletePlacementGroup` on destroy. Without the actions the apply failed outright, and without the tag read-back a second `terraform apply` never converged to "No changes".

### DIFF
--- a/ministack/services/ec2.py
+++ b/ministack/services/ec2.py
@@ -12,6 +12,7 @@ Supports:
                    AuthorizeSecurityGroupIngress, RevokeSecurityGroupIngress,
                    AuthorizeSecurityGroupEgress, RevokeSecurityGroupEgress
   Key Pairs:       CreateKeyPair, DeleteKeyPair, DescribeKeyPairs, ImportKeyPair
+  Placement Grps:  CreatePlacementGroup, DeletePlacementGroup, DescribePlacementGroups
   VPC / Subnets:   DescribeVpcs, DescribeSubnets, DescribeAvailabilityZones
                    CreateVpc, CreateDefaultVpc, DeleteVpc, CreateSubnet, DeleteSubnet
                    CreateInternetGateway, DeleteInternetGateway, DescribeInternetGateways,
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks for this!

PR #1045, opened 2026-07-06
fix(eks): configurable OIDC issuer scheme

c-julin:
`_ministack_issuer_base()` hard-coded `http://`, so `DescribeCluster` advertised an `http` OIDC issuer. terraform-provider-aws's `aws_iam_openid_connect_provider` client-side-validates that `url` is https and rejects the plan before any request reaches the API: `expected "url" to have a url with schema of: "https"`. Real EKS issuers are always https, so IRSA terraform against ministack could never apply.

Changes in `eks.py`:

- Add an `EKS_ISSUER_SCHEME` env override to `_ministack_issuer_base()` (default `http` for backward compatibility). Setting `EKS_ISSUER_SCHEME=https` makes `DescribeCluster` advertise an https issuer.
- The issuer is computed in one place: both `DescribeCluster`'s `identity.oidc.issuer` (set from `_issuer_url` at create time) and the OIDC discovery document (`_oidc_discovery`, which re-derives `_issuer_url`) flow through `_ministack_issuer_base`, so they stay self-consistent — the discovery document's `issuer` echoes the same https URL, which oidc clients that fetch it require. The discovery/JWKS endpoints are still served over the plain-http gateway, so clients that only register the issuer URL (e.g. terraform with explicit thumbprints) work regardless.
- Document `EKS_ISSUER_SCHEME` in the README Configuration table alongside the other emulator env vars.

Added `test_eks_issuer_scheme_https_is_self_consistent` (with `EKS_ISSUER_SCHEME=https`, `_issuer_url` and the `_oidc_discovery` document both advertise the same https issuer) and `test_eks_issuer_scheme_defaults_to_http`. All existing EKS tests still pass.

Ran into this using ministack as the AWS endpoint for a Terraform setup: `aws_iam_openid_connect_provider` validates the `url` argument is https before it makes any call, so pointing it at a cluster's `identity.oidc.issuer` failed at plan time until the advertised issuer could be https.

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,7 @@ end-to-end without any client config.
 | `OPENSEARCH_DASHBOARDS_BASE_PORT` | `15601` | Starting host port for per-domain Dashboards containers |
 | `OPENSEARCH_DASHBOARDS_IMAGE` | `opensearchproject/opensearch-dashboards:2.15.0` | Image used when spawning per-domain Dashboards containers |
 | `MINISTACK_OPENSEARCH_ENDPOINT` | _(unset)_ | If set (e.g. `localhost:9200`), every domain returns this endpoint instead of spawning a per-domain container — useful when you bring your own cluster |
+| `EKS_ISSUER_SCHEME` | `http` | URL scheme for the EKS cluster OIDC issuer advertised by `DescribeCluster` and the OIDC discovery document. Set `https` for IRSA terraform: `aws_iam_openid_connect_provider` client-side-validates that `url` is https (real EKS issuers always are). Discovery/JWKS are still served over the plain-http gateway |
 | `PERSIST_STATE` | `0` | Set `1` to persist service state across restarts |
 | `STATE_DIR` | `/tmp/ministack-state` | Directory for persisted state files |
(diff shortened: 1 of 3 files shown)


Nahuel990:
Good catch, the HTTP issuer breaking `aws_iam_openid_connect_provider` is a real bugbut it's not a new environment variable. AWS has no issuer-scheme configuration, and real EKS always uses HTTPS, so having `EKS_ISSUER_SCHEME` default to `http` would just preserve the incorrect behavior by default

The actual bug is that `_ministack_issuer_base` ignores `USE_SSL`. When running with `USE_SSL=1`, the gateway serves HTTPS but still advertises an HTTP issuer. The cluster endpoint already advertises HTTPS because K3s serves it; the issuer just never got wired to the gateway's scheme

We're going to reject this approach and land the fix ourselves by deriving the scheme from `tls.use_ssl_enabled()`: use `https` when `USE_SSL` is enabled and `http` when it isn't

For your Terraform setup, that means running Ministack with `USE_SSL=1`, which is the correct behavior since an HTTPS issuer should correspond to a gateway that is actually serving HTTPS

Thanks for reporting

PR #1046, opened 2026-07-06
fix(autoscaling): report mock in-service instances so capacity waiters converge

c-julin:
`CreateAutoScalingGroup` recorded the group but never materialized any instances, so `DescribeAutoScalingGroups` always returned an empty `Instances` member. terraform-provider-aws's `aws_autoscaling_group` capacity waiter polls DescribeAutoScalingGroups until `DesiredCapacity` instances are `InService`/`Healthy`, so every apply with the default `wait_for_capacity_timeout` (10m) blocked the full timeout and then failed: `waiting for Auto Scaling Group (<name>) capacity satisfied: context canceled`. Real ASGs launch instances; an emulator reporting zero forever is the parity gap.

Changes in `autoscaling.py`:

- Add `_reconcile_instances`, which resizes the group's `Instances` list to exactly `DesiredCapacity` mock entries (`InstanceId` `i-<17 hex>`, `LifecycleState` `InService`, `HealthStatus` `Healthy`, round-robin `AvailabilityZone` across the group's AZs, `ProtectedFromScaleIn` inherited from the group's `NewInstancesProtectedFromScaleIn`, echoing the group's LaunchTemplate/LaunchConfigurationName). Scale-down removes from the end, matching a Default termination policy. It is called from `CreateAutoScalingGroup`, `UpdateAutoScalingGroup`, and a new `SetDesiredCapacity` handler (registered in `_ACTION_MAP`), so the count tracks every capacity change.
- `DescribeAutoScalingGroups` now emits the instances (via a shared `_instance_member_xml` renderer) instead of the hardcoded `<Instances/>`. `DescribeAutoScalingInstances` — previously a stub returning an empty set — now returns the same instances across all groups, carrying the owning `AutoScalingGroupName` and honoring the `InstanceIds` filter.
- Instances live on the group record, so they inherit the existing persistence / `reset` / `restore_state` plumbing (the same as the `InstanceRefreshes` list). Instance refreshes are unaffected — they still complete immediately as `Successful`, and now coexist with a non-empty instance set.

Added `test_asg_reports_desired_capacity_instances` (create -> describe reports N InService/Healthy instances, also visible via DescribeAutoScalingInstances), `test_set_desired_capacity_reconciles_instances` (SetDesiredCapacity up then down reconciles the count), `test_update_asg_reconciles_instances` (UpdateAutoScalingGroup DesiredCapacity 0 -> 2), and `test_asg_instances_round_robin_azs` (AZ round-robin across two zones). All existing AutoScaling tests still pass.

Ran into this using ministack as the AWS endpoint for a Terraform setup: `aws_autoscaling_group` with the default (non-zero) `desired_capacity` and no `wait_for_capacity_timeout = "0"` override waits on DescribeAutoScalingGroups for the group to report that many healthy in-service instances. Against an emulator that reported zero, every apply hung for the full 10-minute timeout and then failed instead of converging.

### DIFF
--- a/ministack/services/autoscaling.py
+++ b/ministack/services/autoscaling.py
@@ -5,7 +5,8 @@ All in-memory, no actual instance scaling.
 
 Supports:
   ASG:       CreateAutoScalingGroup, DescribeAutoScalingGroups, UpdateAutoScalingGroup,
-             DeleteAutoScalingGroup, DescribeAutoScalingInstances, DescribeScalingActivities
+             DeleteAutoScalingGroup, SetDesiredCapacity, DescribeAutoScalingInstances,
+             DescribeScalingActivities
   Refresh:   StartInstanceRefresh, DescribeInstanceRefreshes, CancelInstanceRefresh
   LC:        CreateLaunchConfiguration, DescribeLaunchConfigurations, DeleteLaunchConfiguration
   Policies:  PutScalingPolicy, DescribePolicies, DeletePolicy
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1067, opened 2026-07-10
fix(dynamodb): validation parity on reads, deletes, batches, and transactions

ifutivic:
## What

Fixes DynamoDB validation gaps where MiniStack silently accepted requests that real AWS rejects (or returned the wrong error shape). All behaviors verified against **real AWS DynamoDB** (`eu-central-1`).

- **`GetItem` / `DeleteItem` / `BatchGetItem`** — reject empty-string/binary key values (as `PutItem` already did).
- **`BatchWriteItem`** — validate the whole request before applying; a bad member no longer leaves earlier members written.
- **`TransactWriteItems`** — validate members before applying, with AWS's two error shapes: an empty-string key value is a top-level `ValidationException`, while a wrong-typed key or update-expression type error is a `TransactionCanceledException` with a positional `ValidationError` reason (via the `_transact_cancel_response` path from #1041).
- **`Query`** — reject `BETWEEN` with inverted bounds, empty-string key operands, and an `ExclusiveStartKey` that falls outside the range predicate (all at parse time).
- **`UpdateItem`** — reject `SET` on an unresolvable path (e.g. `list_append` on a missing attribute), and `ADD`/`DELETE` with a wrong operand type or a type that mismatches the existing attribute.

## Why

We're migrating a large integration-test fleet from LocalStack to MiniStack. These are the exact behaviors where code relying on DynamoDB's strict validation (e.g. atomic batches/transactions) diverged from production. Initially validated against DynamoDB Local, which turned out not to match real AWS on the transaction error shape — so every point was re-checked against real AWS.

## Testing

- New tests in `tests/test_dynamodb.py` for every gap (including both `TransactWriteItems` shapes) plus sanity tests for the valid paths.
- Full suite passes (298 tests); `ruff check` clean.


ifutivic:
@Nahuel990 Could you please take a look? This is blocking our LocalStack -> MiniStack migration 🙏

Nahuel990:
@ifutivic I've seen you validated against DynamoDB Local and LocalStack. If you run item 3 against real DynamoDB you'll see part of it fails, the error shape for transactions is different there (we fixed that a few weeks ago validated against real DDB) I'll verify each item against real DynamoDB during July 13th and follow up

ifutivic:
> if you run item 3 against real DynamoDB you'll see part of it fails, the error shape for transactions is different there

Wow, I just assumed DynamoDB local behaves the same as the real one considering it's written by AWS.

Nahuel990:
@ifutivic We thought the same when we started building Ministack, but we soon ran into several parity issues. If you're curious, this project gives a great overview of the current state of things: https://paritysuite.org/

ifutivic:
Very sorry about this then. I should not have made such assumptions.
Let me spend some time also validating against actual DynamoDB to save you a bit of time.

ifutivic:
I re-validated all the points against real DynamoDB (test table in `eu-central-1`). You're right about the transactions, and I found one more issue.

**Two fixes I've pushed:**

1. **TransactWriteItems error shape.** Real AWS only returns a plain `ValidationException` for an empty-string key value. A wrong-typed key or an update-expression type error comes back as `TransactionCanceledException` with a `ValidationError` reason. My change returned a plain `ValidationException` for all of them. It now returns `TransactionCanceledException` for the type-mismatch cases and keeps `ValidationException` only for the empty-string key.

2. **DELETE operand-type message.** Real AWS emits `typeSet: ALLOWED_FOR_ADD_OPERAND` even for the DELETE operator. I was emitting `ALLOWED_FOR_DELETE_OPERAND`. Now matches real AWS.

Thanks for the feedback.

ifutivic:
Did the same on: https://github.com/ministackorg/ministack/pull/1068

Nahuel990:
Thanks! we'll include them in the release 1.4.2 (July 13th)

Nahuel990 (approved):

Nahuel990 (approved):

PR #1068, opened 2026-07-10
fix(dynamodb): return LastEvaluatedKey when Query/Scan results end exactly at Limit

ifutivic:
## What

`Query` and `Scan` only returned a `LastEvaluatedKey` when strictly more items remained (`len(results) > limit`). Real DynamoDB doesn't look ahead: whenever it stops *because of* the limit — including when the results end exactly at it — it returns a `LastEvaluatedKey`, and the follow-up page comes back empty with no key. Verified against DynamoDB Local (LocalStack 4.2.0):

```
# 3 items in partition, Limit=3
DynamoDB Local: Count=3, LastEvaluatedKey present  → next page: Count=0, no key
MiniStack:      Count=3, LastEvaluatedKey absent
```

Both handlers now use `>=`, with `Limit > Count` still returning no key.

## Why

SDK pagination helpers commonly implement "has more pages" as `LastEvaluatedKey != nil`. Against MiniStack they report the exact-limit page as the last one; against real AWS the same code performs one more (empty) round trip. Our DynamoDB SDK's conformance suite caught this as a behavioral divergence when migrating our integration tests from LocalStack.

## Testing

- 4 new tests in `tests/test_dynamodb.py`: exact-limit key present for Query and Scan, follow-up page empty with no key, no key when `Limit > Count`, and a Limit=1 pagination walk that visits every item exactly once and terminates.
- Full `tests/test_dynamodb.py` passes (283 tests).
- `ruff check` clean.
- Our internal 383-spec DynamoDB conformance suite: 6 previously-failing pagination specs now pass against a build of this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- **DynamoDB — `Query`/`Scan` return `LastEvaluatedKey` when results end exactly at `Limit`** — real DynamoDB doesn't look ahead: whenever it stops because of the limit it returns a `LastEvaluatedKey`, even if nothing remains, and the follow-up page comes back empty with no key. MiniStack only returned the key when strictly more items remained, so SDK pagination helpers that rely on the exact-limit page (e.g. "has more pages" checks) diverged from AWS. Verified against DynamoDB Local. Contributed by @ifutivic.
+
 ## [1.4.1] — 2026-07-09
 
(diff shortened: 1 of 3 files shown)


ifutivic:
@Nahuel990 Could you please take a look? This is blocking our LocalStack -> MiniStack migration 🙏

ifutivic:
Validated against real DynamoDB (test table in `eu-central-1`), behavior matches the PR.

With 3 items in one partition:

| limit | Query | Scan |
|-------|-------|------|
| 3 (== count) | LEK present | LEK present |
| 4 (> count) | no LEK | no LEK |

Paging with the exact-limit `LastEvaluatedKey` returns an empty follow-up page (`Count=0`, no key). So real AWS returns a key when it stops exactly at the limit, which is what this PR implements.

Nahuel990 (approved):

PR #1085, opened 2026-07-14
feat(cloudformation, apigw-v2): support AWS::ApiGatewayV2::Authorizer

### LINKED ISSUE #1084: CloudFormation: AWS::ApiGatewayV2::Authorizer resource type not supported
### Affected Service

API Gateway (v2) / CloudFormation

### What happened?

Deploying a CDK app whose `HttpApi` uses a JWT authorizer (`HttpUserPoolAuthorizer`, or any `AWS::ApiGatewayV2::Authorizer`) fails outright — the whole stack rolls back, not just that resource:

```
CREATE_FAILED | AWS::ApiGatewayV2::Authorizer | Unsupported resource type: AWS::ApiGatewayV2::Authorizer
```

`ministack/services/cloudformation/provisioners.py` has provisioners for `AWS::ApiGatewayV2::Api`, `::Stage`, `::Integration`, and `::Route`, but no `_apigw_v2_authorizer_create`/`_delete`, and no registry entry — even though the runtime already supports JWT authorizers end-to-end via the control-plane `CreateAuthorizer` API (used by Terraform's `aws_apigatewayv2_authorizer`, see #802) and enforces them correctly on request (`_validate_jwt_authorizer`/`_resolve_jwks_url`).

This forces anyone deploying via CDK/CloudFormation with a protected HTTP API route to work around it (e.g. verifying the JWT in-Lambda instead of via the authorizer), even though the exact same authorizer works fine when created through Terraform or the raw API.

### Reproduction steps

```
1. Start MiniStack.
2. cdklocal deploy a stack containing an HttpApi with a route using
   HttpUserPoolAuthorizer (or a raw AWS::ApiGatewayV2::Authorizer resource
   with AuthorizerType: JWT).
3. Deploy fails: CREATE_FAILED, Unsupported resource type:
   AWS::ApiGatewayV2::Authorizer, and the whole stack rolls back.
```

### Expected behavior

`AWS::ApiGatewayV2::Authorizer` should provision through CloudFormation the same way it already works via Terraform/the raw API — mapped onto the existing in-memory authorizer store so it's enforced identically at request time.

### MiniStack version

latest (also reproduces on 1.4.2)

### SDK / tool used

AWS CDK (`HttpUserPoolAuthorizer` / `aws-cdk-lib/aws-apigatewayv2-authorizers`), cdklocal

ryan-bennett:
## Problem

Deploying a CDK/CloudFormation stack with a JWT authorizer on an `HttpApi` (e.g. `HttpUserPoolAuthorizer`) fails outright:

```
CREATE_FAILED | AWS::ApiGatewayV2::Authorizer | Unsupported resource type: AWS::ApiGatewayV2::Authorizer
```

...and the whole stack rolls back, not just that resource. `ministack/services/cloudformation/provisioners.py` has provisioners for `AWS::ApiGatewayV2::Api`, `::Stage`, `::Integration`, and `::Route`, but never gained one for `::Authorizer`.

This is despite the fact that the same JWT authorizer already works end-to-end when created via Terraform's `aws_apigatewayv2_authorizer` or the raw `CreateAuthorizer` API (#802), and is already enforced correctly at request time by `_validate_jwt_authorizer`/`_resolve_jwks_url`. It's specifically the CloudFormation resource-type registration that's missing, not the underlying authorizer logic.

This surfaced migrating a CDK app (Cognito user pool + `HttpUserPoolAuthorizer` protecting HTTP API routes) onto MiniStack — every route needing auth had to fall back to in-Lambda JWT verification instead of the native authorizer, just to get a local deploy to succeed at all.

## Fix

Add `_apigw_v2_authorizer_create`/`_apigw_v2_authorizer_delete`, mirroring the existing `AWS::ApiGateway::Authorizer` (v1 REST API) provisioner pattern: map CFN properties onto the same in-memory `_authorizers` store the control-plane API already writes to. `jwtConfiguration` is normalized from CFN's PascalCase `Audience`/`Issuer` to the API's camelCase wire shape — the response JSON isn't deep-transformed, so passing the raw PascalCase dict through silently dropped those fields from `GetAuthorizers`.

Also fixes `AWS::ApiGatewayV2::Route`, which read `AuthorizationType` from CFN properties but silently dropped `AuthorizerId` and `AuthorizationScopes`. Without this, even after adding the Authorizer resource, a route referencing one would deploy successfully but never actually invoke it at request time (`_authorize` reads `route.get("authorizerId")`).

## Test

Adds `test_cfn_apigwv2_authorizer_jwt`: deploys an `HttpApi` + JWT `Authorizer` + `Route` referencing it via `AuthorizerId`, and asserts (via the control-plane API) that the authorizer's `AuthorizerType`/`JwtConfiguration` and the route's `AuthorizerId`/`AuthorizationType` all round-trip correctly, then verifies cleanup on stack delete.

- On `main` (before the fix), the stack deploy itself fails at the `CREATE_FAILED` step above — it never gets far enough to reach the assertions.
- With the fix, `pytest tests/test_cfn.py -k apigwv2` passes 10/10, and the full `test_cfn.py` suite passes 106/106 (no regressions).
- Also re-verified directly against a real CDK app using `HttpUserPoolAuthorizer` (the construct that originally surfaced this) — full stack deploy now succeeds with the native authorizer instead of the in-Lambda workaround.

## Checklist
- [x] Tests added and passing (`pytest tests/test_cfn.py -k apigwv2`, full `test_cfn.py` 106/106)
- [x] Linting passes (`ruff check ministack/ tests/`)

(New-service checklist items don't apply — this is a bug fix to an existing service.)

Closes #1084


ryan-bennett:
The two failing jobs (`test`, `test (parallel, 2, 3)`) both fail on the same two tests in `tests/test_appsync_events.py` — `test_websocket_publish_frame_acks_and_fans_out` (`assert None is not None`, an ack not arriving within the 3s window) and `test_publish_rejects_segment_with_bad_chars` (a raw socket `TimeoutError`). This PR doesn't touch AppSync Events at all — only `provisioners.py`, `tests/test_cfn.py`, and `CHANGELOG.md`.

Looks like parallel-worker timing flakiness rather than a real failure:
- The other parallel shards (`0, 3` and `1, 3`) running the same suite under the same load passed.
- Both tests pass immediately when run in isolation locally.
- Nothing on `main`'s recent CI history shows this failing.

I don't have rerun permission on the repo, so flagging in case a re-run is needed.

Nahuel990 (approved):

Nahuel990:
Thanks @ryan-bennett !

PR #1094, opened 2026-07-16
fix(lambda): redirect Node fd-1 writes off warm-worker stdout (#1093)

### LINKED ISSUE #1093: Node.js Lambda invocation result is misread as failure when the handler writes to stdout via a raw fd write
### Affected Service

Lambda

### What happened?

A Lambda handler that writes to stdout via a raw fd write (`fs.writeSync(1, ...)`) instead of `console.log`/`process.stdout.write` gets its invocation reported as failed (`Runtime.HandlerError` / `Unknown error`), even though it actually ran fine and returned a valid response. This isn't specific to any trigger — plain synchronous `aws lambda invoke` is enough to reproduce it.

This matters because synchronous logging libraries do exactly this — e.g. pino's `{ sync: true }` destination uses sonic-boom's `fs.writeSync(fd, ...)` internally, a common AWS-documented Lambda logging pattern (flush before freeze).

The Node.js warm worker (`ministack/core/lambda_runtime.py`) talks to its Python parent over the worker's stdout using a JSON-line protocol, and only overrides `process.stdout.write` to keep that channel clean. That doesn't catch a raw `fs.writeSync(1, ...)`, so the stray log line lands in the real stdout and gets misread by `_read_response` as the invocation result. It has no `status`/`error` keys, so it falls into the `"Unknown error"` fallback in `_execute_function_warm`.

Downstream, this breaks anything that depends on the invocation result being accurate — e.g. an SQS event source mapping with `ReportBatchItemFailures` never deletes the message, redelivering it every `VisibilityTimeout` and causing duplicate side effects, even though the handler already succeeded the first time.

### Reproduction steps

```javascript
exports.handler = async () => {
  require('fs').writeSync(1, 'hi\n');
  return { ok: true };
};
```

Deploy this as a Node.js Lambda function (PackageType: Zip, the default — not a container image) and invoke it. The invocation comes
back as `FunctionError: "Unhandled"` / `{"errorMessage": "Unknown error", "errorType": "Runtime.HandlerError"}`,
even though the handler ran fine and would have returned `{"ok": true}`.
Replacing the `fs.writeSync` line with `console.log('hi')` makes it succeed.

### Expected behavior

The invocation should be reported as successful with the handler's actual return value, regardless of whether it logs via `console.log` or a raw fd write — matching real AWS Lambda, where stdout/stderr are just forwarded to CloudWatch Logs and neither carries invocation-result data.

### MiniStack version

1.3.72

### SDK / tool used

AWS CLI 2.25, Node.js 24.x Lambda runtime

vvaide:
Fixes #1093
## Summary

Node.js warm workers communicate invoke results over stdout as JSON lines (`{"status":"ok",...}` / `{"status":"error",...}`). #373 already redirected `process.stdout.write` to stderr so `console.log` would not pollute that channel, but handlers that write directly to file descriptor 1 — notably sync loggers like pino (`fs.writeSync(1, ...)`) — bypassed that hook. Stray JSON on stdout could be parsed as the protocol response, producing false `Runtime.HandlerError` / `Unknown error` even when the handler returned successfully.

## Fix

- **`ministack/core/lambda_runtime.py` (Node worker bootstrap)** — patch `fs.writeSync` and `fs.write` so writes to fd `1` or `process.stdout.fd` are redirected to stderr, mirroring the existing `process.stdout.write` hook.
- **`ministack/core/lambda_runtime.py` (`Worker.invoke`)** — when scanning stdout for the invoke response, accept only JSON lines whose `status` is `"ok"` or `"error"`. Skip structured logs (pino JSON), HTTP-style `{"status":200}` objects, init `{"status":"ready"}` lines, malformed JSON, blank lines, and plain text until a real protocol line appears (still capped at 200 lines).

## Test plan

- [x] `test_invoke_ignores_json_logs_on_stdout` — pino-style JSON on stdout before the real `status: ok` line must not end the read loop early.
- [x] `test_invoke_ignores_raw_text_on_stdout` — plain fd-1 text is skipped; the handler result is still read.
- [x] `test_invoke_ignores_json_with_unrelated_status_key` — JSON with `"status": 200` must not be treated as a protocol response.
- [x] `test_invoke_ignores_many_log_lines_before_protocol` — a burst of structured logs must not hide the real response.
- [x] `test_invoke_skips_malformed_json_lines` — broken JSON is ignored; a valid protocol line later still succeeds.
- [x] `test_invoke_skips_empty_lines` — blank lines between noise and the protocol line are ignored.
- [x] `test_invoke_skips_ready_status_during_invoke` — init `{"status":"ready"}` must not satisfy an invocation read.
- [x] `test_invoke_still_surfaces_protocol_error` — real `status: error` still fails the invocation and terminates the worker.
- [x] `test_invoke_error_after_junk_stdout` — log noise before `status: error` must not mask the handler failure.
- [x] `test_invoke_gives_up_after_max_lines` — 200+ non-protocol lines returns an error instead of hanging.
- [x] `test_nodejs_worker_fd_write_sync_succeeds` — `fs.writeSync(1, ...)` must not break the worker protocol.
- [x] `test_nodejs_worker_fs_write_async_succeeds` — async `fs.write(1, ...)` is redirected like `writeSync`.
- [x] `test_nodejs_worker_fd_write_stdout_fd_succeeds` — writes via `process.stdout.fd` are redirected.
- [x] `test_nodejs_worker_fd_write_many_lines_succeeds` — many fd-1 writes in one invocation still yield the handler result.
- [x] `test_nodejs_worker_json_log_with_status_field_succeeds` — JSON log with `"status": 200` on fd 1 must not break invoke.
- [x] `test_nodejs_worker_fd_write_then_handler_error` — fd-1 logging before a thrown error must not mask the failure.
- [x] `test_lambda_nodejs_fd_write_sync_invoke_succeeds` — end-to-end invoke with `fs.writeSync(1)` returns 200, no `FunctionError`.
- [x] `test_lambda_nodejs_fd_write_sync_warm_reinvoke` — warm re-invoke after fd-1 logging keeps returning correct results.
- [x] `test_lambda_nodejs_pino_style_json_log_invoke_succeeds` — pino-style JSON on fd 1 succeeds; log appears in `LogResult` with `LogType=Tail`.
- [x] `test_lambda_nodejs_fd_write_sync_no_log_type` — default invoke still returns the handler payload when fd 1 is written.
- [x] `test_lambda_nodejs_fd_write_and_console_log_invoke_succeeds` — `console.log` and `fs.writeSync(1)` together both work and appear in tail logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- **Lambda — Node warm-worker invoke no longer fails on fd 1 writes** — `fs.writeSync` / `fs.write` on stdout bypassed the existing `process.stdout.write` redirect to stderr (#373), so sync loggers (e.g. pino with `{ sync: true }`) could emit on the JSON-line protocol channel and surface false `Runtime.HandlerError` / `Unknown error` responses even when the handler succeeded. The Node worker bootstrap now redirects fd 1 through `fs.writeSync` / `fs.write`, and the warm-worker stdout reader accepts only protocol lines whose `status` is `ok` or `error`. Reported by @kofuk (#1093).
+
 ## [1.4.2] — 2026-07-13
 
 ### Added
--- a/ministack/core/lambda_runtime.py
+++ b/ministack/core/lambda_runtime.py
@@ -162,10 +162,25 @@ const url = require("url");
 const Module = require("module");
 
 // Redirect stdout to stderr so stdout stays clean for JSON-line protocol
+const fs = require("fs");
 const _realStdoutWrite = process.stdout.write.bind(process.stdout);
 const _stderrWrite = process.stderr.write.bind(process.stderr);
 process.stdout.write = function(chunk, encoding, callback) {
   return _stderrWrite(chunk, encoding, callback);
+};
+const _stdoutFd = process.stdout.fd;
+const _realWriteSync = fs.writeSync.bind(fs);
+const _realWrite = fs.write.bind(fs);
+function _isStdoutFd(fd) {
+  return fd === 1 || fd === _stdoutFd;
+}
+fs.writeSync = function(fd, ...args) {
+  if (_isStdoutFd(fd)) fd = 2;
+  return _realWriteSync(fd, ...args);
+};
+fs.write = function(fd, ...args) {
+  if (_isStdoutFd(fd)) fd = 2;
+  return _realWrite(fd, ...args);
 };
 
 // Synthetic AWS SDK v3 stubs — real AWS Lambda (Node.js 18+) ships these
@@ -924,8 +939,9 @@ class Worker:
                         if response_line.startswith("{"):
                             try:
                                 response = json.loads(response_line)
-                                result_box.append(response)
-                                return
+                                if response.get("status") in ("ok", "error"):
+                                    result_box.append(response)
+                                    return
                             except json.JSONDecodeError:
                                 continue
                     result_box.append({"status": "error", "error": "No JSON response from worker after 200 lines"})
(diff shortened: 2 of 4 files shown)


Nahuel990 (approved):

Nahuel990:
@vvaide thanks!

PR #1104, opened 2026-07-17
Add Aurora shared storage parity

Areson:
## Why
Aurora cluster members currently provision independent database containers, so writer and reader endpoints can expose different data and RDS Data API state. Model cluster-owned shared storage so local Aurora behavior more closely matches AWS for application and Terraform testing.

*Note*: This is a bit of a trade off. It _is_ possible to model replicas in a region using MySQL replication, but that increases the complexity quite a bit. Unless we have a strong use case for that mocking replicas via the API and serving replicas in a single region using a single backing MySQL instance seems like a lower-maintenance route. The biggest tradeoff is that "read" replicas will technically be writeable, which means it _could_ fail to mimic actual behavior in some scenarios.

## What
- Back every Aurora cluster with one cluster-owned database container and persistent volume shared by all members
- Route writer, reader, direct SQL, and RDS Data API access to the shared cluster endpoint
- Preserve data while a cluster has zero provisioned members, restart it when a member is attached, and reserve cleanup for cluster deletion or reset
- Coordinate readiness, password rotation, persistence restore, and legacy-state migration around the cluster-owned resource
- Document the intentionally read/write local reader endpoint and add focused unit plus live Docker integration coverage

## Risk Assessment
Medium: this changes Aurora container ownership, lifecycle, credentials, and persistence behavior in the local RDS emulator. Focused RDS and Data API suites, a live Docker-backed Aurora lane, and provider acceptance coverage exercise the affected paths.

## Validation

* `uv run --extra dev ruff check ministack/services/rds.py ministack/services/rds_data.py tests/test_rds.py tests/test_rds_data.py tests/test_rds_aurora_cluster_integration.py tests/test_stepfunctions.py`
* `uv run --extra dev pytest tests/test_rds.py -q` with local MiniStack server (`116 passed, 1 skipped`)
* `uv run --extra dev pytest tests/test_rds_data.py -q` with local MiniStack server (`43 passed`)
* `uv run --extra dev pytest tests/test_stepfunctions.py -q` with local MiniStack server (`155 passed`)
* `DOCKER_NETWORK=bridge uv run --extra dev pytest tests/test_rds_aurora_cluster_integration.py -q` with local MiniStack server (`4 passed`)


Nahuel990:
@Areson The fact that the reader endpoint currently accepts writes is a parity gap because real Aurora rejects write operations on reader endpoints. This does not require a replication process, since both endpoints connect to the same db.... Instead, the reader endpoint can simply enforce a read-only session by using PostgreSQL's default_transaction_read_only or a read-only role, or a read-only session in MySQL, I'll merge this and apply that approach on top

Nahuel990 (approved):

PR #1105, opened 2026-07-17
feat(apigateway): support custom domain routing and v2 mappings

### LINKED ISSUE #1030: feat(apigateway): route registered custom domain names to their mapped API (data plane)
## Summary

API Gateway **custom domain names** and **base path / API mappings** are accepted and stored today, but they are inert — the execute-api data plane never consults them, so a request whose `Host` header is a custom domain doesn't reach the mapped API. I'd like to make custom domains functional for invocation **on the existing gateway port**, so an API can be called via a real domain instead of only the `{apiId}.execute-api.localhost/{stage}/…` form.

Explicitly **out of scope** for this proposal (see Scope below): TLS / port 443, and any DNS resolution — I'm fine hitting the domain on `:4566` over HTTP with a `/etc/hosts` entry. This is purely about host-based routing.

## Motivation / use case

We run ministack for local dev of a Terraform-provisioned microservice platform. In real AWS the services are reached through an API Gateway custom domain (`apigw.<stage>.klearly.eu`) with base-path mappings per service (`/table-pay`, `/crm`, …). Locally we can't reproduce that, so today we run a **Caddy reverse proxy** in front of ministack whose only jobs are:

1. rewrite the `Host` header to `{apiId}.execute-api.localhost`, and
2. prepend the stage to the path (`/{stage}{uri}`),

because ministack only accepts the execute-api host or a stage-as-first-segment path. That's a workaround for functionality that maps cleanly onto AWS's own custom-domain feature. If ministack resolved the `Host` against registered domains + mappings, the proxy disappears and our Terraform (`aws_api_gateway_domain_name` + `aws_api_gateway_base_path_mapping` / `aws_apigatewayv2_api_mapping`) works locally the same way it does in AWS.

## Current behavior (with references)

- Invoke is reachable only via `_EXECUTE_API_RE` (`app.py:39`) — `Host: {8-hex}.execute-api.localhost/{stage}/…` — or the path forms `/_aws/execute-api/{id}/{stage}/…` and `/restapis/{id}/{stage}/_user_request_/…` (`_parse_execute_api_url`, `app.py:1150`).
- v1 `CreateDomainName` / `CreateBasePathMapping` are CRUD-only stubs (`apigateway_v1.py:1969–2042`, stores `_domain_names` / `_base_path_mappings` at `apigateway_v1.py:109–110`). Nothing reads them for routing.
- v2 `CreateApiMapping` / `GetApiMappings` / `DeleteApiMapping` are **not implemented**.
- A request with `Host: apigw.local.klearly.eu` matches nothing and falls through to the default service (S3).

## Proposed behavior

Given a registered custom domain and mapping, e.g.:

```bash
awslocal apigateway create-domain-name --domain-name apigw.local.klearly.eu
awslocal apigateway create-base-path-mapping \
  --domain-name apigw.local.klearly.eu \
  --rest-api-id abcd1234 --stage local --base-path table-pay
```

this should invoke the API (Host on the normal gateway port, HTTP is fine):

```bash
curl -H 'Host: apigw.local.klearly.eu' http://localhost:4566/table-pay/orders
# -> invokes rest-api abcd1234, stage "local", path "/orders"
```

Semantics to match AWS: the base path (v1) / mapping key (v2) is stripped from the path, the stage comes from the mapping (not the URL), and an empty base path acts as a catch-all. Longest-matching base path wins when multiple mappings exist.

## Proposed implementation sketch

1. Add a `find_api_by_host(host)` resolver that looks up `host` in the `_domain_names` store and follows base-path / api mappings to `(api_id, stage, matched_base_path)`. This mirrors the pattern ALB already uses (`alb_module.find_lb_for_host`, called at `app.py:1270`).
2. In `_handle_execute_api_request` (`app.py:1225`), before the `_EXECUTE_API_RE` branch, try `find_api_by_host`; on a hit, strip the base path, set the stage from the mapping, and dispatch to the existing v1/v2 `handle_execute` (unchanged).
3. Implement v2 `CreateApiMapping` / `GetApiMappings` / `DeleteApiMapping` in `apigateway.py` and wire them into the same resolver.
4. Tests (following the existing execute-api tests): create a REST API + domain + base-path mapping via boto3, then invoke via raw HTTP with a `Host` header and assert the Lambda-proxy response.

No change to the existing execute-api host/path forms — this is purely additive.

## Scope boundaries

- **In scope:** host-based routing for registered custom domains; v1 base-path mappings + v2 api mappings driving the data plane; base-path strip + stage resolution; matching tests.
- **Out of scope (intentionally):** TLS / port 443 (fine to serve the domain over HTTP on the existing gateway port); any DNS server or resolution — the operator provides a `/etc/hosts` entry or wildcard record, same as today; making Route53 records functional (Route53 stays a CRUD store).

## Open questions for maintainers

1. Preferred spot for the resolver — inline in `app.py` next to the ALB lookup, or a helper exported from `apigateway_v1.py`?
2. For v2, confirm you'd want the real `CreateApiMapping` API surface here rather than folding it under the v1 domain store.
3. Any objection to custom domains being served over plain HTTP on the gateway port (no TLS), given TLS is explicitly out of scope for this change?

## Would you like to contribute this?

- [x] Yes, I'd like to implement this and open a PR

> Note: this is an enhancement to the existing API Gateway service (not a new service), so it doesn't fit the New Service / Bug Report templates — filing as a free-form enhancement.

vvaide:
## Why

API Gateway custom domains and v1 base-path mappings could be created and stored, but the data plane never consulted them. Requests carrying a registered custom domain in the `Host` header therefore missed API Gateway entirely and fell through to S3 instead of invoking the mapped API. API Gateway v2 also lacked the DomainName and ApiMapping control-plane APIs needed by boto3 and Terraform.

This forced local environments to run a separate reverse proxy that rewrote the host to `{apiId}.execute-api.localhost` and prepended the stage. Resolving registered domains and mappings on MiniStack's existing gateway port makes local invocation follow AWS semantics directly, without adding TLS or DNS behavior.

## Summary
- route API Gateway v1 and v2 custom-domain requests through base path/API mappings with longest-prefix and root fallback semantics
- add API Gateway v2 DomainName and ApiMapping CRUD, persistence, region/account isolation, lifecycle conflicts, and domain tag ARN support
- preserve execute-api request context for path-based addressing while exposing custom domains to Lambda proxy events

Out of scope: TLS/DNS, `MaxResults`/`NextToken` pagination, and CloudFormation domain resources.

## Test plan
- [x] API Gateway v1 custom-domain tests
- [x] API Gateway v2 custom-domain and tag ARN tests
- [x] Botocore contract: CreateDomainName does not accept top-level `CertificateArn` or `ApiMappingSelectionExpression`
- [x] Existing API Gateway v1/v2, ARN parser, and Lambda proxy regression suites
- [x] Ruff checks for all changed Python files

Fixes #1030


Nahuel990:
Hi @vvaide, this issue was already being worked on by the author of the PR. Please avoid starting work on issues that are already in progress unless you have a specific AWS-related use case that's been discussed first. Otherwise, it ends up duplicating effort. Please check the issue and any linked PRs before starting work. Thanks!

PR #1107, opened 2026-07-18
feat(kms): add GenerateDataKeyPair and GenerateDataKeyPairWithoutPlaintext

yl:
## Summary

Adds `GenerateDataKeyPair` and `GenerateDataKeyPairWithoutPlaintext` to the KMS emulator. It previously implemented `GenerateDataKey` / `GenerateDataKeyWithoutPlaintext` but no data-key-*pair* operation, so both actions returned `InvalidAction: Unknown action`.

- Provisions an asymmetric data key pair whose private key is wrapped under a symmetric CMK.
- `GenerateDataKeyPair` returns `PrivateKeyPlaintext`, `PrivateKeyCiphertextBlob`, `PublicKey`; the `WithoutPlaintext` variant omits `PrivateKeyPlaintext` so the private key never reaches the caller in the clear.
- `KeyPairSpec` follows the AWS enum: RSA_2048/3072/4096, ECC_NIST_P256/P384/P521, ECC_SECG_P256K1, ECC_NIST_EDWARDS25519 (SM2 excluded, as `CreateKey` doesn't implement it either).
- `EncryptionContext` binds the wrapped private key; the blob decrypts back to a PKCS#8 DER private key that pairs with the returned public key.

## Ciphertext layout change

The symmetric ciphertext layout gains a per-blob random nonce (`key_id + ctx_hash + nonce + data`) mixed into key derivation. Without it the XOR keystream repeats across every blob sharing a CMK and `EncryptionContext`; because a key pair's plaintext embeds its public modulus, the wrapped private key could be recovered from the public outputs alone. `Encrypt`, `GenerateDataKey` and `Decrypt` adopt the same layout so every wrap/unwrap path stays consistent. Blobs are in-memory and cleared on reset, so the format change is transparent within a run.

## Testing

- `pytest tests/test_kms.py` — 79 passed (adds round-trip, per-blob-nonce, and keystream-attack regression tests).
- `pytest tests/test_secretsmanager.py tests/test_ssm.py` — 85 passed, confirming the layout change stays contained in `kms.py`.
- `ruff check` — clean.

## Checklist
- [x] Tests added and passing
- [x] Service added to the README table
- [x] CHANGELOG entry
- [x] `ruff check` passes


Nahuel990 (approved):

Nahuel990:
@yl thanks for this, I will make some minor adjustments before release

Issue #1118, opened 2026-07-20
Possible issue with re-spawning Lambda docker containers


ykharko:
### Affected Service

Lambda

### What happened?

It looks like the latest version **1.4.3** doesn't replace Lambda docker containers for the "docker" executor after Lambda configuration change. Existing container(s) started before Lambda configuration change keeps running and new configuration is not used for new Lambda invocations.

Lambda version "$LATEST" is used.

What I noticed:
- Version **1.3.72** (I'm currently using): Ministack stops existing docker container(s) right after submitting "update-function-configuration" request and starts new containers with updated configuration for new invocations.
- Version **1.4.3** (tried today): old docker containers keep running and there's no clear way to replace them and start using new configuration.

It seems this behavior got changed (broken) somewhere between these 2 releases.
Is this expected or it's a regression?

If this is an expected change, could you refer to a way of re-spawning Lambda docker containers?

### Reproduction steps

```python
1. Start Ministack service.
2. Create Lambda function.
3. Invoke Lambda function. New docker container with Lambda runtime is spawned. Check the container ID / name.
4. Update the Lambda function configuration using either "aws" cli or "boto3" lib.
5. Invoke the function again. The same docker container is still running. The old configuration is used for the invocation.
```

### Expected behavior

Docker containers with the old Lambda version / configuration are stopped and new containers are started after Lambda configuration change.

### MiniStack version

1.4.3

### SDK / tool used

boto3

Nahuel990:
Hello, thanks for the report. This should be fixed in v1.4.4

PR #1119, opened 2026-07-20
Fix: stop Lambda docker workers on configuration update to force respawn

AnasBabari:
## Summary
The bug is that Lambda docker containers (workers) are not stopped/recreated when a function configuration is updated. The worker pool keyed by function ARN persists old workers, and there is no invalidation hook on update-function-configuration. We add a helper to terminate workers for a given function ARN and call it from the Lambda service config update path via the runtime module.

**Fixes:** https://github.com/ministackorg/ministack/issues/1118

---

### Changes Made
- `ministack/core/lambda_runtime.py`

### DIFF
--- a/ministack/core/lambda_runtime.py
+++ b/ministack/core/lambda_runtime.py
@@ -76,6 +76,30 @@ def _account_region_from_function_config(config: dict) -> tuple[str, str]:
 _workers: dict = {}
 _lock = threading.Lock()
 
+
+def stop_workers_for_function(function_arn: str) -> None:
+    """Terminate and remove all persistent workers for a Lambda function ARN.
+
+    Used when a function's configuration (or code) changes so that subsequent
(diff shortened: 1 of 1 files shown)


Nahuel990:
Thanks for taking this on. This issue was already fixed in the 1.4.4 release (#1118), and this change wouldn't have addressed the underlying bug.

The issue was in the Docker warm container pool (`_warm_pool` / `_pool_kill_function` in `lambda_svc.py`), where the reaper stopped matching the region-scoped key. `stop_workers_for_function` operates on a different pool (`_workers`, the local subprocess pool), and it matches on `k == function_arn`, while those keys are actually in the form `account:region:function:qualifier`. As a result, it doesn't reap anything, and it's also not part of the update path.

I'll close this since 1.4.4 already resolves the issue. Thanks again for the contribution and for taking the time to investigate it.

PR #1126, opened 2026-07-21
Add Cognito resource server support

ryan-bennett:
## Summary

Adds Cognito resource server support: `CreateResourceServer`, `UpdateResourceServer`, `DescribeResourceServer`, `DeleteResourceServer`, `ListResourceServers`, plus `AWS::Cognito::UserPoolResourceServer` CloudFormation support.

Previously every resource-server action returned `InvalidAction: Unknown Cognito IDP action`, and the CFN resource type failed the whole stack with `Unsupported resource type`. This blocks any app pool that defines custom OAuth scopes (`{identifier}/{scopeName}`) for use with API Gateway `authorizationScopes` gating on an `HttpJwtAuthorizer` — a pattern used by CDK's `OAuthScope.resourceServer()` / `UserPool.addResourceServer()` and by Terraform's `aws_cognito_resource_server`.

Resource servers are keyed by their caller-supplied `Identifier` within a pool (matching real AWS, where `Identifier` also forms the OAuth scope string), and `Ref` on the CFN resource returns that `Identifier` per the [CFN resource reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpoolresourceserver.html#aws-resource-cognito-userpoolresourceserver-return-values).

## Changes

- `ministack/services/cognito.py`: five new action handlers, `_resource_servers` added to the pool state shape (defensive `setdefault` so pools created before this change / via the CFN provisioner's own pool-dict construction still work), registered in `_dispatch_idp`.
- `ministack/services/cloudformation/provisioners.py`: `AWS::Cognito::UserPoolResourceServer` create/delete provisioner, registered in the resource-type table.
- `tests/test_cognito.py`: 6 new tests (create/describe, duplicate-identifier error, list, update, delete, not-found errors).
- `tests/test_cfn.py`: 1 new test covering the CFN resource type end-to-end (create, `Ref`, delete).
- `README.md` / `CHANGELOG.md` updated per the contribution checklist.

## Test plan

- [x] `pytest tests/test_cognito.py tests/test_cfn.py -v` — 250 passed, 2 skipped (unrelated `SAM transform` skips)
- [x] `pytest tests/ -n 4 --dist=loadfile -m "not serial"` (full suite) — no new failures; the 61 pre-existing failures are unrelated to this change (missing optional deps like `cryptography`/`docker` for KMS/ECS tests in my environment, one network-dependent EKS OIDC test)
- [x] `ruff check` clean on all changed files
[...]

PR #1127, opened 2026-07-21
Add AWS::Cognito::UserPoolGroup CloudFormation support

ryan-bennett:
## Summary

Adds CloudFormation support for `AWS::Cognito::UserPoolGroup`.

The raw `CreateGroup`/`DeleteGroup`/`GetGroup`/`ListGroups`/`ListUsersInGroup` API actions already existed, but there was no CFN resource-type mapping, so deploying a stack containing a `UserPoolGroup` resource failed the whole stack with `Unsupported resource type: AWS::Cognito::UserPoolGroup` — even though the underlying group functionality worked fine via the raw API and via Terraform's own group support presumably goes through a different path. This came up alongside #1126 (Cognito resource servers): a CDK stack using `UserPool.addGroup()` (the L2 construct behind admin/user role groups) hit this immediately after the resource-server gap was fixed.

## Changes

- `ministack/services/cloudformation/provisioners.py`: `_cognito_user_pool_group_create`/`_cognito_user_pool_group_delete`, registered in the resource-type table. `Ref` returns `GroupName` (matching the [CFN resource reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpoolgroup.html#aws-resource-cognito-userpoolgroup-return-values)); delete also removes the group from any member users' `_groups`, mirroring the existing raw `DeleteGroup` handler's cleanup.
- `tests/test_cfn.py`: one new end-to-end test — create via CFN, `Ref` resolves to `GroupName`, `admin_add_user_to_group`/`admin_list_groups_for_user` see the CFN-created group, delete removes it.

## Test plan

- [x] `pytest tests/test_cognito.py tests/test_cognito_custom_auth.py tests/test_cfn.py -v` — 281 passed, 2 skipped (unrelated SAM transform skips)
- [x] `ruff check` clean on changed files
- [x] Verified against a real consumer: the same CDK stack from #1126 also creates two `UserPool.addGroup()` groups — previously failed at this exact gap once the resource-server issue was resolved, now deploys successfully end-to-end.


Nahuel990 (approved):

Nahuel990:
Thanks @ryan-bennett !

PR #1130, opened 2026-07-21
Scope Batch state by region

Areson:
## Why
AWS Batch resources are regional, but MiniStack currently shares compute environments, queues, definitions, and jobs across regions within an account.

## What
- Scope all four Batch stores by account and region
- Preserve every legacy account while deriving each resource's region from its nested ARN
- Preserve current snapshots whose resources exist only outside the boot region
- Version the regional persistence schema so older binaries refuse incompatible snapshots
- Add end-to-end regional isolation, legacy migration, current round-trip, and rollback-safety coverage

## Risk Assessment
Low — the change is isolated to Batch state storage and retains legacy snapshot compatibility.

## Validation

* `uv run --extra dev ruff check ministack/app.py ministack/core/persistence.py ministack/services/batch.py tests/test_batch.py`
* `uv run --extra dev pytest tests/test_batch.py -q` with local MiniStack server (`16 passed`)
* `uv run --extra dev pytest tests/test_persistence.py -q` (`168 passed`)

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -344,7 +344,7 @@ _state_map = {
     "stepfunctions": "stepfunctions", "alb": "alb",
     "glue": "glue", "mwaa": "mwaa", "efs": "efs", "waf": "waf",
     "athena": "athena", "emr": "emr", "cloudfront": "cloudfront",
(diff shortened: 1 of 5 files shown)


Nahuel990:
Thanks!

PR #1145, opened 2026-07-22
fix(sqs): return NextToken from ListQueues when MaxResults is set

### LINKED ISSUE #1144: SQS ListQueues: results silently truncated — NextToken never returned nor accepted
## Summary

`SQS ListQueues` never returns a `NextToken` and never accepts one — results are silently truncated at `MaxResults` (default 1000). Any client that owns more than 1000 queues cannot enumerate them all, and gets no signal that the listing is incomplete.

## Current behaviour

```python
# ministack/services/sqs.py
def _act_list_queues(data: dict, _u: str) -> dict:
    pfx = data.get("QueueNamePrefix", "")
    mx = int(data.get("MaxResults", 1000))
    urls = [u for u, q in _queues.items()
            if not pfx or q["name"].startswith(pfx)]
    return {"QueueUrls": urls[:mx]}
```

Reproduce (with >N queues matching the prefix):

```bash
aws --endpoint-url=http://localhost:4566 sqs list-queues \
  --queue-name-prefix my-prefix --max-results 100 --no-paginate
# → exactly 100 QueueUrls, response contains no NextToken key
```

## Expected behaviour (AWS parity)

Per the [SQS ListQueues API](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ListQueues.html):

- `MaxResults` **not set** → up to 1000 results, `NextToken` is never included.
- `MaxResults` set and more results remain → response includes `NextToken`; passing it back returns the next page.
- Page ends exactly at the last result → no `NextToken`.

## Real-world impact

We run a backend with 1000+ SQS queues against MiniStack in CI. A startup check that lists queues by prefix silently lost the alphabetical tail of the fleet and reported healthy queues as missing — false positives that are indistinguishable from real missing queues without a per-queue `GetQueueUrl` double-check.

Same bug class as #1068 (DynamoDB `LastEvaluatedKey` when results end exactly at `Limit`), but for SQS.

Both the JSON protocol handler and the legacy XML/query path (`_to_xml` / `_normalise`, which doesn't forward `NextToken`) are affected.

I have a fix ready (integer-offset token, mirroring the existing SNS pagination pattern) — PR incoming.

bfreitastgtg:
Fixes #1144

## Problem

`SQS ListQueues` truncated results at `MaxResults` (default 1000) with no `NextToken` — emitted or accepted. Fleets larger than one page could not be enumerated, and callers had no signal the listing was incomplete (we hit this in CI with 1000+ queues: healthy queues reported as missing).

## Fix

AWS-parity pagination in `_act_list_queues`, using the integer-offset token pattern already used by the SNS handlers in this repo:

- `MaxResults` not set → up to 1000 results, never a `NextToken` ([documented AWS behaviour](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ListQueues.html))
- `MaxResults` set, more results remain → `NextToken` returned; passing it back continues from that offset
- Page ends exactly at the last result → no `NextToken` (same edge as #1068 for DynamoDB)

Both protocols covered: JSON handler, plus legacy XML response serialization (`_to_xml`) and query-param normalisation (`_normalise`) so old query-protocol clients paginate too.

## Tests

Three new integration tests in `tests/test_sqs.py`:
- `test_sqs_list_queues_paginates_with_max_results` — walks all pages via NextToken, asserts full coverage and no duplicates
- `test_sqs_list_queues_no_next_token_without_max_results`
- `test_sqs_list_queues_no_next_token_on_exact_fit`

`pytest tests/test_sqs.py -m "not serial"` → 84 passed.


Nahuel990 (approved):

Nahuel990:
Thanks @bfreitastgtg

PR #1157, opened 2026-07-22
feat(rds): emulate Aurora MySQL global replication

Areson:
## Why
MiniStack tracked Aurora global database membership but did not reproduce the MySQL data replication and lifecycle behavior applications depend on when exercising a provisioned global cluster.

## What
- Add GTID-based Aurora MySQL 8 replication between global writers and secondaries, with bounded binlog retention and fail-closed handling for legacy volumes
- Preserve replication across readiness, credential rotation, warm restart, headless secondaries, detach, and standalone re-add lifecycles
- Add recovery-safe detach and retry state so partial failures cannot leave stale channels or incorrect membership
- Add focused unit coverage plus a live Docker test for replay, streaming, password rotation, headless restart, detach, and cleanup
- Keep global failover and switchover outside this initial parity slice

## Risk Assessment
Medium — this changes MySQL startup and lifecycle behavior for Aurora MySQL 8 clusters, with most new behavior activated by global-database membership. Legacy initialized volumes fail closed, retention is bounded, and the full RDS/RDS Data suites plus the live replication path pass.

## Validation
* `uv run --extra dev ruff check ministack/services/rds.py tests/test_rds.py`
* `uv run --extra dev pytest tests/test_rds.py -q` with local MiniStack server (`174 passed, 6 skipped`)
* `uv run --extra dev pytest tests/test_rds_data.py -q` with local MiniStack server (`43 passed`)
* `git diff --check` (clean)

### DIFF
--- a/ministack/services/rds.py
+++ b/ministack/services/rds.py
@@ -87,6 +87,17 @@ _docker = None
 _ministack_network = None
 _shared_container_lock = threading.RLock()
 
+_MYSQL_REPLICATION_USER = "rdsrepladmin"
+_MYSQL_REPLICATION_PASSWORD = "ministack-rds-replication"
+_MYSQL_CONTROL_USER = "rdsadmin"
+_MYSQL_CONTROL_PASSWORD = "ministack-rds-control"
+_MYSQL_REPLICATION_RETRY_ATTEMPTS = 60
+_MYSQL_REPLICATION_RETRY_INTERVAL = 1
+_MYSQL_BINLOG_RETENTION_SECONDS = max(
+    1,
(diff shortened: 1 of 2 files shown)


Nahuel990:
@Areson I'm going to swap the env var for a hardcoded value before release. Binlog retention isn't exposed in AWS directly. Aurora configures it per cluster using the rds_set_configuration stored proc, so I'm exploring emulating that behaviour at the SQL layer (similar to how RDS handles it)

Nahuel990 (approved):

Areson:
@Nahuel990 Sounds good. I hadn't waded into trying to emulate Aurora sprocs as of yet, and wasn't sure if I wanted to go that deep.

Issue #1152, opened 2026-07-22
SQS Get from queue not working within lambda


Millroy094:
### Affected Service

Lambda

### What happened?

I’ve got a SQS queue for which I’ve created a API Gateway V2 route to Lambsa integration. The lambda is in charge of validating and post events to SQS. 

When I try to post events to SQS from lambda it complains it doesn’t all connection to localhost. 

It works can I specify the AWS endpoint in the nodejs sqs client. But you have to specifically specify it in the object initialisation. 

I’ve tried setting AWS endpoint in the lambda environment but it still doesn’t work, I have to specify it.

### Reproduction steps

```python
Create an sqs queue;
Create lambda for nodejs and post to the queue and see if the message gets through to it
```

### Expected behavior

You should be able to post the queue without having to specify the endpoint, it should be handled in the ministack lambda docker environment 

### MiniStack version

Latest

Nahuel990:
@Millroy094, could you clarify which AWS SDK you're using?

- If you're using **aws-sdk v2**, this behaviour is expected. We recommend migrating to **@aws-sdk/client-sqs (v3)**, which will automatically pick up the injected endpoint. Alternatively, you can continue passing the `endpoint` explicitly.
- If you're already using **AWS SDK v3**, please make sure you're on a recent version, as older releases predate support for the injected environment variables.

To help us reproduce and verify the issue, could you also share:

1. Which SDK you're using (`aws-sdk` v2 or `@aws-sdk/client-sqs` v3), along with the version.
2. The Node.js runtime configured for the Lambda function.
3. A minimal handler that reproduces the issue.

Millroy094:
@Nahuel990 is this fixed now? I am using SDK v3 and Node v22.

Here is my lambda handler

```

import type { APIGatewayProxyEventV2, APIGatewayProxyResult } from 'aws-lambda';
import { SendMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
import { GetSecretValueCommand, SecretsManagerClient } from '@aws-sdk/client-secrets-manager';
import { getBody, getHeader } from '@preference-center/api-gateway-util';

import { ConsentSnowflakeEventSchema } from './validation/schema';

type Event = APIGatewayProxyEventV2;

let cachedSecret: string | undefined;

async function getExpectedSecret(): Promise<string> {
  if (cachedSecret) {
    return cachedSecret;
  }

  const secretsManager = new SecretsManagerClient({});

  const response = await secretsManager.send(
[...]

PR #1201, opened 2026-07-24
fix(sqs): reflect caller Host header in QueueUrl so Docker Lambdas can reach queues (#1152)

thejusdutt:
## Summary

The Lambda-side environment injection was already correct — `AWS_ENDPOINT_URL` is injected and rewritten to `host.docker.internal` for the Docker executor. The actual failure is on the SQS side: every `QueueUrl` was rendered with a hardcoded `localhost` host (`_queue_url_for_account`).

The AWS JS SDK v3 SQS client sends requests to the **queue URL's host**, not the configured endpoint (`useQueueUrlAsEndpoint` defaults to `true`). So inside a Docker Lambda container, `SendMessage` connects to `localhost:4566` — the Lambda container itself — and is refused. That also explains the exact symptoms reported: setting the endpoint via environment variables doesn't help (the queue URL still wins), while hard-coding `endpoint` in the client constructor works (a custom client endpoint suppresses queue-URL-as-endpoint).

## Fix

SQS responses now echo the caller's `Host` header in `QueueUrl` / `QueueUrls` (the approach LocalStack uses). A Lambda calling through `host.docker.internal:4566` gets queue URLs it can actually reach; host-side clients keep getting `localhost` URLs.

- Internal queue state stays keyed on canonical URLs; lookups were already host-agnostic, so rewritten URLs round-trip from any client.
- An explicitly set `MINISTACK_HOST` disables rewriting entirely, preserving the existing contract (and `test_sqs_queue_url_reflects_env_host`).
- The Host header is validated against a strict charset before being echoed, and only URLs matching ministack's own canonical netloc are rewritten.
- Rewriting happens at a single choke point (`_dispatch`), so internal paths (SNS fan-out, ESM polling) keep canonical URLs.

## Testing

- `pytest tests/test_sqs.py` against a live server: **88 passed**, including 4 new tests — CreateQueue / GetQueueUrl / ListQueues with a forced `Host: host.docker.internal:4566` header assert the rewritten URL, and a round-trip test sends/receives through the rewritten URL via a localhost client.
- `ruff check` clean; CHANGELOG entry added.


Nahuel990:
@thejusdutt a few things:

Please don't link issues from PR comments, use the sidebar unless is your own reported issue.

We're still waiting on the reporter to clarify their setup, before we review this, can you tell us: are you hitting this issue yourself? If so, what's your setup?

thejusdutt:
Thanks for the steer on issue-linking — noted, I've kept any closing keyword out of the PR (the one remaining reference is just an automatic cross-reference, which I can't remove).

To answer your question directly: no, I'm not hitting this myself — I don't have that setup running. This PR came from reading the report against the source rather than a firsthand reproduction. The root cause looked clear from the code: SQS renders every `QueueUrl` with a hardcoded `localhost` host, and the JS SDK v3 client uses the `QueueUrl` as its request endpoint (`useQueueUrlAsEndpoint`), so a Lambda in the Docker executor receives a URL that points back at its own container. The tests I added cover the URL-rewriting behavior via crafted `Host` headers, but I want to be upfront that I did **not** reproduce the full Docker-Lambda-to-SQS path end to end.

So I completely understand waiting on the reporter before reviewing. If it's useful I'm happy to add a genuine end-to-end test once their setup is confirmed — or if you'd rather I close this and let them drive.

Nahuel990:
@thejusdutt no worries! you are adding a valuable fix, the problem is when I merge this it will close the other open issue and we need to confirm if the user is having this issue or is related to the node sdk

Nahuel990 (approved):

PR #1204, opened 2026-07-25
fix(s3tables): Iceberg REST compat fixes for DuckDB and Spark

squirmy:
#### DuckDB fixes (ALTER TABLE / CREATE TABLE):
- Advance last-column-id on add-schema commits; it never moved, so field
  IDs collided across ALTER TABLE ADD COLUMN calls
- Recognize add-spec, the Iceberg REST spec's actual wire action name for
  adding a partition spec, alongside the old non-standard add-partition-spec
- Honor the partition spec supplied at create-table time instead of
  hardcoding an empty one
- Switch Iceberg REST 404s to the spec's IcebergErrorResponse shape
  ({"error": {"message", "type", "code"}}) instead of AWS JSON-1.0

#### Spark and retry-safety fixes:
- Populate next-row-id on V3 table metadata (it was never set at all,
  so Spark failed to parse metadata: "Cannot parse missing long:
  next-row-id"); advance it by each snapshot's added-rows on commit
- Make add-schema/add-spec/add-sort-order idempotent by id, so
  re-declaring an unchanged one (Spark does this on every write) no
  longer piles up duplicate entries that crash id-keyed lookups
  ("Multiple entries with same key")
- Make add-snapshot idempotent by snapshot-id for the same reason, on a
  resent commit after a client retry
- Reject a resent CreateTable with 409 instead of silently wiping an
  existing table's metadata, matching the control-plane CreateTable's
  existing conflict check

### DIFF
--- a/ministack/core/responses.py
+++ b/ministack/core/responses.py
@@ -421,6 +421,20 @@ def error_response_json(code: str, message: str, status: int = 400) -> tuple:
     }, body
 
 
+def error_response_iceberg(error_type: str, message: str, status: int = 400) -> tuple:
+    """Iceberg REST catalog error response, per the spec's ErrorModel.
+
+    This is a different wire shape from AWS's own JSON-1.0 errors (`error_response_json`)
+    -- clients speaking the Iceberg REST protocol (e.g. duckdb-iceberg) require a top-level
+    `error` object with `message`/`type`/`code`, or they can't tell "not found" apart from
+    any other failure and raise an opaque error instead of proceeding (e.g. treating a
+    missing table as create-worthy).
+    """
+    data = {"error": {"message": message, "type": error_type, "code": status}}
+    body = json.dumps(data, ensure_ascii=False).encode("utf-8")
+    return status, {"Content-Type": "application/json"}, body
+
+
 def now_iso() -> str:
     """Current time in AWS ISO format."""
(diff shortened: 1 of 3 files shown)


Nahuel990:
@squirmy I pulled in most of it. The Iceberg REST error envelope was the key fix, and it shipped in 1.4.7, so a `LoadTable` 404 now comes back as a proper `NoSuchTableException` / `NoSuchNamespaceException`. That allows DuckDB and Spark to proceed with table creation instead of aborting. The `add-schema` `last-column-id` bump, the `add-spec` action name, and the `create-table` 409-on-exists behavior all went in too.

The one thing I left out is the commit idempotency / dedup. Real AWS S3 Tables doesn't deduplicate snapshots server-side; it relies on the Iceberg commit requirements plus optimistic concurrency and expects the client to retry on conflict (the `aws-samples` streaming examples do exactly that with backoff). Adding server-side dedup would push us past AWS behavior, and parity is the line we hold here, so that part stayed out.

Rather than sending the branch back to you for those trims, I applied the clean parts directly on top of `main` so it lands in this release. Credit is in the changelog.

squirmy:
Thanks - i've created a follow up PR. I think I got a little carried away with this one - but some of those  changes are still necessary to avoid a crash in spark. 👍

PR #1210, opened 2026-07-26
fix(sns): honour {"exists": false} in subscription filter policies

ddy4633:
## The failure

A subscription with this filter policy receives **nothing** — not "the wrong messages", nothing at all:

```json
{"store": [{"exists": false}]}
```

```
attribute absent   ->  not delivered   (AWS delivers)
attribute present  ->  not delivered   (correct)
```

Both branches decline, so the subscriber is permanently dead.

## Root cause

`ministack/services/sns.py:1477-1479` returned `False` the moment a policy key was missing from the published message attributes, before any rule for that key was looked at:

```python
attr = message_attributes.get(key)
if attr is None:
    return False
```

`exists: false` is evaluated against **presence**, not against a value, so it can never be reached from that branch. The `exists: true` half was already implemented in `_attr_matches_any`, which is why the gap is easy to miss.

## Why this is a bug rather than an intentional gap

MiniStack's own EventBridge matcher implements the same operator correctly, with a comment that describes exactly what SNS gets wrong — `ministack/services/eventbridge.py:1058-1073`:

> `[{"exists": true|false}]` is evaluated against key presence/absence, NOT against the value, so an absent key must NOT short-circuit to False before that check.

Same project, same operator, one surface has it and the other doesn't.

No open issue or PR mentions SNS filter policies.

## The fix

Nine lines in `_matches_filter_policy`: normalize the rule list first, then on an absent attribute check for an `{"exists": false}` rule and `continue` instead of returning. Present-attribute behaviour is untouched.

## How to test

```
hypercorn ministack.app:app --bind 127.0.0.1:4566 &
pytest tests/test_sns.py -q          # 85 passed
```

One test added — `test_sns_filter_policy_exists_false_matches_absent_attribute` — using boto3 with an SQS subscriber, in the style of the neighbouring `test_sns_filter_policy_blocks_non_matching`. It asserts both directions: attribute present ⇒ not delivered, attribute absent ⇒ delivered.

Regression proof — source reverted, server restarted on clean source, test kept:

```
>       assert len(msgs2.get("Messages", [])) == 1, "Message without the attribute must be delivered"
E       AssertionError: Message without the attribute must be delivered
E       assert 0 == 1
FAILED tests/test_sns.py::test_sns_filter_policy_exists_false_matches_absent_attribute
```

CHANGELOG `Unreleased → Fixed` entry included per the PR checklist.

## What I could not verify

**My baseline is not green.** This machine has no Docker, so 96 tests that need a Lambda/ECS/MWAA container fail before any change of mine. The SNS surface itself is 100% green, and `_matches_filter_policy` is reachable only from `_fanout`, but I want to state plainly that "no regressions" here rests on before/after diffing rather than on a fully green run.

Full-suite diff after the fix: my test flipped fail→pass, `test_cfn_apigwv2_full_http_api_stack_invokes_lambda` also flipped to pass, and 5 apigateway/appsync tests flipped to fail. Re-running those 5 in isolation **on the clean stashed tree** also fails 3 of them, with a different split per invocation — they are order/state-dependent flakes inside the already-broken Docker-dependent cluster, and none of them touch SNS.

**No AWS ground truth.** I have no credentials to check `exists: false` against real SNS. The justification is AWS's published filter-policy semantics plus MiniStack's own EventBridge implementation. If SNS deliberately diverges here, this PR is wrong — but nothing in the code or docs says so.

## Adjacent bugs I did not touch

Same function, all real, all separate changes — happy to file issues if useful:

- `FilterPolicyScope: MessageBody` returns `True` unconditionally; message-body filtering is not implemented at all
- `anything-but` compares strings against numeric list members
- `_check_numeric` silently ignores a trailing lone operator


Nahuel990:
@ddy4633 Sorry could you send how are you reproducing this issue against AWS? this is not AWS behaviour

PR #1212, opened 2026-07-27
fix(s3tables): make add-schema/add-spec/add-sort-order idempotent by id

squirmy:
Hey @Nahuel990 - raising this because there's still some issues when trying to use spark / iceberg on my end with the 1.4.7. This PR resolves that. It introduces some of the changes again, but leaves the commit idempotency out. Details below.

That was something I added after a review, and wasn't the right approach. I believe the **add-schema/add-spec/add-sort-order** are good to keep in though.

## Problem

Spark's `compact.py` crashed with `IllegalArgumentException: Multiple entries with same key: 0=tabl...` — Iceberg's `schemasById()` found two `schema-id: 0` entries in the table's metadata (`0=tabl` is `Schema.toString()`, `"table {...}"`).

## Root cause

`_iceberg_commit_table` appended unconditionally on `add-schema`/`add-spec`/`add-sort-order`, with no check for an already-registered id. Spark re-declares the current, unchanged schema/spec/sort-order on ordinary writes (not retries), so a second commit duplicates the entry, and the next load crashes building the id-keyed map.

## Relation to #1204 / #1207

Most of #1204 landed via #1207. What was held back on AWS-parity grounds (server shouldn't dedup; client should retry via OCC) — was bucketed as "commit idempotency." That reasoning fits `add-snapshot` dedup, but not this:

- **add-schema/add-spec/add-sort-order dedup-by-id (this PR)** — not a parity question. It's a structural invariant of `TableMetadata` (no duplicate ids) that Iceberg's own builder enforces too, and it isn't retry-driven — Spark sends these on every write regardless of conflict, so OCC/requirements checking wouldn't prevent it either.
- **add-snapshot dedup — still deferred.** ministack validates zero commit `requirements`, so a real retry could hit the same crash via `snapshotsById()`. Correct fix is requirements validation (`assert-ref-snapshot-id`, etc. → 409), not blind dedup — separate follow-up.

## Change

`s3tables.py`: `add-schema`/`add-spec`/`add-partition-spec`/`add-sort-order` now no-op if the id is already present. `test_s3tables.py`: 3 new regression tests (resend same commit twice, assert no duplicate id) — reproduce the crash on unfixed code, pass with the fix. Full suite: 23/23 passing.

### DIFF
--- a/ministack/services/s3tables.py
+++ b/ministack/services/s3tables.py
@@ -524,7 +524,13 @@ def _iceberg_commit_table(namespace, table_name, data, allow_cross_region):
                     "type": update.get("type", "branch")}
             elif action == "add-schema":
                 new_schema = update.get("schema", {})
-                metadata.setdefault("schemas", []).append(new_schema)
+                existing = metadata.setdefault("schemas", [])
+                # Idempotent by schema-id: a client re-declaring its current,
+                # unchanged schema on every write (Spark does this) must not pile
+                # up a second entry with the same id -- Iceberg's schemasById()
+                # crashes on load with "Multiple entries with same key" if it does.
+                if not any(s.get("schema-id") == new_schema.get("schema-id") for s in existing):
+                    existing.append(new_schema)
                 # Advance last-column-id to the highest field ID in the new
                 # schema, so a client allocating the next column's ID (e.g.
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

Issue #1206, opened 2026-07-26
AWS::KinesisFirehose::DeliveryStream not supported in CloudFormation (Iceberg destination)


ryan-bennett:
## Summary

`AWS::KinesisFirehose::DeliveryStream` isn't wired up in CloudFormation — any stack containing one fails immediately with `Unsupported resource type: AWS::KinesisFirehose::DeliveryStream`, even though the direct Firehose API (`ministack/services/firehose.py`) already implements `CreateDeliveryStream` and friends.

Confirmed via a minimal `cdklocal deploy` repro: a stack with just an `AWS::S3Tables::TableBucket` + an `AWS::KinesisFirehose::DeliveryStream` (Iceberg destination) deploys the table bucket fine (`CREATE_COMPLETE`) and fails outright on the delivery stream.

## Which AWS service / API namespace

Kinesis Data Firehose (`firehose`) — specifically the `AWS::KinesisFirehose::DeliveryStream` CloudFormation resource type. The direct API already exists in `ministack/services/firehose.py`; this is about registering that existing service in the CloudFormation provisioner layer (`ministack/services/cloudformation/provisioners.py`'s `_RESOURCE_HANDLERS`), the same way `AWS::S3Tables::TableBucket`/`Namespace`/`Table` are wired to `ministack/services/s3tables.py`.

## What I actually need

1. **CFN resource-type registration** — `create`/`delete` provisioners that translate the CFN `Properties` shape (`DeliveryStreamType`, `S3DestinationConfiguration`, `IcebergDestinationConfiguration`, etc.) into whatever `firehose._create_delivery_stream()` / `_delete_delivery_stream()` already expect, mirroring the existing `_s3tables_bucket_create`-style provisioners.
[...]

PR #1214, opened 2026-07-27
fix(s3tables): wire CFN table schema through for Iceberg delivery

ryan-bennett:
Fixes the second of two gaps blocking Firehose's Iceberg destination from actually delivering data (first gap — duckdb missing from the standard image — turned out to already be solved by the `:full` image variant, see comments below; that part of this PR has been dropped).

## The bug

With `ministackorg/ministack:full` (which already ships `duckdb`), Firehose delivery to an Iceberg destination still failed:
```
Binder Error: Table "spike_table" does not have a column with name "id"
```
`_s3tables_table_create` (the `AWS::S3Tables::Table` CFN provisioner, in `ministack/services/cloudformation/provisioners.py`) hardcodes the Iceberg schema to `[]` and never reads `IcebergMetadata.IcebergSchema.SchemaFieldList` from the template — so every table created via CloudFormation ends up with zero columns, regardless of what schema is declared. The direct `CreateTable` API path (`_create_table` in `s3tables.py`) already parses this correctly; the CFN provisioner was just never wired up to do the same.

## Fix

Parse `SchemaFieldList` the same way the API path does.

## Tests

Added `test_cfn_s3tables_table_schema_from_iceberg_metadata` in `tests/test_cfn.py`: creates a table via CFN with a declared schema, then asserts the columns actually show up via the Iceberg REST catalog's `LoadTable`. Confirmed it fails against unpatched `main` (`got set()`) and passes with the fix.

## End-to-end verification

Deployed a CDK stack with `AWS::S3Tables::TableBucket`/`Namespace`/`Table` + `AWS::KinesisFirehose::DeliveryStream` (Iceberg destination) + an EventBridge rule targeting it, against `ministackorg/ministack:full` with this fix applied. `PutRecord` → polled the underlying S3 bucket → confirmed real objects landed:
```
spike_ns/spike_table/data/<uuid>.parquet
spike_ns/spike_table/metadata/<uuid>-m0.avro
spike_ns/spike_table/metadata/snap-<id>-<uuid>.avro
```
Full `test_cfn.py` + `test_s3tables.py` suites also pass against the patched `:full` image (153 passed, 2 skipped, 1 pre-existing unrelated failure that reproduces identically on unpatched `main`).

Refs #1206.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,19 @@ RUN pip install --no-cache-dir --no-compile \
         "boto3>=1.34" \
         "awscli"
 
+# duckdb publishes no musllinux wheels on PyPI, so it must be compiled from
+# source here — this is why it was excluded from the image previously (see
+# the original Dockerfile's now-removed "excluded intentionally" comment),
+# but the Firehose Iceberg-delivery path (added in 1.4.7) now hard-depends
+# on it at runtime, so the image needs to actually provide it.
+# CMAKE_BUILD_PARALLEL_LEVEL is capped at 3: an unbounded ninja job count
+# (one per core) can OOM-kill the compiler mid-build, since several of
+# DuckDB's unity-build translation units need multiple GB of RAM each on
+# their own; 3 keeps peak memory well under typical CI budgets at the cost
+# of a longer build.
+RUN apk add --no-cache gcc g++ make cmake ninja git python3-dev musl-dev linux-headers \
+    && CMAKE_BUILD_PARALLEL_LEVEL=3 pip install --no-cache-dir --no-compile duckdb
+
 # Strip awscli help examples (~25 MB) and Python cache files (~15 MB).
 RUN rm -rf /usr/local/lib/python3.13/site-packages/awscli/examples \
     && find /usr/local/lib/python3.13/site-packages -type d -name '__pycache__' -exec rm -rf {} + 2>/dev/null \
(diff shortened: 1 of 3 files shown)


Nahuel990:
@ryan-bennett we keep 2 images, latest and full. For this you need to use full. https://github.com/ministackorg/ministack#quick-start

ryan-bennett:
Thanks — you're right, and confirmed: `duckdb` imports and runs fine out of the box in `ministackorg/ministack:full`. I've dropped the Alpine/duckdb-compile part of this PR entirely (it was solving a problem `:full` already solves, and per DuckDB's own docs it runs ~30% slower on musl anyway, so building it there wasn't even the better option).

What's left is the second bug I found while testing against `:full`: the `AWS::S3Tables::Table` CFN provisioner discards the declared Iceberg schema entirely (hardcoded to `[]`), so Firehose failed with `does not have a column with name "id"` even with `duckdb` present and working. That's fixed now and re-verified end-to-end against `ministackorg/ministack:full` — a real `PutRecord` → Firehose → Iceberg write now lands actual Parquet + Iceberg metadata/snapshot files in the S3 Tables bucket. PR description and title updated to match. Apologies for the noise on the image mixup.

Nahuel990 (approved):

Nahuel990:
Thanks @ryan-bennett !

Nahuel990 (approved):

PR #1215, opened 2026-07-27
fix(stepfunctions): resolve indexed context paths

felixp-square:
👋 I'm fixing a small issue I ran across when testing step functions:

 ## Issue

  Step Functions context-object paths only traversed dot-separated dictionaries. As a result, reference paths containing array
  indexes, such as `$$.Map.Item.Value.values[0].name`, resolved to `null` in a Map state's `ItemSelector`, including when
  supplied to an intrinsic function.

  For example, given:

  ```json
  {"values": [{"name": "first"}]}
```

  `$$.Map.Item.Value.values[0].name` should resolve to `"first"`.

  ## Fix

Preserve validation and root selection for `$$.` context-object paths and delegate the remaining traversal to the existing JSONPath resolver.

  ## Testing

  - Added regression test:
    - `test_sfn_map_item_selector_context_path_resolves_array_index`
    - Covers direct `ItemSelector` resolution and use as a `States.Format` argument
  - Reran existing context-path and intrinsic-function tests:
    - `test_sfn_pass_parameters_resolve_context_object`
    - `test_sfn_intrinsic_format`
  - Result: `3 passed`
  - `uv run ruff check ministack/services/stepfunctions.py tests/test_stepfunctions.py`

  Repository-wide `uv run ruff check ministack/` currently reports six unrelated violations in untouched `dynamodb.py` and
  `lambda_durable.py` files.

  ## References

  - https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html
  - https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-paths.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- **Step Functions — context object paths now resolve array indexes** — JSONPath references such as `$$.Map.Item.Value.values[0].name` returned `null` because context paths used a separate dotted-key-only resolver. Context paths now retain their distinct `$$.` root while sharing the standard path traversal, so indexed references work in `ItemSelector` fields and intrinsic-function arguments. Contributed by @felixp-square.
+
 ## [1.4.7] — 2026-07-26
 
 ### Added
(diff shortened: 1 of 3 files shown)


Nahuel990:
Thank you @felixp-square !

Nahuel990 (approved):

PR #1244, opened 2026-08-01
Harden Glue and AppSync flaky tests

Areson:
## Why

The Glue crawler tests depended on observing short-lived asynchronous states, which can be missed under shared-server or parallel test scheduling. A few AppSync Events realtime fan-out tests also exercise shared websocket delivery behavior and can interfere with the parallel-safe test lane.

## What

- Update Glue crawler tests to wait for stable completion state instead of requiring the transient `RUNNING` state to be observed immediately.
- Keep race-safe handling for `StopCrawler` while preserving deterministic coverage that a running crawler can be stopped successfully.
- Move the four AppSync Events realtime fan-out websocket tests into the existing serial test phase.

## Behavior changes and risk

This is a test-only change. It does not modify MiniStack service behavior.

Risk is limited to test coverage shape: the Glue assertions now verify stable final crawler state and include direct service-level stop coverage, while the AppSync tests remain fully asserted but run serially.

## Validation

- `uv run --extra dev ruff check tests/test_glue.py tests/conftest.py` → passed
- `uv run --extra dev pytest -q tests/test_glue.py::test_glue_crawler_v2 tests/test_glue.py::test_glue_stop_crawler tests/test_glue.py::test_glue_stop_crawler_service_stops_running_crawler` → 3 passed
- `uv run --extra dev pytest -q -m serial tests/test_appsync_events.py::test_websocket_subscribe_receives_published_event tests/test_appsync_events.py::test_websocket_single_level_wildcard tests/test_appsync_events.py::test_websocket_unsubscribe_stops_delivery tests/test_appsync_events.py::test_websocket_publish_frame_acks_and_fans_out` → 4 passed
- `uv run --extra dev pytest --collect-only -q -m serial tests/test_appsync_events.py::test_websocket_subscribe_receives_published_event tests/test_appsync_events.py::test_websocket_single_level_wildcard tests/test_appsync_events.py::test_websocket_unsubscribe_stops_delivery tests/test_appsync_events.py::test_websocket_publish_frame_acks_and_fans_out` → 4 collected
- `uv run --extra dev pytest --collect-only -q -m "not serial" tests/test_appsync_events.py::test_websocket_subscribe_receives_published_event tests/test_appsync_events.py::test_websocket_single_level_wildcard tests/test_appsync_events.py::test_websocket_unsubscribe_stops_delivery tests/test_appsync_events.py::test_websocket_publish_frame_acks_and_fans_out` → no tests collected; 4 deselected
- `git diff --check upstream/main...HEAD` → passed

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -103,6 +103,12 @@ _SERIAL_TESTS = {
     "tests/test_appsync.py::test_appsync_lambda_failing_authorizer_returns_unauthorized",
     # AppSync Events service mutations; shared state racing under xdist.
     "tests/test_appsync_events.py::test_publish_with_appsync_sigv4_scope_on_events_vhost",
+    # AppSync Events realtime websocket fan-out uses shared subscriptions and
+    # connection delivery state; keep fan-out assertions out of the xdist lane.
+    "tests/test_appsync_events.py::test_websocket_subscribe_receives_published_event",
+    "tests/test_appsync_events.py::test_websocket_single_level_wildcard",
+    "tests/test_appsync_events.py::test_websocket_unsubscribe_stops_delivery",
+    "tests/test_appsync_events.py::test_websocket_publish_frame_acks_and_fans_out",
     # Credential report reflects all users in the account; run serially to avoid
     # parallel-test interference on the account-global CSV snapshot.
     "tests/test_iam.py::test_iam_credential_report_mfa_and_password",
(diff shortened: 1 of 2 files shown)


Nahuel990:
@Areson I don't see the need to stop asserting "Running" we already have a 5 seconds delay to avoid the race condition

Is this something you're seeing in your local setup? Based on CI, it has only failed twice and both failures were due to GitHub Actions issues

PR #1253, opened 2026-08-02
IoT — binary topic-rule payloads are no longer corrupted

maximoosemine:
### What

**IoT — binary topic-rule payloads are no longer corrupted.** Stacked on #<SQL evaluation PR>. Rule events were built with `json.loads(payload.decode("utf-8"))` falling back to `decode("utf-8", "replace")`, so a payload that is not valid UTF-8 reached the action with every non-ASCII byte replaced by `U+FFFD` — unrecoverable, so binary telemetry could not travel the rule path at all.

The published bytes are now threaded into the projection, and `encode(<expr>, 'base64')` is supported: `encode(*, 'base64')` encodes the payload exactly as published, so a rule declaring `SELECT encode(*, 'base64') AS data FROM 'telemetry'` delivers `{"data": "<base64>"}` whose `b64decode` returns the publish payload byte-for-byte. That is the pattern AWS documents for Lambda actions, which "don't support binary payload input" and require base64 inside a JSON payload. No lossy decode remains anywhere on the publish → rule → invoke path.

A payload that is not valid UTF-8 and whose SELECT projects no attributes (`SELECT *`) now dispatches nothing rather than corrupted text. AWS treats that combination as unsupported for a Lambda action and raises a rule error; MiniStack has no error-action support, so it logs and skips.

### Testing Plan

- Unit tests added to `tests/test_iot.py` (`encode(*, 'base64')` round-tripping a binary payload, `SELECT *` skipping a payload that is not valid UTF-8) and integration tests to `tests/test_iot_data.py` covering `encode(*, 'base64')` over Basic Ingest and over the topic-filter path, both asserting a byte-for-byte round-trip.
- The payload under test is every byte value (`bytes(range(256))`), so the assertions cover the full range rather than one encoding's byte patterns.
- `pytest tests/test_iot.py tests/test_iot_data.py` passes with the parent PR applied.
- `ruff check ministack/` passes with no new findings on the changed files.


maximoosemine:
This is stacked on https://github.com/ministackorg/ministack/pull/1252. Merge https://github.com/ministackorg/ministack/pull/1252 first.

PR #1254, opened 2026-08-02
DynamoDB Streams — backlogs and ESM read positions leaked, degrading a long-lived container

maximoosemine:
### What

**DynamoDB Streams — backlogs and ESM read positions leaked, degrading a long-lived container.** Every stream record a table emitted stayed resident for the life of the process: the per-table list was never trimmed, `DeleteTable` dropped the table's tags, TTL, PITR, and Kinesis destinations but not its backlog, and `DeleteEventSourceMapping` left the ESM's stream position, Kinesis positions, and invoke-backoff entry behind. A container cycling stacks with auto-named tables, where each deploy creates a new stream, accumulated both indefinitely, so the poller walked a growing set every tick and stream→Lambda delivery slowed until it was restarted. Records now expire 24 hours after they are written — the whole of AWS's rule, which caps retention by time and not by count — and both deletes release their state, from the CloudFormation delete paths as well as the API ones. Because expiry shifts the surviving slice, read positions became absolute positions tracked against a per-table trim horizon: `TRIM_HORIZON` anchors there, `LATEST` past every record held, and a consumer behind expired records resumes at the horizon, as an AWS shard iterator does.

### Testing Plan

- Tests added to `tests/test_dynamodb.py` (age expiry advancing the trim horizon, reads resuming there, and `DeleteTable` releasing the backlog — asserted over the wire via an empty `TRIM_HORIZON` iterator on a recreated table) and to `tests/test_lambda.py` (ESM delete releasing poller state, the poller resuming at the horizon, `LATEST` / `TRIM_HORIZON` anchoring on a trimmed stream).
- `pytest tests/test_lambda.py tests/test_dynamodb.py tests/test_dynamodb_streams.py tests/test_pipes.py tests/test_persistence.py` passes in both phases; `test_lambda_docker_flags_applied_to_run_kwargs` fails identically with and without this change.
- `ruff check ministack/` passes with no new findings on the changed files.


Nahuel990:
@maximoosemine bug is correctly spotted, but trimming breaks the Streams API: `_get_records` reads by absolute index while `_trim_stream_records` shifts the list. Could you add the same fix there too, along with a test? Also can you share how did you spot this issue without touching the stream api?

maximoosemine:
@Nahuel990 This should be good to merge whenever you're ready.

Nahuel990:
@maximoosemine I was still trying to understand how you ran into the error without the stream API

Nahuel990 (approved):

maximoosemine:
@Nahuel990 sorry for the delayed response. I had a long running container that I ran e2e tests against. I noticed this issue after the container had been alive for a while and there was a growing delay between when I put rows in a DynamoDB table and when the effects of the attached Lambda manifested.

Thank you for the merge!

Issue #1035, opened 2026-07-04
rawQueryString comes out URL-decoded, breaks lambda_http


crestonbunch:
### Affected Service

API Gateway (v2)

### What happened?

the v2 event's rawQueryString comes back URL-decoded, so %20 turns into a literal space. lambda_http builds the request URI from rawQueryString, and http::Uri rejects the space, hence the panic. On real AWS rawQueryString stays percent-encoded, so the same request works.

### Reproduction steps

```python
1. Put an HTTP API (payload format 2.0) in front of a Rust function built on lambda_http (provided.al2023).
2. Call it with a query param whose value has a space, e.g. GET /authorize?scope=openid%20profile.
3. You get a 502, and the function log shows:
failed to build request: http::Error(InvalidUri(InvalidUriChar))
```

### Expected behavior

Expected: rawQueryString percent-encoded like AWS, so lambda_http functions don't choke on spaces or any other char that's invalid raw in a URI.

### MiniStack version

1.3.63

### SDK / tool used

AWS Rust SDK

Nahuel990:
@crestonbunch this should be fixed in v1.3.71

Issue #406, opened 2026-04-20
API Gateway v2 HTTP API ignores `cors_configuration` — wildcard CORS returned unconditionally


whittin3:
## Summary

On an API Gateway v2 HTTP API, the `cors_configuration` specified at API creation time has no effect — MiniStack 1.3.5 returns a wildcard CORS response on every OPTIONS preflight regardless of what was configured. Browsers sending `credentials: "include"` reject the `*` origin, breaking cookie-authenticated APIs.

## Version

- MiniStack: `ministackorg/ministack:1.3.5`
- Terraform AWS provider: `6.x`

## Steps to reproduce

Create an HTTP API with an explicit `cors_configuration` that restricts origins and enables credentials:

```hcl
resource "aws_apigatewayv2_api" "demo" {
  name          = "demo-http"
  protocol_type = "HTTP"

  cors_configuration {
    allow_origins     = ["http://localhost:3000"]
    allow_methods     = ["GET", "POST", "OPTIONS"]
    allow_headers     = ["content-type", "cookie"]
    allow_credentials = true
    max_age           = 600
  }
}
```

Apply against MiniStack and send an OPTIONS preflight:

```sh
$ curl -i -X OPTIONS "http://<api-id>.execute-api.localhost:4566/live/api/hello" \
    -H 'Origin: http://localhost:3000' \
    -H 'Access-Control-Request-Method: GET'
HTTP/1.1 200
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, PUT, DELETE, HEAD, OPTIONS, PATCH
Access-Control-Allow-Headers: *
Access-Control-Expose-Headers: *
Access-Control-Max-Age: 86400
```

(The `/live/` path prefix is the workaround for #404 — the same CORS behavior surfaces without it.)

## Expected behavior

The response should reflect what's in `cors_configuration`:

```
Access-Control-Allow-Origin: http://localhost:3000
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Allow-Headers: content-type, cookie
Access-Control-Allow-Credentials: true
Access-Control-Max-Age: 600
```

Requests from origins not in `allow_origins` should not receive these headers.

From [Configuring CORS for an HTTP API](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-cors.html):

[...]

PR #1260, opened 2026-08-03
feat(lambda): serve the Function URL data plane

liammizrahi:
## Why

`CreateFunctionUrlConfig` stores a config and returns a `{urlId}.lambda-url.{region}.on.aws` URL, but nothing ever served it. A request addressed to that host matched no route in the gateway, fell through to S3 virtual-host addressing, and came back as:

```xml
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message></Error>
```

So the control plane looks complete — Terraform's `aws_lambda_function_url` applies, `AWS::Lambda::Url` provisions since 1.4.6 — while the URL it hands you is inert. Anything whose local topology puts a proxy or CloudFront-shaped edge in front of a Function URL can't be exercised at all.

`lambda-url` is also missing from `_S3_VHOST_EXCLUDE_RE`, which is what turns the miss into a confusing S3 error rather than a 404. That's the same class of fall-through the 1.3.17 routing sweep fixed for the Function URL *control-plane* paths (`/2021-10-31/functions/.../url`); this is the data-plane half.

## What

Resolve the URL id to its function and invoke it.

- **Addressing** — host-based `{urlId}.lambda-url.{region}.*`, plus a path-based `/_aws/lambda-url/{urlId}/...` form. The path form exists for the reason `_parse_execute_api_url` already documents: browsers on macOS don't resolve `*.localhost`, and many clients can't override `Host`. The host pattern deliberately accepts any suffix, since the stored URL carries AWS's own `.on.aws`.
- **Event** — payload format 2.0, matching what a Function URL actually delivers: `$default` for both `routeKey` and `stage`, no `pathParameters`, `rawQueryString` percent-encoded the way AWS sends it, and `body` / `queryStringParameters` omitted rather than sent as `null` (the same strict-validator problem #1035 and the v2 proxy event fixed).
- **Auth** — `AuthType` is enforced. `AWS_IAM` returns `403 Forbidden` to an unsigned request and accepts header-signed and presigned SigV4; `NONE` is open.
- **CORS** — the `Cors` config drives both preflight and response headers. An origin outside `AllowOrigins` gets no CORS headers at all, as on AWS, so a browser blocks it the same way. A Function URL now owns its `OPTIONS` preflight instead of getting the generic wildcard one, mirroring the execute-api carve-out from #406.
- **`InvokeMode: RESPONSE_STREAM`** — `awslambda.HttpResponseStream.from()` frames a reply as a JSON prelude, eight NUL bytes, then the body. Real Function URLs consume that framing at the edge; without it the prelude leaks into the response body and the caller sees `{"statusCode":200,...}\x00\x00\x00\x00\x00\x00\x00\x00` prepended to its payload. The prelude now supplies status and headers. A handler that writes to the stream without calling `HttpResponseStream.from()` sends no prelude and falls back to `200`, as AWS does.
- **Cookies** — the format-2.0 `cookies` array becomes `Set-Cookie` headers.

Header merging is case-insensitive throughout, so a lowercase `content-type` from a handler or a prelude replaces the seeded `Content-Type` rather than shipping alongside it — the same handling #750 applied to the v1 `multiValueHeaders` merge.

## Changes

- `ministack/app.py` — `_LAMBDA_URL_RE`, `_parse_lambda_url`, `_handle_lambda_url_request` dispatched ahead of S3 virtual-host; `lambda-url` added to `_S3_VHOST_EXCLUDE_RE`; `OPTIONS` deferred to the Function URL when it has a CORS config.
- `ministack/services/lambda_svc.py` — URL-id resolution across account/region scopes, format-2.0 event builder, auth and CORS handling, prelude-aware and buffered response shaping.
- `tests/test_lambda.py` — 8 integration tests: event shape, absent query string, request body, path-based addressing, unknown id, `AWS_IAM` signed and unsigned, CORS allowed and disallowed origins, and prelude consumption.

## Scope

Not a new service, and no Dockerfile / CI / dependency changes — this is the data plane for operations the Lambda emulator already exposes.

Streaming is delivered as one response rather than incrementally. That is a property of the executor, not of this change: MiniStack invokes through the AWS RIE's `/2015-03-31/functions/function/invocations`, which buffers, and the RIE exposes no `response-streaming-invocations` endpoint (404). Confirmed against `public.ecr.aws/lambda/nodejs:22` — a handler emitting five chunks a second apart delivers all five at once when the handler returns.

Worth noting for a follow-up: the AWS base image's `/lambda-entrypoint.sh` execs `/var/runtime/bootstrap` directly when `AWS_LAMBDA_RUNTIME_API` is set, bypassing the RIE. Pointing that at a Runtime API served by MiniStack yields genuinely incremental chunks — the RIC sets `Lambda-Runtime-Function-Response-Mode: streaming` with chunked encoding on its own. I verified this against the same image and got chunks at the handler's real 1s cadence rather than batched at the end. That's an executor change with a much wider blast radius, so it isn't in this PR; happy to open an issue for it if it's of interest.

## Testing

```
pytest tests/test_lambda.py -k function_url    # 13 passed
pytest tests/test_lambda.py -m "not serial"    # 276 passed, same 9 pre-existing
                                               # environment-dependent failures as on main
ruff check ministack/                          # no new findings
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to MiniStack will be documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 Versioning follows [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+- **Lambda — Function URL data plane** — `CreateFunctionUrlConfig` returned a `{urlId}.lambda-url.{region}.on.aws` URL that nothing served: a request addressed to it matched no route, fell through to S3 virtual-host addressing, and came back as `NoSuchBucket`. Function URLs are now invocable. A request reaching the gateway on a `{urlId}.lambda-url.{region}.*` host — or on the path-based `/_aws/lambda-url/{urlId}/...` form, for clients that can't set `Host` and browsers that won't resolve `*.localhost` — resolves the URL id to its function and invokes it with a payload-format-2.0 event carrying `$default` for `routeKey` and `stage`, `rawPath`/`rawQueryString` percent-encoded as AWS sends them, and `body`/`queryStringParameters` omitted rather than null. `AuthType` is enforced (`AWS_IAM` returns `403 Forbidden` to an unsigned request, header-signed and presigned both pass; `NONE` is open), the `Cors` config drives preflight and response headers with a non-allowed origin getting none, and `InvokeMode: RESPONSE_STREAM` responses are unwrapped from the `HttpResponseStream` framing so the prelude supplies the status and headers instead of leaking into the body. Cookies returned via the format-2.0 `cookies` array become `Set-Cookie` headers. Contributed by @liammizrahi.
+
 ## [1.4.10] — 2026-08-03
 
 ### Added
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -39,6 +39,11 @@ if _VERSION == "dev":
 _EXECUTE_API_RE = re.compile(
     r"^([a-f0-9]{8})\.execute-api\." + re.escape(_MINISTACK_HOST) + r"(?::\d+)?$"
 )
+# Lambda Function URL: {urlId}.lambda-url.{region}.<anything>[:port]. The stored
+# FunctionUrl carries AWS's own `.on.aws` suffix, so we match any suffix rather
+# than only _MINISTACK_HOST — pointing a proxy or an /etc/hosts entry at the
+# AWS-shaped hostname is the whole point of addressing a function this way.
+_LAMBDA_URL_RE = re.compile(r"^([0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12})\.lambda-url\.[a-z0-9-]+\.")
 # AppSync Events realtime WebSocket: {apiId}.appsync-realtime-api.<anything>[:port].
 _APPSYNC_REALTIME_RE = re.compile(r"^([a-z0-9]+)\.appsync-realtime-api\.")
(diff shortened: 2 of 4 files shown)


Nahuel990:
@liammizrahi I think this is a separate issue from the one originally raised. I thought you were working on that one. Should we close it?

Could you also share a configuration, or a description of whatever you're trying to run, so we can understand what's missing on our side? It would be good to avoid having multiple open issues and PRs covering different problems that don't line up.

liammizrahi:
@Nahuel990 yes, #1030 is a separate thing and I'm not working on it currently.
That one is API Gateway custom domain names, this one is Lambda Function URLs. The only overlap is "host-based routing on 4566".

On what we're running: we have AI agents on Lambda, agentic workflows built with the Vercel AI SDK and LangChain, streaming tokens and tool calls back to the browser while the run is still going. Runs go up to 15 minutes.

We can't put that behind API Gateway. It buffers the whole response and caps out around 29s, so the stream only shows up at the end, if at all. So in AWS the agents sit behind CloudFront → Lambda Function URL with `InvokeMode = RESPONSE_STREAM`, which is the normal setup for this.

The same Terraform applies fine against MiniStack, and `CreateFunctionUrlConfig` hands us back a URL but nothing serves it, so the request ends up in S3 and we get `NoSuchBucket`. That leaves the agents as the one part of our platform we can't run locally at all. That's what this PR is for.

One thing I want to be upfront about: it doesn't make streaming incremental. Responses still come back in one piece, because invokes go through the RIE, which buffers. For us that's still the difference between unreachable and usable. There is a way to get real streaming, but it's a bigger change to how functions are executed, happy to open a separate issue if that's interesting.

PR #1264, opened 2026-08-03
fix(ecs): runner OOM -> reduce number of spawned containers

Lukasdoe:
ECS test container spawning exhausts github workflow runner memory. Most tests don't need the containers, they just waste memory and prolong tests. 

This PR only spawns containers for tests that actually need them, serializing the tests to prevent container sprawl.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,10 @@ jobs:
 
       - name: Start MiniStack
         run: |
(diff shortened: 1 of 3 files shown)


Lukasdoe:
Do you have any comment on why this was closed?

Nahuel990:
@Lukasdoe please get familiar with https://github.com/ministackorg/ministack/blob/main/CONTRIBUTING.md

Nahuel990:
Also the change is introducing code in one service that is not related to AWS creating a new env variable

PR #1271, opened 2026-08-04
fix(stepfunctions): do not expand acronyms in EC2 aws-sdk parameter names

bandle:
### Issue

`_dispatch_aws_sdk_query` runs every parameter name through `_sfn_key_to_api_name`, which expands uppercase acronym runs for the query services that need it (RDS: `DbClusterIdentifier` → `DBClusterIdentifier`). EC2 is the opposite case: its Query serialiser only capitalises a `locationName`'s first letter, so `VpcId` and `EnableDnsHostnames` already ARE the wire names.

```python
>>> from ministack.services.stepfunctions import _sfn_key_to_api_name
>>> _sfn_key_to_api_name("VpcId"), _sfn_key_to_api_name("EnableDnsHostnames")
('VPCId', 'EnableDNSHostnames')
```

No handler reads those, so the parameter silently vanished and the call ran with the argument absent instead of failing. `modifyVpcAttribute` from a state machine reported `SUCCEEDED` and changed nothing; the next state failed with `Ec2.InvalidVpcID.NotFound: The vpc ID '' does not exist`, while the identical boto3 call against the same server worked.

Checked against botocore over every action in `ec2._ACTION_MAP`: **36 input members would be corrupted** by the expansion (`VpcId`, `EnableDnsHostnames`, `KmsKeyId`, `EbsOptimized`, `IamInstanceProfile`, `NetworkAclId`, `PeerVpcId`, `Iops`, `DnsOptions`, …).

### Fix

Skip the expansion for EC2, via an `expand_acronyms` flag on `_convert_params_to_api_names`.

Shorten the override table so it only contains renames, like `Description` → `GroupDescription` 

### Testing

Two tests, split by what they need:

- `test_sfn_aws_sdk_ec2_params_keep_their_sdk_names` — three live executions, each read back from EC2 rather than from the task output, since the task reported success while changing nothing.
- `test_sfn_aws_sdk_ec2_param_names_never_need_the_acronym_expansion` — walks botocore's EC2 model and asserts no member has the expansion for its wire name.

186 passed in `test_stepfunctions.py`.

Behaviour was confirmed on real AWS, then checked against MiniStack with the script below, which runs against either:

| target | result                           |
| --- |----------------------------------|
| MiniStack before | 2/3 (VPCId was handled manually) |
| this branch | 3/3                              |
| real AWS, eu-central-1 | 3/3 — same detail on every line  |

```
                                          before                                              this branch / AWS
modifyVpcAttribute applies EnableDnsHostnames  Ec2.InvalidVpcID.NotFound: The vpc ID '' ...   attribute is set
createSecurityGroup VpcId and Description      in the given VPC, description kept             in the given VPC, description kept
a param without acronyms still applies         narrowed to the one volume                     narrowed to the one volume
```

The middle row passing before the fix is the point: that action is the one the old `"VpcId": "VpcId"` shield covered.

<details>
<summary>check_sfn_ec2_param_names.py</summary>

```python
#!/usr/bin/env python3
"""Check that aws-sdk:ec2 task parameters reach EC2 under the names the caller wrote.

The SDK-key -> wire-name conversion expands uppercase acronym runs, which the other query
services need (RDS: DbClusterIdentifier -> DBClusterIdentifier). EC2 is the opposite case: its
Query serialiser only capitalises the first letter of a locationName, so `VpcId` and
`EnableDnsHostnames` already ARE the wire names. Expanding them to `VPCId` / `EnableDNSHostnames`
sends a parameter no handler reads, so the call runs with the argument absent instead of failing:
modifyVpcAttribute reported success and changed nothing, and the next state failed with
"The vpc ID '' does not exist".

Against MiniStack:
  python check_sfn_ec2_param_names.py --endpoint http://localhost:4566 --region eu-central-1

  python check_sfn_ec2_param_names.py --profile <profile> --region eu-central-1

Creates and deletes one VPC, one security group, the state machines, and (on AWS only) a scoped
IAM role. Pass --keep to leave them behind. Exit status is 0 when every check passes.
"""

import argparse
import json
import sys
import time
import uuid
from contextlib import suppress

import boto3

DUMMY_ROLE = "arn:aws:iam::000000000000:role/sfn-ec2-params-check"
TRUST_POLICY = {
    "Version": "2012-10-17",
    "Statement": [{
        "Effect": "Allow",
        "Principal": {"Service": "states.amazonaws.com"},
        "Action": "sts:AssumeRole",
    }],
}
CHECK_POLICY = {
    "Version": "2012-10-17",
    "Statement": [{
        "Effect": "Allow",
        "Action": ["ec2:ModifyVpcAttribute", "ec2:CreateSecurityGroup", "ec2:DescribeVolumes"],
        "Resource": "*",
    }],
}


def options():
    parser.add_argument("--keep", action="store_true", help="Leave resources behind")
    return parser.parse_args()


def run_task(sfn, name, role_arn, action, params, created, timeout=60):
    """Run one aws-sdk:ec2 task and return (status, parsed output, failure text)."""
    definition = json.dumps({
        "StartAt": "Task",
        "States": {"Task": {
            "Type": "Task",
            "Resource": f"arn:aws:states:::aws-sdk:ec2:{action}",
            "Parameters": params,
            "End": True,
        }},
    })
    arn = sfn.create_state_machine(name=name, definition=definition, roleArn=role_arn)["stateMachineArn"]
    created.append(("state_machine", arn))
    execution_arn = sfn.start_execution(stateMachineArn=arn, input="{}")["executionArn"]
    deadline = time.monotonic() + timeout
    execution = sfn.describe_execution(executionArn=execution_arn)
    while execution["status"] == "RUNNING" and time.monotonic() < deadline:
        time.sleep(0.25)
        execution = sfn.describe_execution(executionArn=execution_arn)
    if execution["status"] != "SUCCEEDED":
        return execution["status"], None, f"{execution.get('error')}: {execution.get('cause')}"
    return "SUCCEEDED", json.loads(execution.get("output") or "{}"), None


def main():
    args = options()
                                aws_secret_access_key="test")
    else:
    client_args = {"endpoint_url": args.endpoint} if args.endpoint else {}
    sfn = session.client("stepfunctions", **client_args)
    ec2 = session.client("ec2", **client_args)
    iam = session.client("iam", **client_args)

    prefix = f"sfn-ec2-params-{uuid.uuid4().hex[:8]}"
    created = []
    results = []

    def check(label, ok, detail):
        print(f"{'PASS' if ok else 'FAIL'}  {label:42} {detail}")

    try:
        vpc_id = ec2.create_vpc(CidrBlock="10.48.0.0/16")["Vpc"]["VpcId"]
        created.append(("vpc", vpc_id))

        if args.endpoint:
            role_arn = DUMMY_ROLE
        else:
            role = iam.create_role(RoleName=f"{prefix}-role",
                                   AssumeRolePolicyDocument=json.dumps(TRUST_POLICY))["Role"]
            iam.put_role_policy(RoleName=role["RoleName"], PolicyName="check",
                                PolicyDocument=json.dumps(CHECK_POLICY))
            created.append(("role", role["RoleName"]))
            role_arn = role["Arn"]
            time.sleep(15)  # IAM propagation before Step Functions accepts the role

        def task(name, action, params):
            return run_task(sfn, f"{prefix}-{name}", role_arn, action, params, created)

        # 1. Two acronym-carrying params in one call: the VPC id says which VPC, the attribute name
        #    says what to set. The effect is read back from EC2 rather than from the task output --
        #    the task reported SUCCEEDED while changing nothing.
        status, out, failure = task("dns", "modifyVpcAttribute",
                                    {"VpcId": vpc_id, "EnableDnsHostnames": {"Value": True}})
        applied = ec2.describe_vpc_attribute(
            VpcId=vpc_id, Attribute="enableDnsHostnames")["EnableDnsHostnames"]["Value"]
        ok = status == "SUCCEEDED" and applied is True
        check("modifyVpcAttribute applies EnableDnsHostnames", ok,
              "attribute is set" if ok
              else (f"task {status}: {failure}" if status != "SUCCEEDED"
                    else f"task SUCCEEDED but attribute is {applied}"))

        # 2. A param that decides where the resource lands, and one that is genuinely renamed:
        #    EC2 calls Description "GroupDescription" on the wire, which no rule derives, so an
        #    override supplies it. Skipping the expansion must not skip the overrides with it --
        #    without them the description silently becomes the group name and the call still
        #    succeeds. A dropped VpcId is equally quiet: the group lands in the default VPC.
        status, out, failure = task("group", "createSecurityGroup",
                                    {"GroupName": prefix, "Description": "param name check",
                                     "VpcId": vpc_id})
        group_id = (out or {}).get("GroupId")
        landed = description = None
        if group_id:
            created.append(("group", group_id))
            group = ec2.describe_security_groups(GroupIds=[group_id])["SecurityGroups"][0]
            landed, description = group["VpcId"], group.get("Description")
        ok = status == "SUCCEEDED" and landed == vpc_id and description == "param name check"
        check("createSecurityGroup VpcId and Description", ok,
              "in the given VPC, description kept" if ok
              else (f"task {status}: {failure}" if status != "SUCCEEDED"
                    else f"vpc={landed} description={description!r}"))

        # 3. A call whose params carry no acronym at all, to show the conversion is not simply off
        #    for everything: describeVolumes still narrows to the volume it was given.
        zone = ec2.describe_availability_zones()["AvailabilityZones"][0]["ZoneName"]
        volume_id = ec2.create_volume(AvailabilityZone=zone, Size=1, VolumeType="gp3")["VolumeId"]
        created.append(("volume", volume_id))
        status, out, failure = task("volumes", "describeVolumes", {"VolumeIds": [volume_id]})
        volumes = (out or {}).get("Volumes") or []
        ok = (status == "SUCCEEDED" and len(volumes) == 1
              and volumes[0].get("VolumeId") == volume_id)
        check("a param without acronyms still applies", ok,
              "narrowed to the one volume" if ok
              else (f"task {status}: {failure}" if status != "SUCCEEDED"
                    else f"{len(volumes)} volume(s): {[v.get('VolumeId') for v in volumes][:3]}"))
    finally:
        if not args.keep:
            for kind, ident in reversed(created):
                with suppress(ClientError):
                    if kind == "state_machine":
                        sfn.delete_state_machine(stateMachineArn=ident)
                    elif kind == "group":
                        ec2.delete_security_group(GroupId=ident)
                    elif kind == "volume":
                        ec2.delete_volume(VolumeId=ident)
                    elif kind == "vpc":
                        ec2.delete_vpc(VpcId=ident)
                    elif kind == "role":
                        iam.delete_role_policy(RoleName=ident, PolicyName="check")
                        iam.delete_role(RoleName=ident)

    return 1 if failed else 0


if __name__ == "__main__":
    try:
        sys.exit(main())
    except (ClientError, TimeoutError) as exc:
        sys.exit(2)
```

</details>


bandle:
I pushed part of our dev commits during rebasing with the remote. to keep this clean i opened a new PR

PR #1273, opened 2026-08-04
CodeBuild: optionally execute builds via the official local agent

igorgawrys1:
## What

Adds opt-in real build execution to the existing CodeBuild service, behind `MINISTACK_CODEBUILD_EXECUTE=1`. Default behaviour is unchanged: without the flag, builds stay metadata-only and complete immediately with `SUCCEEDED`, exactly as today.

Scoping issue: #1274.

This is not a new service (one existing file changed) and adds no dependency: `docker>=7.0.0` is already in `requirements.txt` and already used by `ecs.py`, `elasticache.py`, `glue.py`, and `mwaa.py`.

## Why

`StartBuild` returns a build that has already succeeded, so a CI pipeline rehearsed against MiniStack reports a pass without a single phase having run. A buildspec that fails on real AWS still looks green locally, which is worse than having no emulator for that use case - you have to notice that no containers were ever created to realise nothing happened.

Concretely: I use MiniStack to rehearse a CodeBuild project (Terraform-managed, `privileged_mode`, a buildspec that boots a Postgres + API + webapp stack and runs a browser test suite) before pushing to AWS. With this flag the local run reproduces the AWS run - same phases, same failures, same exit status.

## How

The buildspec is handed to the **official AWS CodeBuild local agent** (`public.ecr.aws/codebuild/local-builds`) rather than reimplementing an executor here, so phase semantics come from AWS's own tooling. The agent runs the phases in the project's `environment.image`.

- `StartBuild` returns `IN_PROGRESS`, then a worker thread follows the agent container.
- The agent's `Phase complete: <PHASE> State: <STATUS>` lines become `phases` entries, so `BatchGetBuilds` reflects progress while the build runs.
- Container exit status maps to `SUCCEEDED` / `FAILED`; `FAULT` when Docker is unreachable or the agent cannot start.
- `environment.environmentVariables`, `privilegedMode`, and `CODEBUILD_BUILD_ID` / `_ARN` / `_NUMBER` / `_INITIATOR` reach the build.
- Build output lands in CloudWatch Logs. The build record already advertised `logs.groupName` / `logs.streamName` but nothing ever wrote there, so an executed build now streams the agent output into that group and stream (same in-process idiom `lambda_svc.py` uses), and `aws logs tail /aws/codebuild/<project>` reads a real build log back.
- `StopBuild` force-removes the agent container for that build, looked up by label, before marking the record `STOPPED`.
- `timeoutInMinutes` is enforced: the build is stopped and its current phase reported `TIMED_OUT`, so a hung build cannot leave a container and a thread behind forever.
- A build's own AWS calls reach MiniStack. `AWS_ENDPOINT_URL` (resolved from the default bridge gateway) plus placeholder credentials are injected unless the project declares them, so `aws s3 cp` inside a build hits this emulator rather than real AWS.
- With `PERSIST_STATE=1`, a build restored as `IN_PROGRESS` is reported `FAULT`: neither its container nor its worker thread survives a restart, so it would otherwise poll as running forever.
- The worker runs inside `contextvars.copy_context()` so it sees the request's account/region scoping - without it the stores resolve to a different scope and the build record is never updated.

One drive-by fix: `_make_build_record` computed `buildNumber` via `b["projectName"]` across every stored build, which raises `KeyError` on a persisted record that lacks the key. It now uses `.get()`. My restore test hit this.

Configuration:

| variable | default | purpose |
|---|---|---|
| `MINISTACK_CODEBUILD_EXECUTE` | `0` | opt in to real execution |
| `MINISTACK_CODEBUILD_AGENT_IMAGE` | `public.ecr.aws/codebuild/local-builds:latest` | agent image |
| `MINISTACK_CODEBUILD_WORKSPACE` | `/tmp/ministack-codebuild` | per-build source/artifacts/env dirs |
| `MINISTACK_CODEBUILD_SOURCE_PATH` | *(unset)* | fixed source path for buildspecs that start nested containers bind-mounting `CODEBUILD_SRC_DIR`; builds then share it and must not overlap |
| `MINISTACK_CODEBUILD_AWS_ENDPOINT` | *(bridge gateway)* | address builds reach MiniStack on; set empty to inject none |
| `MINISTACK_CODEBUILD_MAX_LOG_EVENTS` | `20000` | cap on the in-memory log stream; `0` keeps everything |

```bash
docker run -p 4566:4566 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v /tmp/ministack-codebuild:/tmp/ministack-codebuild \
  -e MINISTACK_CODEBUILD_EXECUTE=1 ministackorg/ministack
```

`WORKSPACE` must resolve to the same path on the host and in the MiniStack container, because the agent bind-mounts it into the build container.

## Out of scope

Deliberately not included, to keep this reviewable: batch builds (`StartBuildBatch`), reports / test-report groups, and source download from a real provider - only the project's inline `source.buildspec` is executed, so a `GITHUB` / `S3` source is not fetched.

Two of these are limited by the agent rather than by scope, which is worth knowing before anyone asks for them. The agent's compose template (`/LocalBuild/agent-resources/docker-compose-mount-src-dir.yml`) exposes no mount for a cache directory, and keeps build output in a Docker volume at `/codebuild/output` rather than a host path, so **build caching** and **artifact upload to `artifacts.location`** cannot be wired up from the outside without either patching the agent or reimplementing the executor. Likewise `StopBuild` removes the agent container, but a build container the agent has already started is a sibling of it, so it can outlive the stop.

## Tests

`tests/test_codebuild_execution.py` - 12 tests, no Docker required (the client is faked, since the point is the bookkeeping around the agent rather than Docker itself):

- the default path stays metadata-only and never touches Docker
- agent log lines become phase records; the build image, privileged mode, buildspec file and env file are what the agent receives
- agent output is written to the build's CloudWatch Logs group and stream, and the stream is capped
- a build past its timeout is stopped and reported `TIMED_OUT` / `FAILED`; `timeoutInMinutes` is read from the project and falls back to 60
- the env file points the build at MiniStack, and a project-declared `AWS_ENDPOINT_URL` wins
- a restored in-flight build is reported `FAULT` rather than `IN_PROGRESS`
- a non-zero exit maps to `FAILED`, an empty buildspec fails before the agent starts, an unreachable Docker maps to `FAULT`

```
$ python -m pytest tests/test_codebuild_execution.py -q
12 passed
```

Verified end-to-end as well: a real project with `privileged_mode` and a multi-container buildspec runs to completion through `aws codebuild start-build --endpoint-url http://localhost:4566`, with `BatchGetBuilds` tracking phases live and returning the true outcome.


igorgawrys1:
Good catch, this was a real race and it is fixed in 638be2f.

`_stop_build` removed the container while the worker was still following its log stream. The stream then ended, `container.wait()` raised on the removed container, and the `except` branch called `_finish_build(build, "FAULT")` straight over the requested `STOPPED`.

The intent is now recorded before the container is touched, and the worker yields to it:

- `_stop_build` calls `_mark_stopped(build_id)` before removing the container.
- The log loop breaks on `_stop_requested()`, so no phase records are appended after a stop.
- Both completion paths (normal and exception) check the stop first and call `_finish_build(build, "STOPPED")`.
- `_clear_stopped()` runs in `finally`, and `reset()` clears the registry, so nothing leaks between builds or tests.

Setting `STOPPED` explicitly rather than just leaving the record alone also closes the ordering variant where a stop lands between `StartBuild` and the worker's first write - the worker sets `IN_PROGRESS` early, so "keep what is there" would have left a stopped build reporting `IN_PROGRESS`.

Covered by `test_stopped_build_is_not_overwritten_by_the_worker`, which drives exactly the failing sequence: stop the build, then let the worker run against a container that raises on both `logs()` and `wait()` the way a removed container does, and assert the outcome is still `STOPPED` and the registry is empty afterwards. 13 tests pass.

Nahuel990:
Thanks for this @igorgawrys1, the shape looks good and it ran clean... One thing before merge: this should be self-contained on a single flag. I checked each new var against real AWS and only one is a real choice:

- MINISTACK_CODEBUILD_EXECUTE stays. 1 runs the container, 0 (default) keeps it metadata-only.
- AGENT_IMAGE: public.ecr.aws/codebuild/local-builds:latest is the fixed image AWS's own codebuild_build.sh uses, so hardcode it. (and remove the env var)
- AWS_ENDPOINT: you already derive it from the bridge gateway, so drop the override and let it auto-derive (remove the env var)
- WORKSPACE and SOURCE_PATH: AWS builds from the caller's PWD and the agent sets CODEBUILD_SRC_DIR itself, so these are internal scratch paths, not knobs. Make them constants (and remove the env var)
- MAX_LOG_EVENTS: no AWS equivalent, please review parity with Lambda behaviour and remove the env var (and the logic behind) 

Also please move the execution tests into the existing tests/test_codebuild.py instead of a new file, since it is the same module

igorgawrys1:
Thanks for the review - trimmed it down to just the one flag:
- AGENT_IMAGE and WORKSPACE are now fixed constants (public.ecr.aws/codebuild/local-builds:latest and /tmp/ministack-codebuild), no env var.
- SOURCE_PATH is gone entirely - source_dir is always workdir/src now.
- AWS_ENDPOINT override is gone - it always auto-derives from the bridge gateway.
- MAX_LOG_EVENTS is gone along with the capping logic - checked parity with Lambda's _emit_lambda_logs, which appends to the log stream with no cap at all, so this brings CodeBuild in line rather than leaving a gap.
- Moved the execution tests into tests/test_codebuild.py, tests/test_codebuild_execution.py is deleted.
- MINISTACK_CODEBUILD_EXECUTE is the only knob left, as you asked. Also rebased onto latest main to clear the merge conflict. Ready for another look.

Nahuel990 (approved):

PR #1276, opened 2026-08-04
Prevent Docker operations from blocking the event loop

Areson:
## Why
Blocking Docker API calls in RDS, ECS, EKS, ElastiCache, and OpenSearch currently run on MiniStack's asyncio event loop. A slow container operation can therefore freeze unrelated services and delay timing-sensitive work such as Glue crawler transitions.

## What
- Offload the five Docker-heavy service routers while preserving per-service request serialization and cancellation safety
- Make reset, CloudFormation, Step Functions, EventBridge, S3 notifications, RDS Data, and tagging explicit participants in the resulting concurrency model
- Preserve lazy service loading and caller account/region context across HTTP, scheduler, and execution-thread dispatch paths
- Add concurrency regressions for loop responsiveness, reset isolation, callback traffic, task lifecycles, and cross-service imports

## Behavior changes and risk
Slow Docker work no longer blocks unrelated gateway requests; requests within each affected service remain serialized. Reset now drains tracked CloudFormation work and synchronous Step Functions actions before taking the request barrier and service locks; mutating CloudFormation requests and new synchronous Step Functions actions may receive HTTP 503 while that admission is closed. During the CloudFormation drain phase, an ordinary request can still complete and then be cleanly wiped by the reset; blocking all ordinary traffic earlier would deadlock custom resources whose Lambda callbacks make nested SDK requests. Background Step Functions `StartExecution` threads retain their existing main-equivalent reset race. The change replaces implicit event-loop serialization with explicit locks and lifecycle tracking, so the primary risk is a missed non-HTTP or cross-service state consumer; the added regression matrix targets those boundaries.

## Validation
- Complete 20-file MiniStack matrix: 2,044 passed, 48 skipped
- Scheduler, CloudFormation, Step Functions, and app lane: 440 passed, 1 skipped
- Event-loop concurrency suite: 92 passed
- A downstream acceptance suite that runs MiniStack as its test backend passed against an image built from this exact commit
- Ruff, Python compilation, and diff checks passed for all changed Python files
- A same-machine main comparison was inconclusive because Docker/EKS cleanup variance dominated the sample; no aggregate throughput claim is made. The validated behavior is isolation: unrelated gateway requests remain responsive while a Docker worker is blocked.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -21,6 +21,7 @@ import sys
 import tempfile
 import time
 import uuid
+from contextlib import AsyncExitStack
 from urllib.parse import parse_qs, unquote
 
 _MINISTACK_HOST = os.environ.get("MINISTACK_HOST", "localhost")
@@ -161,6 +162,7 @@ _NON_S3_VHOST_NAMES = frozenset({
     "inspector2",
 })
 
+from ministack.core.concurrency import LoopLocal, run_in_thread_to_completion
 from ministack.core.hypercorn_compat import install as _install_hypercorn_compat
 from ministack.core.persistence import PERSIST_STATE, load_state, save_all
 from ministack.core.responses import _12_DIGIT_RE, set_request_account_id, set_request_region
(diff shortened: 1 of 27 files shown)


Areson:
🤖 Closing while we iterate further locally; a fresh PR will follow when the change is finalized.

PR #1277, opened 2026-08-04
Prevent Docker operations from blocking the event loop

Areson:
## Why
Blocking Docker API calls in RDS, ECS, EKS, ElastiCache, and OpenSearch currently run on MiniStack's asyncio event loop. A slow container operation can therefore freeze unrelated services and delay timing-sensitive work such as Glue crawler transitions.

## What
- Offload the five Docker-heavy service routers while preserving per-service request serialization and cancellation safety
- Make reset, CloudFormation, Step Functions, EventBridge, S3 notifications, RDS Data, and tagging explicit participants in the resulting concurrency model
- Preserve lazy service loading and caller account/region context across HTTP, scheduler, and execution-thread dispatch paths
- Add concurrency regressions for loop responsiveness, reset isolation, callback traffic, task lifecycles, and cross-service imports

## Behavior changes and risk
Slow Docker work no longer blocks unrelated gateway requests; requests within each affected service remain serialized. Reset now drains tracked CloudFormation work and synchronous Step Functions actions before taking the request barrier and service locks; mutating CloudFormation requests and new synchronous Step Functions actions may receive HTTP 503 while that admission is closed. During the CloudFormation drain phase, an ordinary request can still complete and then be cleanly wiped by the reset; blocking all ordinary traffic earlier would deadlock custom resources whose Lambda callbacks make nested SDK requests. Background Step Functions `StartExecution` threads retain their existing main-equivalent reset race. When the `SERVICES` allow-list excludes `rds-data`, the RDS Data endpoint now returns the standard unsupported-service 400 instead of being served. The change replaces implicit event-loop serialization with explicit locks and lifecycle tracking, so the primary risk is a missed non-HTTP or cross-service state consumer; the added regression matrix targets those boundaries.

## Validation
- Complete 20-file MiniStack matrix: 2,044 passed, 48 skipped
- Scheduler, CloudFormation, Step Functions, and app lane: 440 passed, 1 skipped
- Full ECS service lane: 41 passed
- Event-loop concurrency suite: 92 passed
- A downstream acceptance suite that runs MiniStack as its test backend passed against an image built from this exact commit
- Ruff, Python compilation, and diff checks passed for all changed Python files

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -21,6 +21,7 @@ import sys
 import tempfile
 import time
 import uuid
+from contextlib import AsyncExitStack
 from urllib.parse import parse_qs, unquote
 
 _MINISTACK_HOST = os.environ.get("MINISTACK_HOST", "localhost")
@@ -161,6 +162,7 @@ _NON_S3_VHOST_NAMES = frozenset({
     "inspector2",
 })
 
+from ministack.core.concurrency import LoopLocal, run_in_thread_to_completion
 from ministack.core.hypercorn_compat import install as _install_hypercorn_compat
 from ministack.core.persistence import PERSIST_STATE, load_state, save_all
 from ministack.core.responses import _12_DIGIT_RE, set_request_account_id, set_request_region
@@ -229,6 +231,24 @@ def _get_module(name: str):
             mod = _ErrorModule(name, str(e))
         _loaded_modules[name] = mod
     return mod
+
+
+def _resolve_loaded_service_module(name: str):
+    """Return an already-loaded service module without importing it."""
+    return _loaded_modules.get(name) or sys.modules.get(f"ministack.services.{name}")
+
+
+# Serialized service keys and every module whose first import establishes the
+# admission lock protecting that request. Keep this as the single source of
+# truth for HTTP dispatch and non-HTTP Step Functions SDK integrations.
+_RESET_ADMISSION_OWNER_MODULES = {
+    "rds": ("rds",),
+    "ecs": ("ecs",),
+    "eks": ("eks",),
+    "elasticache": ("elasticache",),
+    "opensearch": ("opensearch",),
+    "rds-data": ("rds_data", "rds"),
+}
 
 
 def _lazy_handler(module_name: str):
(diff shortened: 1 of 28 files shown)


Areson:
@Nahuel990 This is optional but I ran across CI failures a few times due to contention in the event loop. This is a much larger change to decouple it.

Nahuel990:
@Areson I'm away now. This is a 5,000+ line PR, and reading just the statement doesnt provide much context. (asyncio.Lock) and (asyncio.toThread should take care of this if you ever run into this type of problem) I'll carefully review the PR once I'm back. For this type of rearch, please open an issue first

Nahuel990:
@Areson there are several different things with this PR… I see the concurrency is creating a problem (that is also resolving) I was able to reproduce the original error you reported (we had this issue before with one of the lambda containers, where we moved to asyncio.to_thread) I think the best approach for this are targeted fixes per docker call, instead of moving every service router into a worker thread behind its own admission lock. also we're returning 503s with a code string AWS doesnt use.

going a bit deeper on each part:

concurrency.py: I dont think we need this file. we already offload blocking work with plain asyncio.to_thread all over the place (athena, cognito, lambda_svc, apigateway, alb, sns, mwaa, bedrock_runtime, cloudformation). the only thing run_in_thread_to_completion adds on top is deferring cancellation, and thats only needed because youre running the whole handler + the reset guards inside the thread. if we just move the docker calls off the loop, plain asyncio.to_thread is enough and none of this is needed

ecs.py (and the other four) - we want the docker calls off the loop, not the whole handler in a thread. look at athena.py, the blocking call is wrapped in asyncio.to_thread with a comment about not stalling the loop and thats the whole fix, no lock. and where we do need per service serialization we already have asyncio.Lock in a few places (appsync_events, iot, transfer). putting the whole router in a worker doesnt even help inside the service, every request still queues on the per service lock, so a DescribeTasks waits out a slow docker pull exactly like it does today.

stepfunctions: it does no docker and already runs executions in background threads, so it shouldnt be in a docker offload PR at all

cloudformation - same, and I need to correct something I said earlier: the to_thread calls in stacks.py are all behind `if _is_custom_resource(...)` (164, 175, 215, 249, 357), the non custom provisioners still run inline on the loop (169, 174). so a CFN provisioned AWS::RDS::DBInstance does block it, same as the direct API call. that part is real, but it fixes itself once the 5 container services offload their docker calls, it doesnt need an admission layer on top.

503 errors:  _mutation_admission_error() returns ServiceUnavailableException on CreateStack/UpdateStack/DeleteStack/CreateChangeSet/ExecuteChangeSet/DeleteChangeSet, and handle_request returns it on StartSyncExecution/TestState. the 503 status is fine, both services document a common ServiceUnavailable there. the code string isnt: its ServiceUnavailable with no Exception suffix, and neither the cloudformation 2010-05-15 nor the stepfunctions 2016-11-23 model has any ServiceUnavailable* shape. and the string is the smaller half of it, botocore retries on the 503 status itself and not on the code, so every SDK client and terraform will quietly retry with backoff while a reset is running. thats the opposite of what the barrier is for

so instead of a full change that adds coupling to all services, the best way would be wrapping the blocking docker calls in the 5 container services (rds, ecs, eks, elasticache, opensearch) with asyncio.to_thread, keep state on the loop, and add a per service asyncio.Lock only where ordering matters. thats a few dozen lines. no new concurrency framework, no reset barriers, no 503s, and no touching cloudformation or stepfunctions

Areson:
🤖 Closing in favor of #1386, which supersedes this PR and carries its key mechanisms forward (cancellation shield-and-wait, per-call worker threads, and the Step Functions sync-dispatch fix) as part of the broader concurrency synthesis. We have validated the #1386 preview build against our workloads and will follow up with detailed feedback on that PR. Thanks for the discussion and reviews here.

PR #1280, opened 2026-08-04
Fix durable Lambda restore scoping

Areson:
## Why

Durable Lambda executions persist across MiniStack restarts, but the restore path only rebuilt timers and callback indexes for the ambient account. Running executions in non-default accounts could be left stalled after restart, and non-boot-region executions could be re-armed under the wrong region.

## What

- Rebuild durable execution timers and callback indexes across all persisted account scopes during restore.
- Derive each restored execution's account and region from its DurableExecutionArn before re-arming callbacks and timers.
- Scope the callback index by account so identical callback IDs in different accounts remain resolvable after restart.
- Add durable restore coverage for non-default accounts, non-boot regions, malformed ARN fallback, and cross-account callback ID collisions.

## Behavior changes and risk

Running durable executions from non-default accounts and non-boot regions now resume correctly after process restart instead of stalling or resuming under the boot region. Restored callback IDs keep the same public endpoint shape, but lookup is account-scoped to avoid cross-account collisions.

Malformed persisted DurableExecutionArn values now log a warning and fall back to the persisted account scope and the default region during restore. Risk is limited to the durable restore path and callback index scoping.

## Validation

* `uv run --extra dev ruff check ministack/services/lambda_durable.py tests/test_lambda.py`
* `uv run --extra dev pytest tests/test_lambda.py tests/test_persistence.py -q` with local MiniStack server (`546 passed, 2 skipped`)

### DIFF
--- a/ministack/services/lambda_durable.py
+++ b/ministack/services/lambda_durable.py
@@ -19,14 +19,13 @@ DurableExecution ARN format per the docs' Pattern field:
 from __future__ import annotations
 
 import base64
+import contextvars
 import copy
 import json
 import secrets
 import time
 import uuid
 from urllib.parse import unquote
-
-import contextvars
 
 from ministack.core.persistence import load_state
 from ministack.core.responses import (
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1282, opened 2026-08-05
fix(lambda): normalize SDK client stub lookup

### LINKED ISSUE #1281: local executor: @aws-sdk/client-secrets-manager never matches the stub lookup table (keyed `secretsmanager`) — handlers die at init
**Affected component:** Lambda `local` executor (`LAMBDA_EXECUTOR=local`, the default) — the `require('@aws-sdk/client-*')` intercept in `core/lambda_runtime.py`.

**Summary**

The local executor's `require()` intercept stubs out `@aws-sdk/client-*` packages by matching the package name suffix (everything after `@aws-sdk/client-`) against a lookup table. For Secrets Manager, the table key is `"secretsmanager"`, but the real npm package is `@aws-sdk/client-secrets-manager`, whose suffix is `"secrets-manager"` (hyphenated). The lookup never matches, so the intercept falls through to a real `require()`, which fails because the SDK isn't installed in the stub runtime. Any handler that imports the Secrets Manager client dies at cold start.

For comparison, `@aws-sdk/client-dynamodb` works fine — its suffix (`dynamodb`) happens to equal its table key, which is what makes this confusing to diagnose.

**Repro**

```js
// index.mjs
import { SecretsManagerClient, GetSecretValueCommand } from "@aws-sdk/client-secrets-manager";

export const handler = async () => {
  const client = new SecretsManagerClient({});
  await client.send(new GetSecretValueCommand({ SecretId: "whatever" }));
  return { ok: true };
};
```

```
aws lambda create-function --function-name repro \
  --runtime nodejs20.x --handler index.handler \
  --zip-file fileb://function.zip --role arn:aws:iam::000000000000:role/lambda-role \
  --endpoint-url http://localhost:4566

aws lambda invoke --function-name repro out.json --endpoint-url http://localhost:4566
cat out.json
```

Result:
```json
{"errorMessage":"Worker init failed: Cannot find module '@aws-sdk/client-secrets-manager'\nRequire stack:\n- /tmp/ministack-lambda-.../code/index.js","errorType":"RuntimeError"}
```

Swap the import for `@aws-sdk/client-dynamodb` and the same handler shape works, which is what makes the Secrets Manager failure surprising.

**Root cause**

In `core/lambda_runtime.py` (current `main`, commit `dfc10119`):

- The lookup table `_JSON_RPC_TARGETS` keys Secrets Manager as `"secretsmanager"` (no hyphen), e.g.:
  ```
  "secretsmanager":              "secretsmanager",
  ```
- The `require()` intercept extracts the match key from the real package name via `_SDK_CLIENT_RE = /^@aws-sdk\/client-(.+)$/`, so for `@aws-sdk/client-secrets-manager` the captured key is `"secrets-manager"` (hyphenated, matching the actual npm package name) — which never equals the table's `"secretsmanager"`.
- When the lookup misses, `prefix` is `undefined`, so the generic-stub branch is skipped and the intercept falls through to a real `require()` call that is *not* wrapped in a try/catch, so it throws and propagates the raw "Cannot find module" error to the worker's init path.

**Suggested fix**

1. Rename the table key to `"secrets-manager"` to match the real package suffix. Better: normalize both sides of the comparison (e.g. strip hyphens from the captured suffix before the table lookup, or store table keys already stripped of hyphens) so this entire class of bug — any future `@aws-sdk/client-*` package whose hyphenation doesn't match its table key — can't recur. A few other entries in the same table look like they'd have the identical problem for any future additions (the existing ones happen to be correct, e.g. `cognito-identity-provider`, `dynamodb-streams`, but there's no structural guarantee).
2. Separately: some `@aws-sdk/client-*` packages have no stub entry at all in the local executor (e.g. `client-s3`, `client-sesv2`). For those, a clearer error would help a lot — something like "MiniStack local executor has no stub for `@aws-sdk/client-x`; use `LAMBDA_EXECUTOR=docker`" — instead of surfacing the raw Node module-resolution failure, which looks like a packaging/bundling problem on the user's side rather than an emulator limitation.

**Environment**

- Image: `ministackorg/ministack:latest`, resolved version 1.4.11, edition `light`
- Found while evaluating MiniStack for local Lambda + CDK-deployed stacks (2026-08-04)

roshie548:
## Summary
- key JSON-RPC SDK stubs by exact full module specifier, including the intentional `events` and `logs` aliases
- preserve bundled-module precedence and the actionable local-executor error for missing stubs
- cover `client-secrets-manager` plus an independent hardcoded contract fixture for canonical and alias package names

Exact matching was chosen over hyphen normalization because it is simpler and prevents distinct package names from collapsing onto the same lookup key.

## Tests
- `.venv/bin/pytest tests/test_lambda_runtime_worker.py -v` (17 passed)
- `.venv/bin/ruff check ministack/core/lambda_runtime.py tests/test_lambda_runtime_worker.py`

Fixes #1281

### DIFF
--- a/ministack/core/lambda_runtime.py
+++ b/ministack/core/lambda_runtime.py
@@ -438,41 +438,41 @@ fs.write = function(fd, ...args) {
   // Most AWS SDK v3 packages use awsJson1.x: POST / with X-Amz-Target header.
   // Ministack's router maps target prefixes to service modules.
   const _JSON_RPC_TARGETS = {
-    // JSON-RPC (awsJson1.x) services — keyed by @aws-sdk/client-{key} suffix.
+    // JSON-RPC (awsJson1.x) services — keyed by full module specifier.
     // Target prefixes match Ministack's router.py SERVICE_PATTERNS target_prefixes.
-    "ssm":                         "AmazonSSM",
-    "sfn":                         "AWSStepFunctions",
+    "@aws-sdk/client-ssm":                         "AmazonSSM",
+    "@aws-sdk/client-sfn":                         "AWSStepFunctions",
     // sts, sns: query protocol — @aws-sdk/client-{sts,sns} sends Action= form-encoded POST
     // cloudwatch: smithy-rpc-v2-cbor — @aws-sdk/client-cloudwatch sends path-based requests
     // All three are handled by Ministack's native query/path routing when the real SDK is present
-    "cloudwatch-logs":             "Logs_20140328",
-    "logs":                        "Logs_20140328",
-    "secretsmanager":              "secretsmanager",
-    "events":                      "AmazonEventBridge",
-    "eventbridge":                 "AmazonEventBridge",
-    "kinesis":                     "Kinesis_20131202",
-    "ecs":                         "AmazonEC2ContainerServiceV20141113",
-    "dynamodb":                    "DynamoDB_20120810",
-    "dynamodb-streams":            "DynamoDBStreams_20120810",
-    "sqs":                         "AmazonSQS",
-    "glue":                        "AWSGlue",
-    "athena":                      "AmazonAthena",
-    "firehose":                    "Firehose_20150804",
-    "cognito-identity-provider":   "AWSCognitoIdentityProviderService",
-    "cognito-identity":            "AWSCognitoIdentityService",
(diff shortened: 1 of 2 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1286, opened 2026-08-05
feat(logs): implement GetLogRecord and Insights @ptr result rows

ovsteenb:
## Summary

Adds CloudWatch Logs `GetLogRecord` and makes Insights queries return real rows with opaque `@ptr` values, so callers can round-trip pointer resolution the way AWS does.

**Use case:** tools that expand surrounding context from Logs Insights query artifacts (e.g. resolving `@ptr` via `GetLogRecord(unmask=False)` after `StartQuery` / `GetQueryResults`) currently cannot do that against MiniStack because Insights was stubbed to empty `results` and `GetLogRecord` was unimplemented. LocalStack also does not reliably cover this path.

**Scope (in):**
- Assign opaque pointers on `PutLogEvents` and index full transformed records
- `StartQuery` / `GetQueryResults` scan ingested events in the time window and return AWS-shaped field/value rows (`@timestamp`, `@message`, `@logStream`, `@log`, `@ptr`, plus simple `fields` / `limit` parsing)
- `GetLogRecord` resolves pointers (`unmask` accepted; masking not implemented)
- Persist `_log_records` with the rest of Logs state; clear on `reset()` / group/stream delete
- README table + CHANGELOG + tests

**Scope (out):**
- Full CWLI grammar (`filter`, `stats`, `sort`, etc.)
- Field masking / `logs:Unmask`
- Async query lifecycle beyond immediate `Complete`

## Test plan

- [x] `ruff check ministack/services/cloudwatch_logs.py tests/test_cloudwatch_logs.py`
- [x] `pytest tests/test_cloudwatch_logs.py -o addopts=` against a local MiniStack (`MINISTACK_ENDPOINT=http://127.0.0.1:4566`) — 67 passed
- [ ] Maintainer CI green

## CONTRIBUTING checklist

- [x] Existing service (`cloudwatch_logs.py`) — not a new service (no pre-issue required per CONTRIBUTING)
- [x] Tests added and passing
- [x] Linting passes (`ruff check`)
- [x] README service table updated
- [x] CHANGELOG.md entry added

PR #1287, opened 2026-08-05
feat(logs): GetLogRecord, Insights @ptr rows, and StartLiveTail

ovsteenb:
## Summary
- Implement `GetLogRecord` and make Insights `StartQuery` / `GetQueryResults` return real AWS-shaped rows with opaque `@ptr` values assigned at `PutLogEvents`.
- Implement `StartLiveTail` as a long-lived `application/vnd.amazon.eventstream` that stays open until client disconnect: `initial-response` → `sessionStart` → `sessionUpdate` frames fed by matching concurrent `PutLogEvents`.
- Return `DescribeLogGroups.logGroupArn` (no trailing `:*`) alongside `arn` for StartLiveTail-safe identifiers.
- Add a small ASGI `StreamingResponse` path (`more_body`, no `Content-Length`) used by Live Tail.

**Use case:** tools that resolve Insights `@ptr` via `GetLogRecord` and/or integration-test Live Tail (open tail → `PutLogEvents` → assert frames → disconnect) against MiniStack.

**Scope (in):**
- Opaque pointers on `PutLogEvents`; Insights field/value rows; `GetLogRecord` resolution (`unmask` accepted; masking not implemented)
- Push Live Tail until disconnect; idle empty updates via `MINISTACK_LIVE_TAIL_IDLE_SECONDS` (default 1s); `inject_host_prefix=False` needed for custom endpoints
- README + CHANGELOG + tests

**Scope (out):**
- Full CWLI grammar; field masking
- AWS 3-hour session ceiling / exact 1 Hz batching parity

Supersedes #1286.

## Test plan
- [x] `ruff check` on touched files
- [x] `pytest tests/test_cloudwatch_logs.py -o addopts=` against local MiniStack — 70 passed (includes concurrent PutLogEvents → Live Tail)
- [ ] Maintainer CI green

## CONTRIBUTING checklist
- [x] Existing service (`cloudwatch_logs.py`) — not a new service
- [x] Tests added and passing
- [x] Linting passes


Nahuel990:
Thanks @ovsteenb, this is a strong implementation. The eventstream framing, the payload shapes, and the async session handling all check out against AWS, and the single-loop fanout is clean.

One change before it can go in: drop the two env vars and hardcode them, because AWS already pins both values.

- MINISTACK_LIVE_TAIL_IDLE_SECONDS: AWS sends a LiveTailSessionUpdate every second, empty on an idle second. So make the heartbeat a constant 1.0, not tunable.
- MINISTACK_LIVE_TAIL_QUEUE_MAX: AWS buffers up to 10 LiveTailSessionUpdate events before dropping the oldest, so the bound should be a constant 10, not 1000. The current default changes when events drop, so this is a real behavior fix, not just config removal.

MiniStack matches AWS behavior rather than exposing configs for it. Please also drop the knob mentions from the CHANGELOG and README.

ovsteenb:
@Nahuel990 Addressed in the latest push:

- Removed `MINISTACK_LIVE_TAIL_IDLE_SECONDS` / `MINISTACK_LIVE_TAIL_QUEUE_MAX`
- Idle heartbeat hardcoded to **1.0s**
- Queue depth hardcoded to **10**; on overflow drop the oldest buffered `sessionUpdate` and set `sampled`
- README / CHANGELOG updated (no env knobs)

Hope the changes help. cloudwatch is often an  afterthought :)

Nahuel990 (approved):

Nahuel990 (approved):

PR #1297, opened 2026-08-06
fix(rds): return DBInstanceAlreadyExists wire code for duplicate DB instances

Kiran01bm:
## Summary

Duplicate DB-instance responses from `CreateDBInstance`, `CreateDBInstanceReadReplica`, and `RestoreDBInstanceFromDBSnapshot` returned the wire code `DBInstanceAlreadyExistsFault`. Real AWS returns `DBInstanceAlreadyExists` — the `Fault` suffix exists only in the SDK shape name, not on the wire (instance-level codes drop it; cluster-level codes like `DBClusterAlreadyExistsFault` keep it).

## What

- `ministack/services/rds.py`: the three duplicate-instance error responses now emit `DBInstanceAlreadyExists`.
- `tests/test_rds.py`: new regression test `test_rds_create_duplicate_instance` asserting the exact wire code on duplicate `CreateDBInstance`; the two existing read-replica and snapshot-restore duplicate assertions updated to the correct code.
- `CHANGELOG.md`: `[Unreleased] / Fixed` entry.

## Why

SDKs match the exact wire string to produce their typed error. With the suffixed code, aws-sdk-go-v2 deserializes the response as a generic `smithy.APIError` instead of the typed `*types.DBInstanceAlreadyExistsFault`, so `errors.As`-based duplicate handling that works against real AWS silently fails against MiniStack.

Reference: the AWS RDS API docs list `DBInstanceAlreadyExists` (HTTP 400) as the error code for these operations.

This deliberately reverts the v1.1.18 changelog entry ("`DBInstanceAlreadyExists` corrected to `DBInstanceAlreadyExistsFault` matching AWS error codes"), which mistook the SDK exception *shape name* for the *wire code*. The exact-string test assertions added here pin the correct direction against a future re-flip.

## Before / after

```
Duplicate CreateDBInstance against MiniStack:

before:  <Error><Code>DBInstanceAlreadyExistsFault</Code>...   -> generic APIError in SDKs
after:   <Error><Code>DBInstanceAlreadyExists</Code>...        -> typed DBInstanceAlreadyExistsFault
```

## Testing

- `pytest tests/test_rds.py -m "not serial"` — 180 passed, 5 skipped (source-mounted container on the v1.4.13 image).
- `ruff check ministack/services/rds.py tests/test_rds.py` — clean.

## Note: boto3 docs vs wire code

The boto3 docs list the exception *shape name* (`DBInstanceAlreadyExistsFault`) — the class you catch — not the wire code. Each error shape in AWS's service model carries an `error.code` field, which is the string serialized in `<Error><Code>` and the key every SDK uses to map a response onto its typed exception. From botocore's `data/rds/2014-10-31/service-2.json`:

```
DBInstanceAlreadyExistsFault -> {"code": "DBInstanceAlreadyExists", "httpStatusCode": 400}
DBClusterAlreadyExistsFault  -> {"code": "DBClusterAlreadyExistsFault", "httpStatusCode": 400}
DBSnapshotAlreadyExistsFault -> {"code": "DBSnapshotAlreadyExists", "httpStatusCode": 400}
```

Cluster-level codes genuinely keep the `Fault` suffix on the wire while instance/snapshot codes drop it, which is why this PR changes only the instance-level codes rather than doing a blanket suffix strip.


Nahuel990 (approved):

Nahuel990:
Thanks! Sorry for the delay Github is having some issues

PR #1305, opened 2026-08-06
feat(kms): add HMAC keys, GenerateMac and VerifyMac

nafdev:
### CreateKey
- Supports the four HMAC key specs (`HMAC_224/256/384/512`).
- HMAC keys require `KeyUsage=GENERATE_VERIFY_MAC`; that usage is rejected for every non-HMAC spec, and an HMAC spec is rejected with any other usage.
- `EncryptionAlgorithms` / `SigningAlgorithms` are omitted from HMAC key metadata rather than returned empty.

### GenerateMac / VerifyMac
- Real RFC 2104 HMAC via the stdlib `hmac` module. `VerifyMac` returns `MacValid: true`, or fails with `KMSInvalidMacException`.
- Both validate the `MacAlgorithm` enum, reject an algorithm that mismatches the key spec, enforce the documented 1–4096 byte message limit (checked before base64 decode), bound the Mac blob before decoding, accept key IDs / ARNs / aliases, respect `Disabled` and `PendingDeletion`, and implement `DryRun`.
- Validation order matches AWS, with `DryRun` as the final check: parameters → key resolution → key state → key usage → (VerifyMac) MAC comparison → DryRun. So a disabled key returns `DisabledException` and an invalid MAC return `KMSInvalidMacException` even when `DryRun` is set, because the request have failed in either case anyway.

### Complementary handling
- HMAC keys rejected with `InvalidKeyUsageException` by `Encrypt`, `Decrypt`, `Sign`, `Verify` and `GenerateDataKey*`.
- `EnableKeyRotation` returns `UnsupportedOperationException` — AWS does not rotate HMAC keys. `DisableKeyRotation` and `GetKeyRotationStatus` accept them and report rotation off.
- HMAC key material round-trips through persisted state.
- `_is_hmac_key` keys off the key spec plus presence of HMAC material, never `KeyUsage` alone, so a symmetric key persisted by an earlier build carrying `GENERATE_VERIFY_MAC` isn't misread as HMAC.


Nahuel990:
@nafdev we are currently affected by Github issues so cannot run CI, I see the description you mention KMSInvalidMacException , but in the code you return DryRunOperationException... which one is correct? Are you validating against AWS?

nafdev:
hi @Nahuel990, I have updated the description and made some rebasing changes. I have tested against AWS though I tried and found a few edgecases and aligned those too (DisableKeyRotation and GetKeyRotationStatus)

Nahuel990:
@nafdev I'll merge it but I checked myself against AWS and found a 2 mismatches in error codes so I wil apply the changes directly before release. Thanks

PR #1313, opened 2026-08-07
feat(eventbridge): API-destination dispatch, DLQs for every target type, archive/replay AWS parity

t-rech:
# EventBridge: API-destination dispatch, DLQs for every target type, archive/replay parity

This PR takes EventBridge from control-plane stubs to observable-behavior parity for three pipelines: **API destinations invoked over HTTP**, **dead-letter queues on every target type**, and **archive → replay**. It also carries the CloudWatch Logs support work those pipelines depend on (vended log delivery, resource policies).

Every behavioral claim below is backed by AWS public documentation (linked inline, with the documented fact quoted) or, where the public docs are silent, by live verification against the AWS API (marked **live-verified**).

---

## 1. Dead-letter queues — every target type

Evidence: [Using dead-letter queues to process undelivered events](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rule-dlq.html) (eb-rule-dlq), [DeadLetterConfig API reference](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_DeadLetterConfig.html), [How EventBridge retries delivering events](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rule-retry-policy.html).

| Implemented behavior | Documented fact |
|---|---|
| DLQ context resolved once per invocation and threaded into **all** dispatchers (Lambda, SQS, SNS, Step Functions, CloudWatch Logs, event-bus, API destination) | eb-rule-dlq scopes DLQs to any rule target: "When you configure a DLQ **for a target of a rule**, EventBridge sends the events with failed invocations to the Amazon SQS queue selected." |
| Missing target resource → `ERROR_CODE=NO_RESOURCE`, delivered to the DLQ **without** retry attempts and **without** `EXHAUSTED_RETRY_CONDITION` | "Some events are dropped or sent to a DLQ **without any retry attempts**. For example, for errors that result from missing permissions to a target, **a target resource that no longer exists** … EventBridge sends these events directly to the target DLQ." |
| Message attributes stamped on every DLQ message: `RULE_ARN`, `TARGET_ARN`, `ERROR_CODE`, `ERROR_MESSAGE`, `RETRY_ATTEMPTS`, plus `EXHAUSTED_RETRY_CONDITION` when the retry policy ran out | eb-rule-dlq: "Each message in your DLQ will include the following custom attributes: `RULE_ARN`, `TARGET_ARN`, `ERROR_CODE` … `ERROR_MESSAGE`, `EXHAUSTED_RETRY_CONDITION` (`MaximumRetryAttempts` \| `MaximumEventAgeInSeconds`), `RETRY_ATTEMPTS`" |
| Error codes used: `NO_RESOURCE`, `ERROR_FROM_TARGET`, `INVALID_JSON`, `TIMEOUT`, `CONNECTION_FAILURE`, `INTERNAL_ERROR` | All six appear in eb-rule-dlq's published sample of DLQ error codes |
| FIFO queues rejected as DLQs (warning + drop) | "Only standard queues are supported. You can't use a FIFO queue for a DLQ in EventBridge." |
| DLQ body is the **original pre-input-selection event envelope**, never the transformed target input | DLQ purpose is redrive of the failed *event* ("retain any events that weren't successfully delivered … and process the events at a later time"); **live-verified** that the body is the original envelope, not the `Input`/`InputPath`/`InputTransformer` result |
| `PutTargets` rejects a `DeadLetterConfig.Arn` that is not an SQS queue ARN (`ValidationException`) | API reference: `DeadLetterConfig.Arn` is "The ARN of the SQS queue specified as the target for the dead-letter queue." Queue existence/FIFO checks stay delivery-time concerns |
| Cross-region DLQ resolution fails naturally at delivery (region-scoped queue lookup) | "The Amazon SQS queue you use must be in the same Region in which you create the rule." |
| **Not modeled (documented in README):** the retry pipeline itself — retryable failures dead-letter immediately with `RETRY_ATTEMPTS=0` and `EXHAUSTED_RETRY_CONDITION=MaximumRetryAttempts` pre-stamped | eb-rule-retry-policy: "By default, EventBridge retries sending the event for 24 hours and up to 185 times with an exponential back off and jitter" — a timing pipeline a local emulator cannot usefully reproduce |
| **Kept as-is:** cross-region target failures drop without dead-lettering | **Live-verified** 2026-07-13 (`RESEARCH/MINISTACK_CROSSREGION_SNS_TARGET_POLICY.md`): AWS records `FailedInvocations` without cross-region delivery |

## 2. Archive

Evidence: [CreateArchive API reference](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_CreateArchive.html), [Creating event archives](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-archive-event.html).

| Implemented behavior | Documented fact |
|---|---|
| Replayed events are **never** captured by an archive (`_archive_event` skips events carrying `ReplayName`) | CreateArchive: "all events are sent to the archive **except replayed events. Replayed events are not sent to an archive.**" |
| `CreateArchive` requires an existing source event bus → `ResourceNotFoundException` | `EventSourceArn` is a required member with ARN pattern `^arn:aws([a-z]\|\-)*:events:…`; `ResourceNotFoundException` ("An entity that you specified does not exist") is a declared error |
| `CreateArchive` validates `EventPattern` is JSON → `InvalidEventPatternException` (previously only `UpdateArchive` checked) | `InvalidEventPatternException` ("The event pattern is not valid") is a declared CreateArchive error |
| `RetentionDays >= 0` enforced server-side (unreachable via boto3, which validates client-side; matters for non-validating SDKs such as the Terraform provider) | "Valid Range: Minimum value of 0. … If set to 0, events are retained indefinitely" |
| Archive name capped at 48 chars (earlier commit on this branch) | "Length Constraints: Minimum length of 1. Maximum length of 48." |
| `SizeBytes` accumulates as events are captured (was pinned to 0) | `SizeBytes` is a documented DescribeArchive response member |
| **Not modeled:** `RetentionDays` expiry sweeps, `CREATING`/`UPDATING` state transitions | Time-based lifecycle with no local observable value; archives are `ENABLED` immediately |

## 3. Replay

Evidence: [StartReplay](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_StartReplay.html), [ReplayDestination](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_ReplayDestination.html), [CancelReplay](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_CancelReplay.html) API references, [Creating replays of archived events](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-replay-archived-event.html), [AWS Compute Blog: Archiving and replaying events](https://aws.amazon.com/blogs/compute/archiving-and-replaying-events-with-amazon-eventbridge/).

| Implemented behavior | Documented fact |
|---|---|
| `Destination.FilterArns` limits which rules process replayed events | ReplayDestination: "`FilterArns` — A list of ARNs for rules to replay events to." User guide: "you can specify for EventBridge to send the events to specific rules." |
| Every replayed event carries a top-level `replay-name` field — visible in target envelopes, `InputPath`, `InputTransformer` paths, and event patterns | Compute blog: "The replay process adds a **'replay-name' attribute** to the original event." |
| The documented anti-replay guard `"replay-name": [{"exists": false}]` now works (`_matches_pattern` previously ignored unknown top-level keys, silently matching replays) | Compute blog: "Since the managed rule does not match the replayed event, it is not forwarded to the archive" — the same exists-guard mechanism users apply to their own rules |
| Destination bus must be the archive's source bus | ReplayDestination `Arn`: "You can replay events only to the event bus specified to create the archive." |
| Events replay in event-time order (pacing not modeled) | StartReplay: "A replay processes events to replay **based on the time in the event**, and replays them using 1 minute intervals." |
| `DescribeReplay` reports `EventLastReplayedTime` | StartReplay: "The value returned for `EventLastReplayedTime` indicates the time within the specified time range associated with the last event replayed." |
| `StartReplay` response includes `ReplayStartTime` alongside `ReplayArn`/`State` | Documented response members: `ReplayArn`, `ReplayStartTime`, `State`, `StateReason` |
| `EventStartTime`/`EventEndTime` are required members; `EventEndTime <= EventStartTime` → `ValidationException` | Both members are "Required: Yes"; "Only events that occurred between the `EventStartTime` and `EventEndTime` are replayed" |
| Replay name capped at 64 chars | "Length Constraints: Minimum length of 1. Maximum length of 64." |
| `CancelReplay` outside `RUNNING`/`STARTING` → `IllegalStatusException` (was `ValidationException`; cancelling a `CANCELLED` replay wrongly succeeded) | CancelReplay declared error `IllegalStatusException`: "An error occurred because **a replay can be canceled only when the state is Running or Starting.**" |
| Replay state machine `STARTING → RUNNING → COMPLETED/CANCELLED`, zombies flipped to `FAILED` on state restore | Documented state enum: `STARTING \| RUNNING \| CANCELLING \| COMPLETED \| CANCELLED \| FAILED` |

## 4. API destinations invoked over HTTP (earlier commits on this branch)

Evidence: [API destinations user guide](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-api-destinations.html) (eb-api-destinations), [Knowledge Center: API destination target failures](https://repost.aws/knowledge-center/eventbridge-api-destination-target-failure).

| Implemented behavior | Documented fact |
|---|---|
| Default headers on every request: `User-Agent: Amazon/EventBridge/ApiDestinations`, `Range: bytes=0-1048575`, `Accept-Encoding: gzip,deflate`, `Connection: close`, default `Content-Type: application/json; charset=utf-8` | Verbatim from the "Headers included in requests to API destinations" table |
| `User-Agent` and `Range` cannot be overridden by connection/target parameters | "EventBridge does not allow you to override the following headers: User-Agent, Range" |
| The removed-headers set (`A-IM`, `Accept-Charset`, …, `Warning`) is stripped from all requests | Mirrors the documented "Headers EventBridge removes from requests to API destinations" list verbatim |
| 5-second request timeout | "EventBridge requests to an API destination endpoint must have a maximum client execution timeout of 5 seconds. If the target endpoint takes longer than 5 seconds to respond, EventBridge times out the request." |
| HTTP `401/407/409/429/5xx` treated as retryable → dead-letter with `ERROR_FROM_TARGET` + `EXHAUSTED_RETRY_CONDITION`; other `4xx` non-retryable → dead-letter without retry | "Retries events associated with error codes 401, 407, 409, 429, and 5xx. Does not retry events associated with error codes 1xx, 2xx, 3xx, and 4xx (other than those noted above)." And: "After the maximum number of retries, events are sent to your dead-letter queue if you have one. Otherwise, the event is dropped." |
| Connections store auth parameters in a real Secrets Manager secret named `events!connection/<name>/<uuid>`, surfaced as `SecretArn` | "EventBridge uses this service-linked role to create and store a secret in Secrets Manager" (naming scheme **live-verified**) |
| OAuth client-credentials token requests authenticate with HTTP Basic (`ClientID:ClientSecret`); `grant_type`/`scope`/`audience` only reach the token request via `OAuthHttpParameters`; token cached per connection, proactively refreshed near expiry, refresh-and-retry-once on `401`/`407` | **Live-verified** against the EventBridge API (public docs do not specify the token-request wire format) |
| Connection `InvocationHttpParameters` merge with target `HttpParameters`, connection values taking precedence; `*` path params substituted; body params folded into JSON bodies | [CreateConnection `InvocationHttpParameters`](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_CreateConnection.html): parameters "to be included in every invocation"; precedence **live-verified** |
| Input-transformer object/array variables outside JSON *value* positions fail the invocation **before any HTTP request** and dead-letter as `INVALID_JSON` / "Invalid input for target." | `INVALID_JSON` is in eb-rule-dlq's published error-code list; the value-position constraint and pre-request failure are **live-verified** |
| `InvocationRateLimitPerSecond` accepted but not enforced; `Retry-After` not honored | Documented non-goal alongside the unmodeled retry pipeline (both only shape *timing*, which collapses to a single attempt locally) |
| Connection/API-destination names capped at 64 chars | [CreateConnection](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_CreateConnection.html)/[CreateApiDestination](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_CreateApiDestination.html): "Maximum length of 64" |

## 5. Control-plane validations

| Implemented behavior | Documented fact |
|---|---|
| `PutTargets` rejects non-JSON static `Input` with `ValidationException` | [Target API reference](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_Target.html): `Input` is "Valid JSON text passed to the target" |
| `DescribeConnection` returns the sanitized `ConnectionAuthResponseParameters` shape — secret values and `IsValueSecret` parameters never echo | [DescribeConnection](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_DescribeConnection.html) response carries `AuthParameters` as `ConnectionAuthResponseParameters` (no secret members in the response shape) |

## 6. CloudWatch Logs support work (same branch)

| Implemented behavior | Documented fact |
|---|---|
| Rule targets pointing at a `logs` log-group ARN deliver each matched event as a new opaque-named log stream | [EventBridge targets](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-targets.html): CloudWatch Logs is a supported rule target; per-delivery stream naming observed on live infrastructure |
| `PutResourcePolicy`/`DescribeResourcePolicies`/`DeleteResourcePolicy` served, with the account limit enforced (11th policy → `LimitExceededException`, in-place updates still succeed) | [PutResourcePolicy](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutResourcePolicy.html): "An account can have a maximum of 10 policies" (account-scoped); `LimitExceededException` is the declared error |
| `TagResource`/`UntagResource`/`ListTagsForResource` resolve vended-delivery ARNs (`delivery-source`, `delivery-destination`, `delivery`) | Required by the Terraform AWS provider's read-after-create on `aws_cloudwatch_log_delivery_*` (the community EventBridge module ≥ v4.1 provisions the trio for bus logging) |

## Testing

- 156 EventBridge integration tests pass (boto3 against the running stack), including new coverage for: multi-target `NO_RESOURCE` dead-lettering, non-SQS `DeadLetterConfig` ARN rejection, `FilterArns` + `replay-name` + no-re-archive + `EventLastReplayedTime`, the `exists: false` anti-replay guard, inverted replay windows, and `CreateArchive` validations.
- Adjacent suites exercising the shared dispatch paths pass: scheduler, S3 (EventBridge notifications), SQS, SNS, Pipes, CloudFormation, persistence, multitenancy, tagging.
- `ruff` clean; README service table and CHANGELOG updated per CONTRIBUTING.


t-rech:
Opened against the wrong base by mistake — this belongs on my fork. Apologies for the noise.

PR #1315, opened 2026-08-07
feat(kms): add UpdateKeyDescription

### LINKED ISSUE #1314: KMS: UpdateKeyDescription is not implemented (InvalidAction), breaking aws_kms_key updates
**Version:** 1.4.13 (`ministackorg/ministack:full`)

## Summary

`UpdateKeyDescription` is absent from the KMS dispatch table in `ministack/services/kms.py`, so every call returns:

```
An error occurred (InvalidAction) when calling the UpdateKeyDescription operation:
Unknown action: UpdateKeyDescription
```

The neighbouring key mutators are all present — `EnableKey`, `DisableKey`, `PutKeyPolicy`, `TagResource`, `UpdateAlias` — so this looks like an oversight rather than a deliberate omission.

## Impact

terraform-provider-aws calls `UpdateKeyDescription` whenever an `aws_kms_key` description drifts. The failure is not confined to that one field: it fails the whole update, and because Crossplane/Upbound also reconciles during deletion, the managed resource ends up stuck and undeletable:

```
Encryption-key is at failed state: delete failed: async update failed: failed to update the resource: updating kms key (31d7f7ae-...) description: operation error kms: UpdateKeyDescription, https response error StatusCode: 400, api error InvalidAction: Unknown action: UpdateKeyDescription
```

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test \
       AWS_REGION=us-east-1 AWS_ENDPOINT_URL=http://localhost:4566

KID=$(aws kms create-key --description "orig" --query 'KeyMetadata.KeyId' --output text)
aws kms update-key-description --key-id "$KID" --description "changed"
# actual:   InvalidAction: Unknown action: UpdateKeyDescription
# expected: empty response, and DescribeKey then reports "changed"
```

## Proposed fix

Follow the shape of the neighbouring key mutators — resolve by key id or ARN, `NotFoundException` when missing, empty response on success:

```python
def _update_key_description(data):
    rec = _resolve_key(data.get("KeyId", ""))
    if not rec:
        return error_response_json("NotFoundException", f"Key {data.get('KeyId', '')} not found", 400)
    # An explicit empty string clears the description, which is how AWS removes it.
    rec["Description"] = data.get("Description", "")
    return json_response({})
```

plus `"UpdateKeyDescription": _update_key_description` in the `handlers` dict.

sac-outsystems:
Fixes #1314

## Problem

`UpdateKeyDescription` was not registered in the KMS dispatch table, so every call returned `InvalidAction: Unknown action: UpdateKeyDescription` (HTTP 400).

terraform-provider-aws calls it whenever an `aws_kms_key` description drifts. Under Crossplane/Upbound the whole update fails, and because the provider also reconciles during deletion the managed resource ends up stuck in a failed state.

## Change

Adds `_update_key_description` and registers the action. Follows the shape of the neighbouring key mutators: resolve the key by id or ARN, `NotFoundException` when missing, empty response on success. An explicit empty description clears it, matching AWS.

Like `_enable_key` and `_put_key_policy`, this does not reject a key in `PendingDeletion`, where real AWS raises `KMSInvalidStateException`.

## Tests

Three tests in `tests/test_kms.py`:

- `test_kms_update_key_description` — the description DescribeKey reports changes.
- `test_kms_update_key_description_by_arn_and_cleared` — accepts an ARN, and an empty description clears it.
- `test_kms_update_key_description_unknown_key` — `NotFoundException` for a key that does not exist.

All three fail on 1.4.13 and pass with this change. `tests/test_kms.py` passes in full, as does `tests/test_ec2.py`. `ruff check ministack/services/kms.py` is clean.


Nahuel990:
Thanks!

PR #1319, opened 2026-08-07
feat(eventbridge): invoke API destination targets over HTTP on PutEvents

t-rech:
## What this adds

An *API destination* is EventBridge's built-in way to call an external HTTPS endpoint (a webhook) whenever an event matches a rule. A *connection* holds the credentials for it. MiniStack accepted all the setup calls (`CreateConnection`, `CreateApiDestination`, `PutTargets`) but never made the HTTP call — a matching event was logged as "unsupported event target ARN" and dropped. That made EventBridge → webhook pipelines impossible to test locally.

With this PR, a matching event triggers a **real HTTP request** to the configured endpoint, delivered on a background thread (the same pattern as the existing SNS HTTP delivery), using only the standard library.

## What the request looks like

Each item matches the [API destinations user guide](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-api-destinations.html):

- The event payload, after the target's `Input`/`InputPath`/`InputTransformer` rules are applied — the same payload-shaping every other target type already gets.
- The documented default headers (`User-Agent: Amazon/EventBridge/ApiDestinations`, default `Content-Type: application/json; charset=utf-8`, `Range: bytes=0-1048575`, …). `User-Agent` and `Range` cannot be overridden, and the guide's list of headers EventBridge strips (Host, Referer, Date, …) is applied verbatim — so connection parameters cannot smuggle them in.
- A 5-second timeout — the guide's "maximum client execution timeout".
- The destination's configured HTTP method.

## Authentication

Per the connection's `AuthorizationType`:

- `BASIC` → an `Authorization: Basic` header from the stored username/password.
- `API_KEY` → the configured header name/value.
- `OAUTH_CLIENT_CREDENTIALS` → MiniStack fetches a token from the connection's authorization endpoint, caches it per connection, refreshes it shortly before expiry, and — like real EventBridge — refreshes and retries once if the endpoint answers `401`/`407`.

## Extra parameters

Both a connection and a target may declare additional headers, query-string, and body parameters. They are merged; when both set the same key, **the connection's value wins** — that precedence comes straight from the [`HttpParameters` API reference](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_HttpParameters.html). Target `PathParameterValues` fill `*` wildcards in the endpoint URL, and body parameters fold into JSON-object bodies.

## Deliberately not modeled (documented in CHANGELOG/README)

AWS retries failed deliveries for up to 24 hours / 185 attempts with backoff, honors `Retry-After`, and rate-limits via `InvocationRateLimitPerSecond`. A local emulator can't usefully reproduce a timing pipeline, so delivery here is a single attempt.

## Tests

Integration tests spin up real local HTTP servers (stdlib `ThreadingHTTPServer`) that capture every request, plus a minimal OAuth token issuer — asserting the method, path, headers, auth, and body per auth type, with no mocks. Full EventBridge suite: 142/142 locally on top of v1.4.14.

*Note: supersedes #1313, which I opened prematurely before slicing this work into reviewable pieces — apologies for the noise. Follow-up PRs (dead-letter queue delivery, archive/replay parity) will come as separate scoped slices once this lands.*

## Checklist

- [x] Tests added and passing (parallel `-n 4 --dist=loadfile -m "not serial"` + serial `-m serial` phases)
- N/A — new-service checklist items (existing service)


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1320, opened 2026-08-07
feat(dsql): Aurora DSQL emulator with control plane and SQL-validating PG proxy

ry-allan:
We're in the midst of a migration to DSQL and although we use local Postgres containers, they don't fully simulate actual compatibility.

**Initially added support for control plane operations:**

- CreateCluster
- GetCluster
- UpdateCluster
- DeleteCluster
- ListClusters
- PutClusterPolicy
- GetClusterPolicy
- DeleteClusterPolicy
- TagResource
- UntagResource
- ListTagsForResource

Implements https://github.com/ministackorg/ministack/issues/1323


ministack/services/dsql.py
```diff
@@ -0,0 +1,866 @@
...
+    AccountScopedDict,
+    get_account_id,
+    get_region,
+    json_response,
+)
+
+logger = logging.getLogger("dsql")
+
+BASE_PORT = int(os.environ.get("DSQL_BASE_PORT", "25432"))
+PG_IMAGE = os.environ.get("DSQL_PG_IMAGE", "postgres:17-alpine")
```
vivedo: Feels worth pinning the default to 16 for consistency, even if it's already overridable via env, as AWS advertises compatibility with that version:

https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with.html#dsql-pg-overview-compat

ry-allan: Good spot, pinned.

Nahuel990:
Hey Ryan, the control plane looks good (matches the API and.a few local tests) but before I dig further can you share how you're running this on your end? Your driver and a snippet of your connect + a couple of the statements you're testing (please don't share sensitve data) 

I'm trying it locally and getting mixed results as some statements that DSQL rejects go straight through and run like normal Postgres, and some valid ones come back rejected. Trying to work out if it's my setup or the proxy, and knowing exactly how you're driving it would help me line up against yours (also I think the proxy you created, with some clients, behaves like a plain postgreSQL, which client did you validated with?)

Thanks

ry-allan:
Hey @Nahuel990,

The reason you would have been seeing discrepancies as it didn't cover the extended wire protocol - myself & colleague @vivedo have been discussing this and we've just merged his coverage updates into my branch to support.

For reference, we've been running raw wire-protocol probes against the proxy in front of a real `postgres:16-alpine`, exercising both Q and Parse/Bind/Execute for every rule + a live Aurora DSQL cluster (PostgreSQL 16, eu-central-1) to check the emulator's assertions.

Nahuel990:
@ry-allan this looks awesome! The only things I;d change are DSQL_STRICT (enable/disable containers, default should be control-plane-only for stubs and 1 starts the container) and capping DSQL_BASE_PORT to a number (~20/30 maybe (?) it currently creates x number of containers... eventually if someone complains we can increase it)

ry-allan:
@Nahuel990 those updates have been pushed.

Nahuel990 (approved):

Issue #533, opened 2026-04-29
Missing custom claims for Cognito access tokens; not a new service


aahoughton:
### Affected Service

Router / General

### What happened?

I considered adding this as a new service request, but it's not a brand new service and it's not a bug per se - just missing functionality. Filing it as a bug for want of a better choice, but `COGNITO` isn't a service in that dropdown.

We use a PreTokenGeneration lambda for our auth tokens.  There's no provision in ministack at the moment for that: `_fake_token` (ministack/services/cognito.py:328) hardcodes the access-token claim set to `client_id`, `auth_time`, `origin_jti`, `scope`, `username`, and `cognito:groups`. There's no way to add anything else.

MiniStack doesn't currently emulate the PreTokenGeneration trigger, and LambdaConfig isn't even round-tripped: `_create_user_pool` (cognito.py:732) doesn't include it in the kept fields, and `_update_user_pool` (cognito.py:835) doesn't list it in updatable, so it's silently dropped on both. There's nowhere to attach the configuration even if a future PR were to honour it.

This is the same type of issue as #341/#342 — the claim mint is hardcoded in `_fake_token`, the fix would be to make it data-driven.

Two options; both of these are additive and default-off, so existing behaviour would be preserved.  


**A. Env-var + tag passthrough (simpler, mirrors what's already in the repo)**

1. New env var `MINISTACK_COGNITO_ACCESS_TOKEN_CLAIMS_FROM_ATTRS` (or something less annoying), comma-separated user-attribute names. When set, _fake_token (access-token branch) copies those attributes from the user record into the claim set at mint time.
2. Per-pool override via an `ms-access-token-claims-from-attrs` tag, matching the `ms-custom-id` pattern used by API Gateway.
3. For client_credentials (no user record), a separate env var `MINISTACK_COGNITO_M2M_ACCESS_TOKEN_CLAIMS` holding a JSON object merged into every M2M access token.

Pros: small diff; mirrors the cognito:groups automatic copy and ms-custom-id patterns already accepted in the repo; trivial to test.
Cons: doesn't really address the missing PreTokenGeneration handling, which is the only way to get custom claims into a token in real AWS


**B. Declarative PreTokenGeneration emulation**

Honour a ministack-namespaced JSON config attached to the pool — like an `ms-pretoken-claims` tag — whose value matches the shape of a real PreTokenGeneration v2 Lambda response:
```
 {
    "accessTokenGeneration": {
      "claimsToAddOrOverride": {
[...]

PR #1336, opened 2026-08-10
fix(cognito): forward ClientMetadata into PreTokenGeneration's event.request

ryan-bennett:
## Summary

- `PreTokenGeneration`'s `event.request` now carries `clientMetadata`, matching the [Cognito Developer Guide's documented request shape](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html)
- Covers `InitiateAuth` (`USER_PASSWORD_AUTH`, `REFRESH_TOKEN_AUTH`), `AdminInitiateAuth` (`ADMIN_USER_PASSWORD_AUTH`/`ADMIN_NO_SRP_AUTH`, `REFRESH_TOKEN_AUTH`), and `GetTokensFromRefreshToken`

## Issue

`_build_pretoken_event` (added in #533) builds `event.request` from `userAttributes` / `groupConfiguration` / `scopes` only. `ClientMetadata` passed to `InitiateAuth`/`AdminInitiateAuth`/`GetTokensFromRefreshToken` is silently dropped before it reaches the trigger, even though `_invoke_define_auth_challenge_trigger` (the `CUSTOM_AUTH` path) already extracts and forwards it correctly for `DefineAuthChallenge`.

This breaks apps that pass metadata through a refresh call for the trigger to read — e.g. an org/tenant id on `InitiateAuth(REFRESH_TOKEN_AUTH, ClientMetadata={"orgId": ...})`, a pattern used to embed per-request claims on token refresh without a separate round trip.

## Fix

- `_fake_token` / `_apply_pretoken_trigger` / `_build_pretoken_event`: thread a `client_metadata` parameter through to `request_block["clientMetadata"]`
- `_build_auth_result` / `_refresh_auth_result`: accept and forward `client_metadata`
- `_initiate_auth`, `_admin_initiate_auth`, `_get_tokens_from_refresh_token`: extract `data.get("ClientMetadata", {})` and pass it through (the `CUSTOM_AUTH` branches' own now-redundant local extraction is removed since it's hoisted to the top of each function)

## Scope

Deliberately limited to the flows above. `RespondToAuthChallenge`'s `NEW_PASSWORD_REQUIRED` completion path and the Hosted UI / `/oauth2/token` endpoints don't forward `ClientMetadata` into this trigger either, but I've left those out of this PR to keep the diff reviewable — happy to follow up if that's wanted.

## Validation

- Added `test_cognito_pretoken_receives_client_metadata`, asserting a `PreTokenGenerationConfig` Lambda sees `event.request.clientMetadata` across all five call sites
- `uv run --extra dev pytest tests/test_cognito.py -v` — 228 passed
- `uv run --extra dev ruff check ministack/services/cognito.py tests/test_cognito.py` — clean
- CHANGELOG `Unreleased → Fixed` entry included

## References

- [PreTokenGeneration Lambda trigger](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html) — documents `clientMetadata` as part of `request`
- #533 — added PreTokenGeneration trigger support originally; its resolution comment is what confirmed `clientMetadata` was never included


Nahuel990:
@ryan-bennett  this forwards ClientMetadata on the operations Cognito excludes and skips the ones it actually uses.

Per the AWS pre-token generation docs (and checked in AWS just to be sure):
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html

clientMetadata reaches that trigger only from RespondToAuthChallenge / AdminRespondToAuthChallenge, plus aws_client_metadata on the M2M token endpoint: "Amazon Cognito doesn't include data from the ClientMetadata parameter in AdminInitiateAuth and InitiateAuth API operations in the request that it passes to the pre token generation function."
Refresh via InitiateAuth falls under that exclusion too.. so this populates event.request.clientMetadata on InitiateAuth where real AWS leaves it empty

The fix we'd take is the reverse: forward it from RespondToAuthChallenge / AdminRespondToAuthChallenge, plus aws_client_metadata for M2M, and drop the InitiateAuth / AdminInitiateAuth / refresh side

Can you share in which region you are deploying this or a snipet of your own setup (no sensitive data) to test it further?

ryan-bennett:
You're right, confirmed against the docs — my mistake. Pushed a fix: `ClientMetadata` now forwards from `RespondToAuthChallenge`/`AdminRespondToAuthChallenge` (all challenge types) and `aws_client_metadata` for the M2M `client_credentials` grant, and I dropped the `InitiateAuth`/`AdminInitiateAuth`/refresh forwarding entirely. Thanks for catching this.

Nahuel990:
@ryan-bennett thanks!

PR #1338, opened 2026-08-11
fix(cloudformation): resolve AWS::SSM::Parameter::Value<...> against SSM

ryan-bennett:
## Summary

A CloudFormation template parameter typed `AWS::SSM::Parameter::Value<String>` (or `<List<String>>` / `<AWS::EC2::Image::Id>`) takes an SSM parameter *name* as its Default/provided value — real CloudFormation resolves that name against Parameter Store before `Ref` ever sees it. This is the mechanism behind, e.g., AWS CDK's `StringParameter.valueForStringParameter()`, commonly used for cross-stack (and cross-CDK-app) discovery without CloudFormation's export ratchet-lock.

MiniStack's parameter resolver treated these the same as other AWS-specific parameter types (an AZ name, an AMI id, ...) whose given value already *is* the value — so it passed the literal parameter name straight through as if it were the resolved value. `Ref` returned the parameter's own name instead of its stored value, silently breaking any resource property (or any stack composing with another through a shared SSM parameter) built from it.

## Fix

`_resolve_parameters` now looks up the real value via `ssm.resolve_parameter_value()` (the same in-process helper `ecs.py` already uses for `secrets[].valueFrom`) for the three `AWS::SSM::Parameter::Value<...>` types specifically — not the rest of `_AWS_SPECIFIC_TYPES`, since e.g. `AWS::SSM::Parameter::Type` is an unrelated template-side type constraint, not an SSM lookup at all. A name that doesn't exist now fails `CreateStack`/`UpdateStack` synchronously with a `ValidationError`, matching real AWS's parameter-validation-before-resource-creation behavior.

A follow-up commit fixes a bug in the first: `UsePreviousValue` (what `aws cloudformation deploy` and CDK both send for parameters a change set isn't touching, e.g. CDK's own `BootstrapVersion`) was feeding an *already-resolved* value from the prior deployment back into another SSM lookup, where it's essentially never itself a valid parameter name.

## Test plan
- [x] New test: a `String`-typed SSM parameter is put, then referenced via `AWS::SSM::Parameter::Value<String>` — `Ref` resolves to the stored value, not the parameter name
- [x] New test: naming a parameter that was never put fails `CreateStack` synchronously with `ValidationError`
- [x] New test: `UsePreviousValue` on a stack update reuses the already-resolved value without re-resolving it
- [x] Full `tests/test_cfn.py` + `tests/test_ssm.py` + `tests/test_eventbridge.py` suite (336 tests) passes
- [x] `ruff check` clean on both changed files
- [x] Verified against a real deploy: a second CDK app importing a first stack's Cognito user pool via `StringParameter.valueForStringParameter` + `UserPool.fromUserPoolId`


Nahuel990:
@ryan-bennett taking this one. The code is correct and it's a real bug, so I'll land it  but the bug you describe isn't actually fixed in the code yet. It only resolves three of the ~22 AWS::SSM::Parameter::Value types, so the common ones like VPC::Id, Subnet::Id, KeyPair::KeyName, and the List variants still return the parameter name. List<String> also comes back as a string instead of a list.

Looks like only the String case, which is what CDK's valueForStringParameter emits, got exercised. Did you check the other types against real AWS for your use case?

My fix on top will match the whole AWS::SSM::Parameter::Value prefix and return a list for the List and CommaDelimitedList variants

Going forward, when you hit an error, open an issue first so we can agree on scope before the PR. That'll save us both the round trips.

Thanks!

Nahuel990 (approved):

Issue #1325, opened 2026-08-10
RDS: per-instance containers + streaming replication for aurora-postgresql clusters (real reader endpoints, read-only replicas)


Kiran01bm:
## Problem

An Aurora PostgreSQL cluster in MiniStack owns exactly **one** PostgreSQL container. `DBInstance` records are control-plane aliases that all point at the same process:

- `_start_cluster_shared_container` launches the single shared container on the first `CreateDBInstance`; every subsequent instance is attached to it via `_attach_instance_to_shared_cluster`.
- `_sync_cluster_endpoints` points both `Endpoint` **and** `ReaderEndpoint` at that same container — the code comment itself notes the reader endpoint is read/write and that true read-only behavior would need a separate replicating process.
- Writer selection is metadata-only: `_register_instance_in_cluster` marks the first member `IsClusterWriter=True`, but nothing about the data plane distinguishes writer from reader.

This means several real-AWS behaviors cannot be rehearsed locally:

- A reader instance **accepts writes**. On real Aurora, DML against a reader fails with SQLSTATE `25006` (`cannot execute ... in a read-only transaction`).
- `ReaderEndpoint` and `Endpoint` are indistinguishable — connection-routing logic (write to writer endpoint, read from reader endpoint) passes trivially and falsely.
- There is no replication between members, so nothing that observes replica behavior (read-after-write against a reader, recovery state via `pg_is_in_recovery()`) can be tested.
- `FailoverDBCluster` has no data plane to act on — a future implementation (promote a reader, demote the writer) needs per-instance containers to exist first.

## Proposed design

For `aurora-postgresql` clusters with more than one instance:

1. **First instance (writer)** — unchanged: gets the cluster's container exactly as today.
2. **Each subsequent `CreateDBInstance`** — launches a **sibling container** configured as a hot-standby streaming replica of the writer, using PostgreSQL-native mechanics: `pg_basebackup -R` from the writer before start (which writes `standby.signal` + `primary_conninfo`), so the reader runs in recovery. No new dependencies — same official `postgres:<major>-alpine` images, same Docker SDK launch path.
3. **`ReaderEndpoint`** resolves to a reader container's host port (single reader: that one; multiple readers: first — MiniStack has no load balancer and that seems fine).
4. **Read-only enforcement comes free**: hot standby rejects DML with SQLSTATE `25006` and reports `pg_is_in_recovery() = true`, matching real Aurora reader semantics without any emulation logic.
[...]

PR #1339, opened 2026-08-11
rds: reshape Aurora cluster compute lifecycle for per-instance containers (#1325 slice 1)

Kiran01bm:
## Summary

Internal refactor of the Aurora cluster compute lifecycle in `ministack/services/rds.py` — no behavior change. Slice 1 of #1325.

## Why

#1325 proposes per-instance reader containers with streaming replication for `aurora-postgresql` clusters. Three decisions in the RDS service currently hard-code the single-shared-container model, each in a different way:

- "does this instance own its container?" is a hand-rolled compound test (`_docker_container_id` and not `_shared_cluster_id`) duplicated at instance delete and reset sweep;
- "what compute does this cluster have?" is a direct read of `_shared_container_id`;
- `ReaderEndpoint` is assigned from the writer's endpoint inline.

Landing these as named seams first — with today's semantics — keeps the replication change reviewable on its own and proves this step alone changes nothing.

## What

- `_instance_owns_container(instance)` — the ownership predicate, adopted at `_delete_db_instance` and the reset sweep. A future cluster member that owns its own container is handled like a standalone instance by these paths automatically.
- `_cluster_owned_container_ids(cluster)` — enumerates all containers backing a cluster's compute (today: at most the shared one); `_stop_cluster_shared_container` iterates it.
- `_cluster_reader_endpoint(cluster)` — the endpoint `ReaderEndpoint` resolves to (today: the writer's shared endpoint); `_sync_cluster_endpoints` uses it.
- Three unit tests pinning the helpers' single-occupant semantics; full RDS suite passes unchanged (195 passed, 7 skipped).

Deliberately untouched: the persisted `_shared_*` cluster keys (renaming would break warm-boot restore of existing state) and `_remove_cluster_shared_resources` (DeleteDBCluster requires members deleted first, so member-owned containers cannot exist there).


Nahuel990:
@Areson I think you might be working towards the same goal in your #1321 draft. Could you confirm so we can avoid duplicating work?

Areson:
@Nahuel990 I don't think it's quite the same. This is Postgres focused while the work I was starting is for MySQL. They will probably have some overlapping bits as more PRs are made but I can work through that with @Kiran01bm , so at the moment I don't see this particular PR as being in conflict with my draft PR.

Nahuel990 (approved):

PR #1341, opened 2026-08-11
fix(cloudformation): ApiGatewayV2 Authorizer survives a property update

ryan-bennett:
## Summary

**Depends on #1340** — built on top of that branch, so this diff includes #1340's commits until it merges (at which point this PR's diff will shrink to just the new commits below). Please review/merge #1340 first.

`AWS::ApiGatewayV2::Authorizer` had no update handler, so a property change (e.g. adding a trusted app client to `JwtConfiguration.Audience` — hotshot's multi-app-support Phase B, the same real deploy #1339 and #1340 came from) fell back to calling create again. Its `authorizerId` is a random id generated fresh on every call — unlike the name-based resources #1340's `_physical_name` fix covers, there's no stable identity to derive here at all. Every such update minted a second, orphaned authorizer under a new id, while every `Route`'s own `AuthorizerId` (unchanged, so the Route itself was never reprocessed) kept pointing at the original, now-stale one — so the new audience was silently never actually enforced.

Why this depends on #1340 rather than standing alone: without it, the *parent* `AWS::ApiGatewayV2::Api` resource's own id also churns on every update (same no-update-handler-plus-opaque-id shape, one level up) even when nothing about the Api itself changed — which cascades a wrong `ApiId` into the Authorizer regardless of this fix. Confirmed directly: this fix alone, applied to plain `main`, did not resolve the real-world repro; applied on top of #1340, it does.

## Fix

A real `update` handler for `AWS::ApiGatewayV2::Authorizer` that mutates the existing in-memory authorizer record in place under its own `authorizerId`, matching real API Gateway's `UpdateAuthorizer` (which changes configuration without reassigning the authorizer's id).

## Test plan
- [x] New test: updating an Authorizer's `JwtConfiguration.Audience` (unrelated Route/Api untouched) mutates the same authorizer in place — same `AuthorizerId`, new audience list, no second authorizer created
- [x] Existing `AWS::ApiGatewayV2::Authorizer` create/delete regression test still passes
- [x] Full `tests/test_cfn.py` + `tests/test_apigatewayv2.py` + `tests/test_ssm.py` + `tests/test_eventbridge.py` suite (456 tests) passes on top of #1340
- [x] `ruff check` clean on both changed files
- [x] Verified against the same real multi-stack deploy as #1339/#1340: a second app's client id, once added to `additionalTrustedAppClientIds` and redeployed, is now actually accepted as a trusted audience by the running API Gateway authorizer (confirmed via a real token exchange against the redeployed stack, not just template inspection)


Nahuel990:
Thanks! This one looks good

Issue #1345, opened 2026-08-11
API Gateway authorizers, CloudFormation update semantics, Lambda error classification


iot-rocket:
Context: we run ~55 CloudFormation + Serverless Framework stacks (≈30 Python
Lambda services behind REST APIs) against ministack. Most of it works; these are
the platform-level behaviours we had to work around. Roughly in impact order.

## 1. API Gateway never invokes custom authorizers (highest impact for us)

`CreateAuthorizer` stores the authorizer and `GetMethod` reports
`authorizationType: CUSTOM` with the right `authorizerId`, but on the data path
the authorizer Lambda is **never invoked** and `requestContext.authorizer` is
**never populated**. Requests reach the integration Lambda unauthenticated.

Repro: deploy an API with a TOKEN authorizer, then

```bash
curl "http://localhost:4566/restapis/<api-id>/dev/_user_request_/private"
# 200 - the integration ran; the authorizer Lambda has zero invocations
# AWS: 401 (no Authorization header) / 403 (Deny policy)
```

Expected AWS behaviour: extract the identity source, invoke the authorizer with
`{type, authorizationToken, methodArn}` (or the REQUEST shape), honour
`authorizerResultTtlInSeconds` caching, answer 401/403 on Deny or error, and on
Allow inject `requestContext.authorizer` with the returned `context` (values
stringified) plus `principalId`.

This is what made us put a proxy in front of ministack, so it's the single
change that would shrink our workaround the most. Two related details we hit
while implementing it:

- **Methodless intermediate resources must fall through.** API Gateway creates
  path nodes implicitly: `/jobs` exists only because `/jobs/{operation_type}`
  does, and defines no methods. `GET /jobs` must fall through to the `{proxy+}`
  catch-all, not 403. Selecting purely by path specificity answers 403 for every
  route under such a node.
- **aws_iam methods** should reject requests without a SigV4 `Authorization`
  header (403 `Missing Authentication Token`) - currently they're served.

## 2. CloudFormation `UpdateStack` duplicates resources

Updating a stack re-creates its resources instead of updating them in place.
Minimal repro - a template with a `AWS::ApiGateway::RestApi` and an
`AWS::SSM::Parameter`, changing only the parameter value:

```bash
aws cloudformation deploy --stack-name dup --template-file t1.yaml
aws apigateway get-rest-apis --query 'items[?name==`dup-api`]|length(@)'   # 1
[...]

PR #1348, opened 2026-08-11
fix(lambda): classify uncaught handler exceptions as Unhandled function errors

iot-rocket:
## Summary

- The Docker/RIE executor now reports `X-Amz-Function-Error: Unhandled` for failed invocations, matching real AWS
- Classification is extracted into a pure `_classify_function_error()`, unit-tested over the observed payload shapes of all four runtime families, so the docker-only decision is exercised in every CI run

## Issue

Part of #1345 (§6). The executor keyed "handled vs unhandled" on an error header the RIE never sets, so every failure surfaced as `Handled`, and API Gateway consumers relying on the 502-for-Unhandled contract never took the error path.

Probing the RIE bundled with `public.ecr.aws/lambda/python:3.12`, `nodejs:22` and `provided:al2023`: no error header is returned on the invocation response — not for a raised handler exception, and not even when a custom runtime POSTs its own `Lambda-Runtime-Function-Error-Type` to `/runtime/invocation/<id>/error` (the RIE drops it and logs `RTDONE(status: success)` either way).

## Fix

- Classification keys on the payload shape, which differs per runtime: python sends `stackTrace` — empty (`[]`) for init failures such as `Runtime.ImportModuleError` — nodejs sends `trace`, and go / `provided.*` send no traceback field at all. An error-shaped payload (`{errorType, …}`) therefore reports `Unhandled` unless it also carries a `statusCode`, which marks an HTTP-style envelope the handler returned itself (`Handled`). Testing one `stackTrace` key for truthiness would have left import errors and every nodejs and go failure on the `Handled` path
- Known tradeoff, chosen deliberately: a handler that *returns* a bare `{errorType, errorMessage}` dict is indistinguishable from a go-style runtime error over this wire and reports `Unhandled` on the docker executor — the safe direction, since only Handled-instead-of-Unhandled breaks the 502 contract. The local (default) executor is unaffected: it classifies from the raise itself and still sets no `FunctionError` for a returned dict
- When a header IS present it still wins, passed through as before

## Validation

- Unit tests cover the four observed runtime payload shapes plus the header-present, `statusCode` and non-error cases
- `pytest tests/test_lambda.py` against a live server — 318 passed, 3 skipped (3 failures — `warm_start`, `apigwv2_nodejs_lambda_proxy`, `docker_flags_applied_to_run_kwargs` — fail identically on the base commit in a docker-enabled environment)
- The fixed path is pinned by `test_lambda_docker_uncaught_exception_is_unhandled_function_error` (docker-gated in CI; run here with `LAMBDA_EXECUTOR=docker`), and cross-checked by hand: a `nodejs22.x` function through a real RIE container reports `Unhandled` with this patch and `Handled` without it
- `ruff check` clean; CHANGELOG `Unreleased → Fixed` entry included

## References

- [Invoke API — FunctionError](https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html) — `Unhandled` for errors "detected by the runtime"
- Part of #1345 (§6)


iot-rocket:
The shard-2 failure is the pre-existing flaky test reported in #1356 - test_cfn_unnamed_dynamodb_table_survives_unrelated_update asserts global list_tables equality and raced test_dynamodb.py's table creates on the shared server; it reproduces on unmodified v1.4.16 and is unrelated to this diff (lambda-only). Rerun should be green.

Nahuel990 (approved):

PR #1350, opened 2026-08-11
fix(cognito): return custom attributes and username from /oauth2/userInfo

rjmackay:
`/oauth2/userInfo` builds its response from a fixed standard-OIDC allowlist, so any `custom:` attribute on the user is silently dropped.

**The problem:**

- Real Cognito returns custom attributes to `openid` and `openid profile` tokens, bounded by the app client's read attributes — see [the userInfo docs](https://docs.aws.amazon.com/cognito/latest/developerguide/userinfo-endpoint.html), whose example response includes `"custom:mycustom1": "CustomValue"`.
- `admin-create-user --user-attributes Name=custom:tenant_id,...` stores the attribute fine; it just never comes back out of userInfo.
- The failure is quiet: a 200 with the claim missing, so apps that key off a custom attribute (tenant id, internal user id) reject the user rather than error visibly. It's what currently blocks us swapping a hand-written userinfo stub for MiniStack.

**What changed** (`handle_oauth2_userinfo`):

- Copy every `custom:` / `dev:custom:` attribute into the claims.
- Add the plain `username` claim AWS returns; `cognito:username` stays for callers already reading it here.
- Add `middle_name` to the standard claim list — it's in the `openid profile` set but was missing.

**Not in scope:** scope-based filtering. Real Cognito varies the response by the access token's scopes, but access tokens here always carry `scope: aws.cognito.signin.user.admin`, so honouring scopes means changing token minting first. Returning custom attributes unconditionally matches the `openid` case ("all user attributes that the app client can read").

**Testing:** new `test_oauth2_userinfo_returns_custom_attributes` walks login → token → userInfo and asserts two custom attributes round-trip; `test_oauth2_userinfo` gained a `username` assertion. Full `tests/test_cognito.py` — 228 passed. `ruff check` clean on both changed files.

---

Full disclosure: I wrote this with Claude Code, but I've checked its work.


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1357, opened 2026-08-12
feat(sesv2): add email template CRUD and Content.Template rendering

mikolajk0wal:
# feat(sesv2): add email template CRUD and Content.Template rendering

## What this adds

SESv2's `SendEmail` carries its content in one of three shapes — `Simple`, `Raw`, or `Template`. MiniStack read the first two and ignored the third: a templated send was accepted with a `200` and a `MessageId` while the template was silently dropped, so the stored message — and the SMTP relay when `SMTP_HOST` is set — went out with an empty subject and an empty body. Creating a template through v2 wasn't possible either: `CreateEmailTemplate` and the rest of the template family returned `404 NotFoundException`.

The result was that an app moving from the v1 `ses` client to `sesv2` had no working template path in either direction, and the failure mode was a silent success rather than an error — the send reported a `MessageId` and delivered a blank email.

This PR renders `Content.Template` and serves the five email-template operations.

## Template CRUD

Served at `/v2/email/templates[/{TemplateName}]`; request and response shapes verified against botocore's `sesv2-2019-09-27` model rather than the docs prose.

| Operation | Route | Behaviour |
| --- | --- | --- |
| `CreateEmailTemplate` | `POST /templates` | `AlreadyExistsException` on a duplicate name; `TemplateName` and `TemplateContent` both required |
| `GetEmailTemplate` | `GET /templates/{TemplateName}` | returns `TemplateContent` (`Subject`/`Text`/`Html`), plus `Tags` when they were supplied at create |
| `UpdateEmailTemplate` | `PUT /templates/{TemplateName}` | `TemplateContent` is required in the model, so content is replaced wholesale — an omitted `Html` clears it rather than persisting the old value |
| `DeleteEmailTemplate` | `DELETE /templates/{TemplateName}` | `NotFoundException` for an unknown name |
| `ListEmailTemplates` | `GET /templates` | `TemplatesMetadata` with `TemplateName` and `CreatedTimestamp` |

`SendEmail` resolves `Content.Template` from `TemplateName`, from a `…:template/<name>` `TemplateArn`, or from `TemplateContent` supplied inline — the last of which AWS renders without storing anything, so neither does MiniStack. Rendering reuses the v1 `{{placeholder}}` substitution, so the same template renders identically whichever API version sends it.

## One template store, two API versions

On AWS, `ses` and `sesv2` are two APIs over one backend: a template created with `aws ses create-template` is readable with `aws sesv2 get-email-template`. The v2 operations therefore read and write the existing v1 store (`ses._templates`) instead of introducing a second one. A separate store would invent a behaviour AWS does not have — two disjoint template namespaces — and would break the common local-provisioning pattern where an init script creates templates with the v1 CLI while the application sends through v2.

That property is pinned by `tests/test_ses.py::test_ses_v2_send_email_with_v1_template`. It is worth calling out that the v2-only unit tests do **not** cover it: with the stores split they all stay green, because both sides of those tests reach the same object. Only the cross-version test fails.

## Error parity

Every code and status below comes from the `sesv2` model's error list for the operation in question:

- a named template that does not exist fails `SendEmail` with `NotFoundException` (404) instead of delivering an empty email
- a `Content.Template` carrying none of `TemplateName` / `TemplateArn` / `TemplateContent`, or a `TemplateArn` that isn't a template ARN, is rejected with `BadRequestException`
- nothing is recorded and no SMTP relay happens when a send is rejected

One fix reaches beyond templates: SESv2 error responses now set the `x-amzn-errortype` header. restJson1 resolves the error shape from that header, and without it every SESv2 error arrived at SDKs as a bare HTTP status — boto3 reported `An error occurred (404)` instead of `NotFoundException`, so typed handling (`except client.exceptions.NotFoundException`) never matched on any SESv2 operation. It is a separate `Fixed` entry in the CHANGELOG.

## Inspection

Records for templated sends keep the rendered subject and body alongside the `Template` name and `TemplateData`, so `GET /_ministack/ses/messages` shows what the recipient would actually have received instead of a blank message with template metadata attached.

## Deliberately not modeled

- `TestRenderEmailTemplate` (`POST /v2/email/templates/{TemplateName}/render`) and `SendBulkEmail` (`/v2/email/outbound-bulk-emails`) — both still `404`.
- Email-template tagging through `TagResource` / `UntagResource` / `ListTagsForResource`. AWS tags configuration sets, identities, contact lists and dedicated IP pools, not email templates, so template ARNs stay rejected by the tag API (an existing test asserts this). `Tags` passed to `CreateEmailTemplate` are stored and returned by `GetEmailTemplate` only, rather than dropped.
- `ses.py` still carries a second, older v2 implementation (`_handle_v2` plus 16 `_v2_*` functions). It is superseded by `ses_v2.py`, which `app.py` routes every `/v2/email/*` request to, so no SDK client can reach it; its `_v2_send_email` has the same missing-render bug fixed here. Removing it is a separate cleanup PR so this diff stays reviewable.

## Validation

- 13 new collected cases from 7 test functions in `tests/test_ses_v2.py` (file goes 29 → 42): CRUD round-trip, duplicate and incomplete create, `NotFoundException` on GET/PUT/DELETE of a missing template, rendering from a stored template, from a template ARN and from inline content, and 5 parametrized cases asserting a rejected send records nothing.
- 1 integration test in `tests/test_ses.py` covering the cross-version path over boto3 and asserting through `/_ministack/ses/messages`.
- `pytest tests/test_ses.py tests/test_ses_v2.py` — 83 passed. SES-touching tests elsewhere also pass: `test_persistence.py`, `test_multitenancy.py` (`-k "ses or template"`, 15 passed) and the `/v2/*` routing cases in `test_ecr.py`.
- `ruff check ministack/services/ses_v2.py tests/test_ses.py tests/test_ses_v2.py` — clean.
- End-to-end against a running emulator with real boto3 `ses` + `sesv2` clients: v1 `create_template` → v2 `send_email` → the rendered subject and body read back off `/_ministack/ses/messages`, a display-name recipient (`"Alice Example" <alice@example.com>`) surviving the round trip verbatim, and `NotFoundException` raised as the typed exception for a deleted and for an unknown template.
- `POST /_ministack/reset` verified to clear templates created through both API versions.
- CHANGELOG: `Unreleased → Added` for the template work, `Unreleased → Fixed` for the `x-amzn-errortype` header. README service table updated.

## References

- [SESv2 `SendEmail`](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_SendEmail.html) — `EmailContent` is `Simple` | `Raw` | `Template`
- [SESv2 `CreateEmailTemplate`](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_CreateEmailTemplate.html) and [`Template`](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_Template.html) — `TemplateName`, `TemplateArn`, inline `TemplateContent`, `TemplateData`
- [SESv2 `TagResource`](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_TagResource.html) — the taggable resource list that excludes email templates
- botocore `sesv2-2019-09-27` service model — routes, shapes, error codes and HTTP statuses

## Checklist

- [x] Tests added and passing
- [x] Linting passes (`ruff check` on the changed files)
- N/A — new-service checklist items (existing service)


Nahuel990:
@mikolajk0wal Thanks! There is just one small thing missing from ListEmailTemplates which is support for NextToken and PageSize https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_ListEmailTemplates.html

mikolajk0wal:
> @mikolajk0wal Thanks! There is just one small thing missing from ListEmailTemplates which is support for NextToken and PageSize https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_ListEmailTemplates.html

Added

PR #1361, opened 2026-08-12
feat(stepfunctions): support aws-sdk:route53 tasks

bandle:
### Issue

Every `arn:aws:states:::aws-sdk:route53:*` Task fails at dispatch with `States.Runtime: Service 'route53' is not supported in MiniStack aws-sdk integrations`. Plain boto3 `change_resource_record_sets` works against the same server, so it is the integration only, the same shape as #1272

I needed it for a machine that upserts a DNS record on instance start and deletes it on stop. The create fails outright; the delete fails into a `Catch` that continues because DNS removal is best effort, so the machine reports success and the record survives.

### Fix

Three commits, reviewable separately.

- makes the dispatcher service-agnostic: it was just S3. Fix rest-xml errors: A failing S3 op came back `S3.NoSuchBucket` instead of `S3.NoSuchBucketException`
- Add route53 with a `ChangeResourceRecordSets` spec. Route 53 wraps XML lists (`<Changes><Change/></Changes>`) where S3 repeats the parent element, and nests `Code`/`Message` under `<Error>` with a default xmlns so `find("Code")` misses. Both are silent bugs
- Convert rest-xml body member names to wire names, as the json and query dispatchers do. Step Functions rejects `TTL` at `CreateStateMachine` ("Did you mean 'Ttl'?"), and `<Ttl>` is not what route53 reads, so the record was stored with no TTL while the execution succeeded.

Code written with Claude Code and verified against real AWS as described below.

### Testing

New tests in `tests/test_stepfunctions.py`, each failing before its commit:

- `test_sfn_aws_sdk_route53_change_resource_record_sets` — reads the record back with boto3, both values and the TTL, then deletes through a second execution
- `test_sfn_aws_sdk_route53_request_matches_botocore` — path and body byte-identical to botocore's rest-xml serialiser
- `test_sfn_aws_sdk_route53_ttl_uses_the_sdk_spelling` — asserted on the stored record, since the wrong body is accepted rather than rejected
- `test_sfn_aws_sdk_route53_errors_name_their_service` — `Route53.NoSuchHostedZoneException`, and an unmapped op lists what is covered
- `test_sfn_aws_sdk_s3_error_names_the_sdk_exception` — asserts the full `S3.NoSuchBucketException`, the value real Step Functions returns

The script below runs against MiniStack or real AWS, tested in eu-central-1:

| target | result |
| --- | --- |
| MiniStack v1.4.16 | 2/10 |
| this branch | 10/10 |
| real AWS, eu-central-1 | 10/10 |

```
                                  before                          this branch / real AWS
UPSERT creates the record         States.Runtime, not supported   record present, right value and TTL
output is the ChangeInfo shape    got keys ['Cause', 'Error']     Id=C0388494… Status=PENDING
a multi-value record keeps both   States.Runtime, not supported   both values present
DELETE removes the record         States.Runtime, not supported   record gone
/hostedzone/<id> form rejected    States.Runtime, not supported   Route53.InvalidInputException
bad zone id names the exception   States.Runtime, not supported   Route53.NoSuchHostedZoneException
s3 deleteBucket names it          S3.NoSuchBucket                 S3.NoSuchBucketException
s3 listObjectsV2 names it         S3.NoSuchBucket                 S3.NoSuchBucketException
the S3 message survives           message preserved               (control, passes either way)
no double suffix                  S3.S3Exception                  (control, passes either way)
```

The last two lines are controls: they pass before and after, so the eight failures above are the dispatcher and not the fixture.

The AWS leg found the `TTL` rejection and the id customisation, neither visible against the emulator. The suffix was measured there too: `NoSuchHostedZone`, `InvalidInput` and `InvalidChangeBatch` for route53, and five S3 operations.

<details>
<summary>check_sfn_rest_xml_dispatch.py</summary>

```python
#!/usr/bin/env python3
"""Check MiniStack's rest-xml aws-sdk dispatch: route53 support, and how S3 errors are named.

Two defects, one dispatcher. route53 is missing from _AWS_SDK_SERVICE_MAP, so every
arn:aws:states:::aws-sdk:route53:* Task fails at dispatch with

    States.Runtime  Service 'route53' is not supported in MiniStack aws-sdk integrations

while the identical boto3 change_resource_record_sets call against the same server works. A machine
that upserts a DNS record on start and deletes it on stop cannot run: the create side fails
outright, and the delete side fails into a best-effort Catch, so the machine reports success and
leaves the record behind.

Separately, the dispatcher returned S3's wire <Code> verbatim, so a missing bucket came back as
S3.NoSuchBucket where real Step Functions reports S3.NoSuchBucketException. A Catch written against
AWS never fires, and nothing looks wrong: the failure carries the right message under a name one
token short. Checks 9 and 10 are controls that pass before and after, so a failure there means the
fixture is wrong rather than the naming.

Against MiniStack:
  python check_sfn_rest_xml_dispatch.py --endpoint http://localhost:4566 --region us-east-1

  python check_sfn_rest_xml_dispatch.py --profile <profile> --region us-east-1 \
      --hosted-zone-id Z0123456789ABCDEFGHIJ --role-arn arn:aws:iam::<acct>:role/<sfn-role>

With --hosted-zone-id the zone is used as-is and never deleted; without one a zone is created and
removed again, which only makes sense against an emulator. Records this script creates are named
check-<hex>.<zone> and are deleted on the way out unless --keep is passed. The S3 checks target a
bucket that does not exist, so they create nothing and need no S3 permissions. Exit status is 0
when every check passes.
"""

import argparse
import json
import sys
import time
import uuid

import boto3
from botocore.config import Config as BotoConfig

TTL = 60


def options():
    parser.add_argument("--hosted-zone-id", help="Use this zone instead of creating one; never deleted")
    parser.add_argument("--role-arn", default="arn:aws:iam::000000000000:role/sfn-role",
                        help="Role the state machines assume")
    parser.add_argument("--keep", action="store_true", help="Leave records behind")
    return parser.parse_args()


def main():
    args = options()
                                aws_secret_access_key="test")
    else:
    kwargs = {"endpoint_url": args.endpoint} if args.endpoint else {}
    r53 = session.client("route53", **kwargs)
    # StartSyncExecution normally goes to a sync- prefixed host, which an emulator does not serve.
    sfn = session.client("stepfunctions", config=BotoConfig(inject_host_prefix=not args.endpoint),
                         **kwargs)

    tag = uuid.uuid4().hex[:8]
    created_zone = None
    created_records = []
    results = []

    def check(label, ok, detail):
        print(f"{'PASS' if ok else 'FAIL'}  {label:38} {detail}")

    def run_task(params, action="changeResourceRecordSets", service="route53"):
        """Drive one aws-sdk task and return (status, output, error, cause)."""
        name = f"restxml-check-{tag}-{uuid.uuid4().hex[:6]}"
        arn = sfn.create_state_machine(
            name=name,
            definition=json.dumps({
                "StartAt": "Change",
                "States": {"Change": {
                    "Type": "Task",
                    "Resource": f"arn:aws:states:::aws-sdk:{service}:{action}",
                    "Parameters": params,
                    "End": True,
                }},
            }),
            roleArn=args.role_arn,
        )["stateMachineArn"]
        try:
            # A STANDARD state machine polled to completion, not StartSyncExecution: sync execution
            # is EXPRESS-only on real AWS, and EXPRESS additionally requires a loggingConfiguration.
            # Standard is also what the machines this check stands in for actually run.
            execution = sfn.start_execution(stateMachineArn=arn, input=json.dumps({}))["executionArn"]
            deadline = time.monotonic() + 60
            desc = sfn.describe_execution(executionArn=execution)
            while desc["status"] == "RUNNING" and time.monotonic() < deadline:
                time.sleep(0.25)
                desc = sfn.describe_execution(executionArn=execution)
            output = json.loads(desc["output"]) if desc.get("output") else {}
            return desc["status"], output, desc.get("error", ""), desc.get("cause", "")
        finally:
            try:
                sfn.delete_state_machine(stateMachineArn=arn)
            except ClientError:
                pass

    def batch(action, name, values, ttl=TTL, ttl_key="Ttl"):
        return {"Comment": f"check {tag}",
                "Changes": [{"Action": action,
                             # ttl_key defaults to the SDK spelling "Ttl", because Step Functions
                             # rejects the wire one at CreateStateMachine ("The field \"TTL\" is
                             # not supported by Step Functions. Did you mean 'Ttl'?"), so that is
                             # the only form a definition running on real AWS can carry. The direct
                             # boto3 calls below pass ttl_key="TTL": the Python API takes the wire
                             # name, and botocore rejects "Ttl" outright.
                             "ResourceRecordSet": {"Name": name, "Type": "A", ttl_key: ttl,
                                                   "ResourceRecords": [{"Value": v}
                                                                       for v in values]}}]}

    def records_named(name):
        """Every record set with this name, read back through plain boto3."""
        paginator = r53.get_paginator("list_resource_record_sets")
        wanted = name if name.endswith(".") else name + "."
        return [rs for page in paginator.paginate(HostedZoneId=zone_id)
                for rs in page["ResourceRecordSets"]
                if rs["Name"] == wanted and rs["Type"] == "A"]

    try:
        if args.hosted_zone_id:
            zone_id = args.hosted_zone_id
            zone_name = r53.get_hosted_zone(Id=zone_id)["HostedZone"]["Name"]
        else:
            zone_name = f"sfn-check-{tag}.example.com."
            created = r53.create_hosted_zone(Name=zone_name,
                                             CallerReference=f"check-{tag}")["HostedZone"]
            zone_id = created["Id"]
            created_zone = zone_id
        # The bare id is what the API wants in the URI; boto3 also accepts the /hostedzone/ form.
        bare_id = zone_id.rsplit("/", 1)[-1]
        single = f"check-{tag}.{zone_name}"
        multi = f"check-multi-{tag}.{zone_name}"

        # 1. The create side of the machine: UPSERT a record, then read it back with plain boto3.
        #    Task success alone proves nothing -- the record has to actually be there.
        status, output, error, cause = run_task(
            {"HostedZoneId": bare_id, "ChangeBatch": batch("UPSERT", single, ["10.0.0.1"])})
        created_records.append((single, ["10.0.0.1"], TTL))
        got = records_named(single)
        ok = status == "SUCCEEDED" and len(got) == 1 and \
            [r["Value"] for r in got[0]["ResourceRecords"]] == ["10.0.0.1"] and \
            str(got[0]["TTL"]) == str(TTL)
        check("UPSERT creates the record", ok,
              "record present with the right value and TTL" if ok
              else f"status={status} error={error or '-'} cause={cause[:60] or '-'} got={got}")

        # 2. The output shape the caller reads: ChangeInfo, as the SDK returns it.
        info = output.get("ChangeInfo") or {}
        ok = bool(info.get("Id")) and info.get("Status") in ("PENDING", "INSYNC") \
            and bool(info.get("SubmittedAt"))
        check("output is the ChangeInfo shape", ok,
              f"Id={info.get('Id')} Status={info.get('Status')}" if ok
              else f"got keys {sorted(output)}")

        # 3. Two values in one record set: the XML nests as Changes/Change and
        #    ResourceRecords/ResourceRecord, and a flattened body loses one or both.
        values = ["10.0.0.2", "10.0.0.3"]
        status, _, error, cause = run_task(
            {"HostedZoneId": bare_id, "ChangeBatch": batch("UPSERT", multi, values)})
        created_records.append((multi, values, TTL))
        got = records_named(multi)
        ok = status == "SUCCEEDED" and len(got) == 1 and \
            sorted(r["Value"] for r in got[0]["ResourceRecords"]) == values
        check("a multi-value record keeps both", ok,
              "both values present" if ok
              else f"status={status} error={error or '-'} got={got}")

        # 4. The delete side: the record must be gone afterwards. This is the one that failed
        #    silently, since the machine catches DNS errors and continues.
        status, _, error, cause = run_task(
            {"HostedZoneId": bare_id, "ChangeBatch": batch("DELETE", single, ["10.0.0.1"])})
        gone = not records_named(single)
        if gone:
            created_records = [r for r in created_records if r[0] != single]
        ok = status == "SUCCEEDED" and gone
        check("DELETE removes the record", ok,
              "record gone" if ok else f"status={status} error={error or '-'} still={records_named(single)}")

        # 5. The "/hostedzone/<id>" form createHostedZone hands back must be REJECTED, because that
        #    is what real AWS does. boto3 accepts it only thanks to a client-side customisation
        #    (botocore's fix_route53_ids) that Step Functions does not run, so an emulator that
        #    strips the prefix would pass a definition that fails in production.
        status, _, error, cause = run_task(
            {"HostedZoneId": f"/hostedzone/{bare_id}",
             "ChangeBatch": batch("UPSERT", single, ["10.0.0.4"])})
        got = records_named(single)
        ok = status == "FAILED" and error == "Route53.InvalidInputException" and not got
        check("/hostedzone/<id> form is rejected", ok,
              f"{error}" if ok
              else f"status={status} error={error or '-'} records={len(got)} cause={cause[:50] or '-'}")

        # 6. A bad zone id must surface as a route53 error the machine can Catch on, not as a
        #    States.Runtime dispatch failure or a leaked AttributeError.
        status, _, error, cause = run_task(
            {"HostedZoneId": "Z00000000000000000000",
             "ChangeBatch": batch("UPSERT", single, ["10.0.0.5"])})
        # The full name, not just the prefix: real Step Functions reports the SDK exception class,
        # so the "Exception" suffix is what a Catch clause written against AWS matches on.
        ok = status == "FAILED" and error == "Route53.NoSuchHostedZoneException"
        check("bad zone id names the SDK exception", ok,
              f"{error}" if ok else f"status={status} error={error or '-'} cause={cause[:70] or '-'}")

        # 7-8. The same naming rule on the other rest-xml service, against a bucket that does not
        #      exist. Two operations, so the fix is not one op's special case.
        absent = f"ministack-absent-{uuid.uuid4().hex[:12]}"
        for op in ("deleteBucket", "listObjectsV2"):
            status, _, error, cause = run_task({"Bucket": absent}, action=op, service="s3")
            ok = status == "FAILED" and error == "S3.NoSuchBucketException"
            check(f"s3 {op} names the SDK exception", ok,
                  error if ok else f"status={status} error={error or '-'} cause={cause[:50] or '-'}")

        # 9. Control: the suffix is added to the code without disturbing what the failure says.
        status, _, error, cause = run_task({"Bucket": absent}, action="deleteBucket", service="s3")
        ok = "does not exist" in cause
        check("the S3 message survives", ok,
              "message preserved" if ok else f"cause={cause[:60] or '-'}")

        # 10. Control: a code that already ends in Exception is not suffixed twice. headBucket
        #     returns 404 with no body, the path that falls back to a generic code.
        status, _, error, cause = run_task({"Bucket": absent}, action="headBucket", service="s3")
        ok = status == "FAILED" and not error.endswith("ExceptionException")
        check("no double suffix", ok, error if ok else f"error={error or '-'}")
    finally:
        if not args.keep:
            for name, values, ttl in created_records:
                try:
                    r53.change_resource_record_sets(
                        HostedZoneId=zone_id,
                        ChangeBatch=batch("DELETE", name, values, ttl, ttl_key="TTL"))
                except ClientError:
                    pass
            if created_zone:
                try:
                    r53.delete_hosted_zone(Id=created_zone)
                except ClientError:
                    pass

    return 1 if failed else 0


if __name__ == "__main__":
    try:
        sys.exit(main())
    except ClientError as exc:
        sys.exit(2)
```

</details>


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1362, opened 2026-08-12
fix(apigatewayv2): invoke and enforce REQUEST (Lambda) authorizers

ryan-bennett:
## What's broken

HTTP API (v2) routes support two authorizer types: `JWT` and `REQUEST` (Lambda). The data-plane dispatcher (`_handle_execute_in_scope`) only ever branched on `auth_type == "JWT"`:

```python
auth_type = (route.get("authorizationType") or "NONE").upper()
if auth_type == "JWT":
    ...
```

A route referencing a `REQUEST` authorizer carries `authorizationType: CUSTOM` (the authorizer *resource's* own `authorizerType` is `REQUEST`; the route's `AuthorizationType` enum has no `REQUEST` value — confirmed against the `apigatewayv2` `CreateRoute`/`CreateAuthorizer` botocore service models). With no `CUSTOM` branch, such a route fell straight through to the integration exactly like an unauthenticated `NONE` route. The authorizer Lambda was never invoked — not for a missing token, an invalid one, or a valid one — and `event.requestContext.authorizer` was always `undefined` downstream.

## How we found it

We deploy CDK stacks with `HttpApi` + `HttpLambdaAuthorizer` (`aws-apigatewayv2-authorizers`) against ministack locally. The CloudFormation side is fine — `GetAuthorizers`/`GetRoutes` show the authorizer and route wired correctly (`AuthorizationType: CUSTOM`, `EnableSimpleResponses: true`, correct `AuthorizerId`) — but at runtime, requests with a missing, invalid, or valid bearer token all reached the backend integration identically, and the authorizer Lambda showed **zero invocations** in its logs regardless of the request. It's a pure data-plane/request-routing gap, not a CloudFormation one.

This is the same class of bug tracked in #1345 (item 1), which @Nahuel990 closed in 1.4.16 — but that fix (`apigateway_v1._authorize_request_v1`) is REST (v1) only. We re-verified on the current `ministackorg/ministack:full` image that the gap is still present for HTTP API (v2).

## The fix

Adds `_authorize_request_v2` (and supporting helpers: `_evaluate_authorizer_policy`, `_authorizer_arn_matches`, `_build_route_arn`, `_request_authorizer_identity_sources`, `_invoke_request_authorizer_lambda`) to `ministack/services/apigateway.py`, invoked from a new `elif auth_type == "CUSTOM":` branch in `_handle_execute_in_scope`. It mirrors the shape of the already-merged v1 fix, adapted to HTTP API's own conventions:

- **`authorizerPayloadFormatVersion`**: `1.0` builds a REST-shaped event (`methodArn`, `multiValueHeaders`, etc.) and only accepts an IAM policy response; `2.0` builds the newer event shape (`routeArn`, `rawPath`/`rawQueryString`, `identitySource`, `http.*`) and accepts either the simple `{isAuthorized, context}` response when `enableSimpleResponses` is set, or the same IAM policy response format as `1.0` otherwise.
- **IAM policy evaluation**: explicit `Deny` always wins; `Allow` requires a matching statement and no `Deny` — evaluated against the route ARN, same logic as v1's method-ARN evaluation.
- **`authorizerResultTtlInSeconds` caching**: a TTL-cache keyed by `(api_id, authorizer_id, cache_key)`, including v1's missing-identity-source short-circuit (a declared-but-absent identity source is a 401 without invoking the Lambda, only when caching is on).
- **Downstream context delivery**: populates `requestContext.authorizer.lambda` on the `AWS_PROXY` integration event (JWT authorizers deliver under `.jwt` instead), and wires the same context into `$context.authorizer.lambda.*` for `HTTP_PROXY` request-parameter mappings, mirroring the existing JWT-claims mapping support.

## Testing

- Two new data-plane tests in `tests/test_apigatewayv2.py` covering both 2.0 response formats (`enableSimpleResponses` true and the IAM-policy default). Verified both **fail against the pre-fix code** (a `Deny`/`isAuthorized: false` token gets a silent `200` pass-through instead of `403`) and pass with the fix applied.
- Full `pytest tests/test_apigatewayv1.py tests/test_apigatewayv2.py -m "not serial"` run clean except 3 pre-existing failures (1 JWKS/network-dependent JWT test, 1 WebSocket-JWT test, 1 unrelated `provided.al2`-runtime test) that reproduce identically against unmodified `main` in this sandbox — not regressions from this change.
- Manually verified against a from-source `docker compose up --build` of this branch: deployed an HTTP API with a `REQUEST` authorizer (`enableSimpleResponses: true`) in front of an `AWS_PROXY` integration, and confirmed via the authorizer Lambda's own CloudWatch Logs stream that it is invoked exactly once per request — for a missing token, a wrong token (`403`), and a correct token (`200`) — with `requestContext.authorizer.lambda` populated on the integration side.

Addresses (in part) #1345 — item 1's HTTP API v2 gap specifically. The fix that landed in 1.4.16 covered REST/v1 only (see @Nahuel990's own scoping in that comment); the other 7 unrelated items bundled into that original report are out of scope here.


Nahuel990 (approved):

Nahuel990 (approved):

iot-rocket:
Follow-up: https://github.com/ministackorg/ministack/pull/1370 ports the v1 authorizer caching fixes to this surface.

Issue #1355, opened 2026-08-12
Enforce the documented ruff lint gate in CI


iot-rocket:
CONTRIBUTING's PR checklist names `ruff check ministack/` as the lint gate and `[tool.ruff]` is fully configured in `pyproject.toml` (select E/F/I, per-file-ignores for the compact-tokenizer service files), but no workflow runs it — enforcement relies on contributors remembering the checklist, and the tree itself had drifted to 7 violations.

Proposal (we have both pieces ready to PR, plus a separate baseline-fix PR clearing the 7 pre-existing errors):

1. **`.github/workflows/lint.yml`** — `ruff check ministack/ --output-format=github` (inline PR annotations), ruff pinned exactly for reproducibility, following ci.yml's conventions (ubuntu-latest, checkout@v6, setup-python@v6, py 3.13, pip cache, read-only permissions). Triggers: `pull_request` unfiltered (runs on the merge preview, and covers PRs to non-main branches), `merge_group`, and `push` to main/master as the post-merge net. Scope stays `ministack/` — `tests/` carries 31 further violations worth a separate cleanup issue.
2. **`.github/PULL_REQUEST_TEMPLATE.md`** — codifies the Summary / Issue / Fix / Validation shape maintainers already write by hand (#1347, #1341), with the general checklist from CONTRIBUTING and the new-service checklist collapsed in a details block.

Both pieces are implemented and ready to review on our fork:

- CI gate + PR template: [`iot-rocket:feat/ci-ruff-lint`](https://github.com/ministackorg/ministack/compare/main...iot-rocket:feat/ci-ruff-lint) (the compare includes the baseline-fix commit)

Anything you'd want scoped differently before we open the PR?

Nahuel990:
@iot-rocket thanks for this! I’ll take the lint, but not the template. I think it’s a waste of time since many contributors use different AI models to write PRs, and they end up overwriting a lot of things

It’s similar to the bug-reporting template: everyone just goes with the default instead of selecting the appropriate one, which is already an extra step and a bit of a waste of time.

iot-rocket:
Fixed by #1372

PR #1372, opened 2026-08-13
feat(ci): enforce ruff via GitHub Actions and add PR template

iot-rocket:
## Summary

- New `Lint` workflow runs `ruff check ministack/ --output-format=github` on every PR (merge-preview ref), in merge groups, and on pushes to main/master
- New PR template codifying the Summary / Issue / Fix / Validation shape already used by hand, with the CONTRIBUTING checklists (new-service items collapsed)

## Issue

Part of #1355: the documented lint gate had no CI enforcement.

## Fix

- Workflow follows ci.yml's conventions (checkout@v6, setup-python@v6, py 3.13, pip cache) with read-only `permissions`. ruff is pinned to 0.16.2 so results are reproducible, which makes bumping the pin an explicit change
- `pull_request` is left unfiltered on purpose: the run executes against `refs/pull/N/merge` (the PR merged into the latest base at run time), so the would-be merge result is checked, and PRs into non-main branches are covered too

## Scope

`tests/` is excluded; it has 31 pre-existing violations and wants a separate cleanup. To make the PR check gate on latest main, enable branch protection "Require branches to be up to date before merging" (or a merge queue) and mark the `ruff` job required. GitHub doesn't re-run checks when the base moves after a PR's last sync.

## Validation

- `actionlint` clean on the new workflow; `ruff check ministack/` green on this branch (the 7-error baseline fix merged upstream as #1354)
- Expected side effect: this PR touches `.github/`, so docker-publish-on-pr's guard skips the auto preview build and posts the pending-manual-build comment

## References

- Part of #1355
- [ruff GitHub output format](https://docs.astral.sh/ruff/settings/#output-format)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,40 @@
+name: Lint
+
+# ── Triggers ──────────────────────────────────────────────────────────────
+# pull_request has no branch filter on purpose: every PR is linted, and the
+# run executes against the merge preview (refs/pull/N/merge — the PR merged
+# into the latest base at run time), so the would-be merge result is checked,
+# not just the branch tip. merge_group gates the final merged state when a
+# merge queue is in use; push to main/master is the post-merge safety net.
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+  merge_group:
+
+permissions:
+  contents: read
+
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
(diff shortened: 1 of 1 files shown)


Nahuel990:
@iot-rocket thanks for this! I’ll take the lint, but not the template. I think it’s a waste of time since many contributors use different AI models to write PRs, and they end up overwriting a lot of things

It’s similar to the bug-reporting template: everyone just goes with the default instead of selecting the appropriate one, which is already an extra step and a bit of a waste of time.

iot-rocket:
Fair enough, template dropped. The PR is lint-only now: just .github/workflows/lint.yml.

Nahuel990 (approved):

Issue #1278, opened 2026-08-04
Blocking Docker calls on the gateway event loop freeze the whole server


Areson:
### Problem
MiniStack serves every request from a single asyncio event loop. A handful of services (RDS, ECS, EKS, ElastiCache, OpenSearch) call Docker synchronously on that loop — starting, stopping, or pulling a container can take seconds or longer. While any one of those calls runs, the event loop is frozen: every other request, to every other service, waits behind it. Even `/_ministack/health` stops answering.

So this is not a per-service slowdown. One slow container operation makes the entire emulator unresponsive, for however long that operation takes, and services that already do their Docker work correctly in threads are starved along with everything else.

### Where we saw it
The `test_glue_stop_crawler` CI flake (the failure family that #1244  attempted to harden against indirectly). Glue completes crawlers on a ~3-second timer that must fire on the event loop. An RDS test running in the same shard calls 1container.stop(timeout=5)1, which blocks the loop for the full 5 seconds — the timer cannot fire in time and the Glue test fails. Glue is the victim here, not the culprit: it already offloads its own Docker work. We reproduced this at `main`, and it intermittently fails PRs that touch neither Glue nor RDS.
A hung docker pull froze everything. During validation, an ECS test triggered a pull of an image that was not available locally, and the pull stalled. On `main`, that freezes the whole server — no health check, no other service — until the pull returns. 

### Attempted fix
Move each blocking Docker call off the loop with `asyncio.to_thread`, guarded by one `asyncio.Lock` per service so each service's operations stay serialized exactly as they are today — different services just stop blocking each other. This is what #1277 implements, but it is large due to reaching across multiple services and needing additional safeties.

Nahuel990:
@Areson this was released as part of v1.5.0

PR #1377, opened 2026-08-13
fix(lambda): re-entrant handlers no longer deadlock the worker pool

iot-rocket:
## Summary

- Synchronous Lambda executions run on a dedicated thread per in-flight invocation instead of the event loop's fixed default executor, so a handler that calls back into the emulator can no longer deadlock behind the invocations waiting on it

## Issue

Part of #1345 (§8). Every synchronous invoke rode `asyncio.to_thread`, holding one of `MINISTACK_WORKER_THREADS` (default 64) workers for the whole handler. Lambdas get `AWS_ENDPOINT_URL` pointed back at this server, so a handler's nested call that itself needs a worker (another Invoke, an API Gateway / ALB / Function URL execute, a Cognito trigger) queued behind the very invocations waiting on it.

The deadlock threshold is exactly the pool size, measured both ways: an 8-thread pool passes 7 concurrent re-entrant invokes and wedges at 8; the default pool wedged exactly 64 of 80. Even without re-entrancy, one slow function head-of-line blocked unrelated APIs into 502/504 once concurrency exceeded the pool.

## Fix

- `run_invocation_in_thread` spawns a short-lived thread per synchronous execution. A bounded executor was considered and rejected: nested invokes draw from the same pool, so any bound just moves the deadlock edge
- Every synchronous invoke site moves over, including the ones that don't look like Lambda invocations: the REST custom-authorizer invoke, both Cognito trigger dispatchers, the five CloudFormation custom-resource sites (each blocks on a ResponseURL callback into this same server for up to `ServiceTimeout`, 3600 s default), and the AppSync Events Lambda authorizer on both connect paths
- Thread count is bounded by invocations in flight and drains on its own: three 100-concurrency storms peaked at 333 to 358 threads and returned to 4 at rest

## Scope

The shared pool was an accidental brake on runaway self-invocation; removing it is disclosed in the commit, and real `RecursiveLoop` enforcement is a separate follow-up PR. The RIE read-timeout classification and the `host.docker.internal` mapping that earlier drafts of this branch carried have moved to their own PRs.

Relation to #1277 (issue #1278): that draft addresses a different failure mode, Docker calls in RDS/ECS/EKS/ElastiCache/OpenSearch blocking the event loop, and explicitly does not cover Lambda execution, which already runs off-loop. The two overlap at the five CloudFormation custom-resource call sites in `cloudformation/stacks.py` and would conflict textually there. If #1277 lands first, `run_invocation_in_thread` should be reimplemented on top of its `ministack/core/concurrency.run_in_dedicated_thread_to_completion` rather than keeping a second dedicated-thread helper.

## Validation

- A concurrency regression test at pool size + 8 on a private server (`MINISTACK_WORKER_THREADS=8`, 16 invokes): fails 16/16 "Task timed out" on the pre-fix build, completes in ~7 s fixed
- A source-level guard test that derives the set of blocking-Lambda-execution functions (closure over intra-package calls from the three run-a-handler primitives) and flags any `asyncio.to_thread` target inside it. On the pre-fix tree it names exactly the seven converted sites, and reverting any one site fails it
- `tests/test_lambda.py`, `test_cognito.py`, `test_apigatewayv1.py`, `test_cfn.py`, `test_appsync_events.py`, `test_alb.py`: 896 passed / 3 skipped (one environmental failure identical at the base commit)

## References

- [Lambda scaling and concurrency](https://docs.aws.amazon.com/lambda/latest/dg/lambda-concurrency.html)
- Part of #1345 (§8)


Nahuel990:
@iot-rocket No, We had a similar discussion under 1277. We use asyncio to thread which already handles this. We wont have a framework for concurrency

iot-rocket:
I went back through the #1277 discussion. Agreed with the direction there: no concurrency.py, and targeted fixes per blocking call rather than moving whole service routers behind admission locks. That's what this PR does. The helper is a single function in lambda_svc.py next to the executors it serves. There is no new module and no lock, and everything outside the Lambda invoke paths keeps using plain to_thread.

The situation here is different from #1277 though. That was blocking docker calls running directly on the event loop, and to_thread is the right fix for that. In this case the invocations already run on to_thread. The problem is that its pool has a fixed size (MINISTACK_WORKER_THREADS, default 64) while Lambda invocations can be re-entrant: handlers get AWS_ENDPOINT_URL pointed back at the emulator, so a handler that calls Invoke, or hits an API whose authorizer is a Lambda, needs a second pool slot while its own invocation still holds one. Once in-flight invocations reach the pool size, the nested calls queue behind the very callers that are waiting on them, and everything times out. Raising the pool size only moves the edge, since the nested invokes draw from the same pool.

You can reproduce it: set MINISTACK_WORKER_THREADS=8 and fire 8 concurrent handlers that each invoke one more Lambda. 7 pass, 8 all time out. On the default pool it's exactly 64 of 80. The regression test in this PR does that against its own server, and it fails on main.

This also became more relevant: The v2 REQUEST authorizer from #1362  invokes its Lambda through the same pool, so an API whose backend calls into the emulator can deadlock against its own authorizer under load.

If you'd take the same fix in a different shape, tell me what you'd prefer and I'll rework it.

iot-rocket:
Validating #1386 for our use case

iot-rocket:
@Nahuel990 Closing this. #1386 covers it, and v1.5.0 carries it.

Re-checked on the release rather than assuming: `run_reentrant` lives in
`ministack/core/concurrency.py` and both Lambda invoke paths go through it, so the
starvation this PR was opened for is gone on stock main. Eight re-entrant handlers
with `MINISTACK_WORKER_THREADS=8` all complete, where the eighth used to time out.
The static guard we wrote for this class of bug (walk every `asyncio.to_thread` /
`run_in_executor` call, close over intra-package calls by name, flag any whose
target can reach a handler execution) finds no offenders on main either. That guard
is the one piece of this PR with no upstream equivalent. It is a single test
function, no production code. If you want it, say so and we send it as a small PR;
otherwise it stays on our fork.

One housekeeping find while re-reading the release: `CHANGELOG.md` line 77 on main
is a bare `||||||| Stash base` conflict marker, with none of the other diff3
markers around it, followed by three older copies of entries that already sit in
their proper sections above (RDS Aurora engine versions, EC2 public IP/DNS wire
tags, S3 CRC64NVME). They are stranded at the end of the released `[1.4.20]`
section. It came in with #1437 and has now shipped in v1.4.20, v1.4.21 and v1.5.0.
Deleting the four lines is the whole fix;

Thanks for the synthesis in #1386. The instance-scoped reaper in particular removed
a long-standing pain for us: before it, a second MiniStack on the same Docker
daemon swept our containers.

PR #1386, opened 2026-08-14
poc(concurrency)

Nahuel990:


Nahuel990:
@Areson @iot-rocket  Hi! First I did not want to introduce a concurrency framework here... My initial  preference was to fix the specific blocking paths and keep the existing execution model. After running the workload and testing the failure modes it became clear that the two PRs were each solving different parts of the same underlying problem and ixing them independently leaves important re-entrant cases broken or traded a deadlock for unbounded concurrency

I took pieces of each that survived the measurements, rather than adopting either PR wholesale.... This isn't cherrypicking it is taking the parts that the evidence (I ran) supported. I took the re-entrant execution primitive from #1377 andd the event-loop/Docker observation and cancellation semantics from #1277, and then added the missing self-invocation and concurrency-limit cases exposed by the workload (unbound, urllib, sigterm oomkill, orphan containers on failed concurrency)

The resulting PoC branch is less a new concurrency model than an explicit statement of rules that were previously implicit: blocking work that can re-enter MiniStack cannot run on the shared pool and blocking work that cannot re-enter can; nothing blocking runs on the event loop; and concurrency is bounded once at the common execution boundary

Here is exactly how I've broken down:

From #1377: The re-entrant invocation diagnosis & primitive
The diagnosis of the thread pool wedge (deadlocking at exactly the pool size)s correct. Synchronous Lambda executions ride the event loop's shared thread pool, so a handler calling back into Ministack queued behind the invocations waiting on it. I reproduced this independently: at a pool of 64 exactly 64 of 120 failed. Your thread-per-invocation helper survived the measurements and remains essentially intact as `core.concurrency.run_reentrant`

From #1277: Moving Docker work off the loop & cancellation
The observation that Docker work runs on the event loop is correct. I measured the impact: during one ECS RunTask on a cached image, the health endpoint went from 3ms to 7,264ms, making the whole process unreachable while a container starts. Your "shield-and-wait" cancellation semantics were also adopted, ensuring a disconnected client cannot unwind the awaiter while a worker is still mutating state

(Note on the scope of #1277: I initially thought the scope was inflated, but I realize it wasn't and I'm sorry for that. MiniStack had an unwritten invariant relying on hand-rolled coro.send(None) bridges in stepfunctions.py. The moment a handler actually awaits, those crash. Nothing could move off the loop until that was fixed)
--
What changed: Classify by re-entrancy
The key principle going forward is that blocking work is classified by whether it can re-enter the server, not by which service it belongs to

Per-service locks (as proposed in #1277) were rejected because they inevitably build new deadlocks. Every container MiniStack starts is handed an AWS_ENDPOINT_URL pointing back at MiniStack... if a starting container calls in while the dispatch still holds the lock it freezes. Instead of working around re-entrancy per case, work that can call back into Ministack gets a dedicated thread, work that cannot uses the shared pool, and concurrency is accounted for at the single point all executors pass through.


Lasst, the PoC adds 2 missing pieces the workload exposed:

Self-invocation: A handler invoking its own function deadlocked on main and on #1377 alike (timing out at 25.1s) because the bottleneck was a per-function worker lock, not the thread pool. The PoC adds a per-function worker pool, clearing depth-8 self-invocations in 0.9s.

Concurrency limits: ReservedConcurrentExecutions was never enforced outside the Docker path. Removing the shared pool (#1377) removed the only ceiling on in-flight executions. The PoC restores a real bound with AWS-shaped throttling (TooManyRequestsException) (any workload that escaped boto killed the emulator in seconds) 

Thank you both the loop blocking fixes wouldn't have been isolated without your initial PRs forcing the issue.

Please pull the preview build and check if it solves your specific workload 

`ministackorg/ministack-preview-build:pr-1386-3bcaf94e` #1386

Areson:
Here are a few issues that I've confirmed with tests. I'm still working on testing some others.

1. **ElastiCache deferred create publishes a permanently wrong endpoint (cold image).**
When the engine image is not cached locally, CreateCacheCluster returns immediately and a background finisher pulls the image and starts the container. The create path records the real endpoint under one internal key, but the finisher writes a different key — and fills the port with the engine default (6379/11211) rather than the actual allocated host port. All Describe paths read the first key, so once the cluster flips to available it reports an endpoint no one is listening on, and it never self-corrects. Observed live: Describe reported localhost:6379; the real redis container was on host port 16379. A connection to the reported endpoint is refused; the real port answers. Any client that trusts DescribeCacheClusters after a cold-image create cannot connect.

2. **The periodic reaper destroys containers owned by other MiniStack instances on the same Docker daemon.**
The reaper reclaims any exited ministack-labelled container that is not referenced by the current process's in-memory records. There is no instance identity in the labels or filters, so a second MiniStack gateway sharing the daemon treats the first instance's stopped-but-still-owned containers (e.g. a StopDBCluster'd database, which is exited by design and expected to restart later) as abandoned, and removes them after the 120-second grace. Observed live with two gateways: instance A stopped its DB cluster; instance B removed A's container about 3 minutes later (A logged zero reaps; B logged one). A subsequent start on instance A has lost the container. One accuracy nuance: named volumes survive (only anonymous volumes are reclaimed with the container), so the harm is container destruction and loss of any anonymous/tmpfs-backed state — but the stopped cluster is gone either way. This makes running more than one MiniStack instance per daemon continuously destructive.

3. **ElastiCache delete racing the deferred finisher leaks a running container forever.**
The deferred finisher checks that the cluster record still exists and then publishes the container id into it — a check-then-act with no lock. DeleteCacheCluster locates the container to tear down only through that published id, with no fallback lookup by container name. If a delete lands in the window between the finisher's existence check and its publish, the delete removes the record (finding no container id to act on) while the finisher goes on to start or keep the container running — now referenced by nothing. The result is a running container that no API call can ever reach or clean up; only manual Docker intervention removes it. Reproduced deterministically at the exact publish point. For completeness: OpenSearch is not exposed (its delete tears down by stable container name), and RDS has the same non-atomic pattern but a much narrower window with partial recovery on a same-id re-create.

All three are consequences of moving whole handlers off the event loop with no per-resource locking — check-then-act sequences that used to be atomic under loop serialization are now races, and the new concurrency test suite asserts liveness only, with no concurrent same-resource state test.

Areson:
I don't have any other issues to add to the above. I've confirmed that the PR does address the issues I was seeing in my workload based on running my tests.

Areson (commented):
I've confirmed that this now passes all of my tests, including those for the identified issues.

iot-rocket:
@Nahuel990 
Following up on my note on #1377: we validated `6d2c1d8` against our use case
and re-checked after your merge of main (`8530d26`). It works for us. Our
regression test from #1377 passes on your tree, and a build of main with this
branch merged runs our full stack (backend deploy, smoke tier, a real device
SDK binary over mTLS) with no new failures. From our side #1386 supersedes
our #1377, so I keep that one as a draft until this merges, then close it.

One small thing: `firehose.py` keeps a thread target named `_run` (via
`spawn_background`). Fine for the runtime, but any name-based audit trips
over it; #1377 renamed it to `_deliver_iceberg_groups` for that reason.

Thanks for picking the primitive up and carrying it further.

PR #1389, opened 2026-08-15
fix(s3): compute CRC64NVME checksums instead of refusing them

gaul:
CRC-64/NVME is the algorithm current AWS SDKs and the CLI checksum uploads with by default, and PutObject answered it with InvalidRequest on the grounds that it needs a native CRC library. A stock `aws s3 cp` or `aws s3api put-object` therefore failed before it began, unless the caller knew to set AWS_REQUEST_CHECKSUM_CALCULATION=when_required.

The algorithm is plain arithmetic -- the reflected form of polynomial 0xAD93D23594C93659 with all-ones init and xorout -- so compute it from a byte-at-a-time table in the stdlib, adding no dependency. PutObject and CopyObject now validate a client-supplied x-amz-checksum-crc64nvme (BadDigest on mismatch), and Get/HeadObject return it under x-amz-checksum-mode: ENABLED. The tests pin the result to the algorithm's published check value, b"123456789" -> 0xAE8B14860A799888, and to a bit-at-a-time reference rather than to the table itself.

CRC32C still requires the native google-crc32c and is still refused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1393, opened 2026-08-15
fix(alb): stream target responses to match AWS

### LINKED ISSUE #1392: ALB data plane buffers streaming target responses
`_proxy_http_target` reads the whole target response before returning, so a target that streams is delivered to the client only once it has finished generating.

Measured against a local TTS server that streams WAV audio, same request through both:

| path | TTFB | total |
|---|---|---|
| Caddy reverse proxy | 0.013s | 11.54s |
| ministack ALB | 11.28s | 11.33s |

The same applies to SSE and chunked JSON targets.

Source: `ministack/services/alb.py`, `_proxy_http_target` calls `urlopen(...)` then `resp.read()` and returns `(status, headers, bytes)`.

`core.responses.StreamingResponse` already covers this case and `app._send_if_handled` routes it to `_send_streaming_response`, so the change is local to that one function.

Seen on 1.4.17 and on main (99ecd6a).

### Reference: AWS ALB behaviour

- Chunked responses from targets are supported — *"The supported values for **Transfer-Encoding** are `chunked` and `identity`."* ([troubleshooting, HTTP 501](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-501-issues))
- The target's framing is relayed rather than recomputed — *"The target returns a content-length header that is larger than the entity body. The load balancer timed out waiting for the missing bytes."* ([HTTP 504](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-504-issues))
- *"The load balancer forwards valid HTTP responses from targets to the client, including HTTP errors."* ([target errors](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#target-http-errors))
- Connections are bounded by an idle timeout, not a total deadline — `idle_timeout.timeout_seconds`, default 60s ([load balancer attributes](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes))

So a faithful ALB relays the body as it arrives and preserves whichever framing the target chose.

uttom-akash:
Fixes #1392.

`_proxy_http_target` read the whole target response before returning, so a streaming target reached the client only once it had finished producing. It now returns `StreamingResponse`, which the ASGI layer already supports.

```
before   target ──[ read() entire body ]──▶ ALB ──▶ client      first byte at the end
after    target ──[ read1() chunk ]──▶ ALB ──▶ client           first byte immediately
                        └──── repeat ────┘
```

## Changes

- **`services/alb.py`** — return `StreamingResponse`; only status and headers are awaited, the body is pumped with `read1` off the event loop via `asyncio.to_thread`. `read()` blocks until the buffer fills, which leaves SSE-sized chunks stuck.
- **`app.py`** — `_send_streaming_response` no longer strips `Content-Length`, so a proxied response keeps the framing its target chose. Eventstreams set none and still chunk.

## Behaviour

| target | before | after |
|---|---|---|
| chunked / streaming | delivered once complete | delivered as produced |
| `Content-Length` | preserved | preserved, unchanged |
| 4xx/5xx body, unreachable target | — | unchanged |

Measured against a target streaming 700 KB of WAV: time-to-first-byte **11.28s → 0.004s**, byte-identical output.

## Usage

```python
lb = elbv2.create_load_balancer(Name="stream")["LoadBalancers"][0]
tg = elbv2.create_target_group(Name="stream-tg", Protocol="HTTP", Port=8000,
                               VpcId="vpc-00000001", TargetType="ip")["TargetGroups"][0]
elbv2.register_targets(TargetGroupArn=tg["TargetGroupArn"],
                       Targets=[{"Id": "127.0.0.1", "Port": 8000}])
elbv2.create_listener(LoadBalancerArn=lb["LoadBalancerArn"], Protocol="HTTP", Port=80,
                      DefaultActions=[{"Type": "forward", "TargetGroupArn": tg["TargetGroupArn"]}])

resp = urllib.request.urlopen(f"{endpoint}/_alb/stream/events")
resp.read(1)   # returns as soon as the target emits, not when it finishes
```

## AWS reference

- Chunked responses from targets are supported — *"The supported values for **Transfer-Encoding** are `chunked` and `identity`."* ([HTTP 501](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-501-issues))
- The target's framing is relayed, not recomputed — *"The target returns a content-length header that is larger than the entity body. The load balancer timed out waiting for the missing bytes."* ([HTTP 504](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-504-issues))
- *"The load balancer forwards valid HTTP responses from targets to the client, including HTTP errors."* ([target errors](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#target-http-errors))
- Connecting to a target has its own deadline — *"The load balancer failed to establish a connection to the target before the connection timeout expired (10 seconds)."* ([HTTP 504](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-504-issues))
- An established connection is bounded by an idle timeout — `idle_timeout.timeout_seconds`, default 60s ([load balancer attributes](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes))

## Tests

Four in `tests/test_alb.py`:

- `streams_target_response` — four chunks 0.25s apart; asserts first-byte time is well below completion. Buffering makes them equal.
- `non_streaming_body_is_intact` — 256 KiB `Content-Length` body spanning several reads; asserts byte-equality, `Content-Length` preserved, no `Transfer-Encoding`.
- `idle_gap_does_not_truncate_stream` — a quiet period longer than the connect deadline must not end the stream.
- `truncated_target_body_is_not_passed_off_as_complete` — chunked target that dies mid-body; the client must see an unfinished response.

```
39 passed
```

Controls, stated precisely rather than as a blanket claim:

- against `main`, `streams_target_response` fails;
- against the streaming commits before the timeout/truncation fix, `truncated_target_body_is_not_passed_off_as_complete` fails;
- `idle_gap_does_not_truncate_stream` passes against both. It does not reproduce the original truncation — it guards against collapsing the two deadlines back into one.

## Release note

ALB data plane streams target responses instead of buffering them, bounded by an idle timeout rather than the connect deadline.


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1394, opened 2026-08-15
release v1.4.18

Nahuel990:


iot-rocket:
@Nahuel990 
The `[Unreleased]` section currently holds one entry, and thirteen PRs have merged since
the v1.4.18 release commit.

That is not a coincidence. When a PR conflicts with `main`, the conflict is almost always
in `CHANGELOG.md`, because everyone appends to the same `[Unreleased]` block. Merging
`main` into the PR branch and resolving that hunk in favour of `main` removes the
contributor's line, and the squash-merge then lands the code with no changelog entry.

Verified on our five merged IoT PRs. In each case the merge commit has our head as one
parent with the entry present, `main` as the other with it absent, and the merge result
without it:

| PR | merge commit | our parent | main parent |
| --- | --- | --- | --- |
| #1395 device shadows over MQTT | `3e49850` | `4db5ea6` | `6afcdf9` |
| #1396 `sqs` topic-rule action | `bf0283d` | `b50380b` | `1109997` |
| #1397 CA-certificate registry | `61764df` | `6f48f1e` | `fb6ea83` |
| #1398 MQTT 5.0 | `01a3fdc` | `36651ce` | `dbe792d` |
| #1399 `connectivity.*` fleet-index terms | `1fefed7` | `48e32ec` | `7e154da` |

Six entries in total, #1397 had two. None of them is on `main` today.

It is not only us. The same signature, a commit inside the PR touches `CHANGELOG.md`
while the final PR diff does not, shows up on #1364, #1387, #1393, #1404 and #1410.
Before a release this stays invisible, because the release PR rewrites the section by
hand: #1394 wrote 18 entries for v1.4.18, among them reconstructions for #1364, #1387 and
#1393. After a release the entry is just gone, and whoever cuts the next one has to work
out from a commit subject what changed and why it mattered, in someone else's code.

One line fixes the mechanism:

```gitattributes
CHANGELOG.md merge=union
```

There is no `.gitattributes` in the repo yet. With it, git appends both sides instead of
asking for a resolution, so the conflict that causes the loss stops happening. It can
occasionally produce a duplicate or an odd order, which is visible in review and cheap to
fix, unlike a silently dropped line.

If that suits you we will send the PR, and a second one restoring the six entries above
under `[Unreleased]` in the wording we originally submitted. Tell us if you would rather
keep writing the changelog at release time, then we will stop adding entries to PRs and
save you the conflicts instead.

iot-rocket:
That crossed with the release. 1.4.19 went out two minutes after I posted, and all six
entries are in the `[1.4.19]` section with the credit line, including the `406` you
folded into the CA-registry entry. So nothing is missing on main, and the restore PR I
offered is off the table.
...

Nahuel990:
@iot-rocket the problem was that contributors were adding the changelog in previous releases, or writting a long text describing a small change and it became a bit of a mess.

Now, if there’s a changelog conflict, I’d rather take whatever is in main and, at release time, run gh pr --list to get the entries from there. Sometimes I might also merge PRs and apply small corrections on top to avoid the round trip (for instance when a label differs from AWS)

Issue #1343, opened 2026-08-11
IoT Core: Jobs APIs, shadow-over-MQTT, rule SQL `WHERE`/functions, non-Lambda rule actions


iot-rocket:
Hi! We migrated a fairly large IoT platform's local dev/test stack from
LocalStack Pro to ministack and it went remarkably well - ~55 CloudFormation +
Serverless Framework stacks deploy cleanly, Lambda layers and the Docker RIE
executor work, DynamoDB Streams event-source mappings fire with faithful event
shapes, and presigned S3 PUTs verify signed metadata headers correctly.

A few IoT behaviours are the remaining gap for us, and we currently work around
them in a local front-proxy. They fall into two groups: **three correctness
bugs** (where ministack answers successfully but does the wrong thing - these
are the painful ones, because nothing fails loudly) and **missing surface**.

Everything below is reproducible against a stock `-full` image with
`AWS_ENDPOINT_URL` pointed at the gateway.

### 1.1 Topic-rule `WHERE` clauses are ignored - rules over-fire

A rule whose SQL carries a `WHERE` clause fires for **every** message matching
the `FROM` filter, so messages the rule should skip are dispatched to the action.

```bash
aws iot create-topic-rule --rule-name where_probe --topic-rule-payload '{
  "sql": "SELECT * FROM '\''+/+/get'\'' WHERE topic(1) <> '\''maintenancetasks'\''",
  "ruleDisabled": false,
  "actions": [{"lambda": {"functionArn": "arn:aws:lambda:eu-central-1:000000000000:function:probe-fn"}}]
}'

# Should NOT dispatch (topic(1) == 'maintenancetasks'), but the Lambda is invoked:
aws iot-data publish --topic 'maintenancetasks/dev-1/get' --payload "$(echo -n '{"n":1}' | base64)"
```

AWS evaluates the `WHERE` clause and drops non-matching messages. Real rule sets
use this for routing (`WHERE topic(1) <> …`, `WHERE version = 15`,
`WHERE regexp_matches(topic(), '…')`), so over-firing silently sends traffic to
the wrong handlers.

**Suggestion:** if full SQL evaluation is out of scope for now, it would already
help enormously to **reject rule creation with a clear error when the SQL
contains an unsupported clause**, rather than accepting it and ignoring it -
loud beats silent.

### 1.2 SQL functions are silently dropped from projections

`topic()`, `topic(n)` and `timestamp()` work. `clientid()`, `principal()`,
[...]

PR #1398, opened 2026-08-15
feat(iot): MQTT 5.0 support negotiated per connection

iot-rocket:
## Summary

The embedded broker now speaks MQTT 5.0 as well as 3.1.1, negotiated per connection from the CONNECT protocol level. 3.1.1 clients stay byte-identical on the wire. An unsupported level is refused with CONNACK `0x01` instead of being mis-parsed. The v5 UNSUBACK reason codes are derived from the per-filter removal result that #1379 introduced (`0x00` removed, `0x11` no subscription existed), so they cannot mask a wrong removal.

## Issue

Part of #1343. `handle_packet` skipped the protocol-level byte, so a v5 CONNECT was parsed as 3.1.1 (its property-length field misread as payload) and answered with a 3.1.1 CONNACK. The client then died inside its own decoder (`AWS_ERROR_MQTT5_DECODE_PROTOCOL_ERROR`) with nothing server-side to explain it. The AWS IoT Device SDK v2 builds an `Mqtt5Client` by default, so real device binaries could not connect at all. This one cost us some debugging hours before we understood the broker was answering the wrong protocol.

## Fix

The protocol level is read and stored on the session; 4 and 5 are served, anything else is refused with a 3.1.1-shaped CONNACK `0x01`. Its second body byte is the one field every decoder reads as a return code, so the refusal stays diagnosable no matter what the client speaks.

Implemented: property blocks on CONNECT (including will properties), CONNACK, PUBLISH, PUBACK, SUBSCRIBE, SUBACK, UNSUBSCRIBE, UNSUBACK and DISCONNECT; the subscription-options byte (No Local, Retain As Published, Retain Handling 0/1/2); v5 reason codes; an assigned client identifier for an empty client id; and forwarding of the publish properties a server must pass on (payload format, message expiry, content type, response topic, correlation data, user properties). Cross-version delivery works in both directions. A table of property wire-widths lets the decoder step over identifiers it does not model, so an unknown property never derails the parse, and nothing unmodelled is echoed back. The packet builders take a version argument whose default reproduces the existing 3.1.1 bytes exactly.

Session Expiry Interval is honored as an interval (`0xFFFFFFFF` = never), with access-time expiry rather than timers; that trade-off is documented. Advertised in CONNACK as unavailable, and out of scope here: shared subscriptions, topic aliases, message expiry enforcement, flow control, subscription identifiers.

One 3.1.1 wire correction rides along, called out in the commit: retained-message replay now sets the RETAIN flag, per spec §3.3.1.3.

## Validation

New `tests/test_iot_mqtt5.py` covers the CONNACK properties, the unchanged `20 02 00 00` 3.1.1 CONNACK, protocol level 3 refusal, the assigned client id, unknown-property skipping, subscription options, v5 publish round trips with properties, both cross-version directions, QoS 1 PUBACK reason codes, DISCONNECT with reason, and a raw byte-stream test that feeds packets in 3/5/4-byte chunks so the splits land inside the new property fields. Malformed v5 property blocks (truncated value, value bleeding into the next field) answer the spec's `0x81` instead of dying silently. Two will tests pin the v5-only path: an ungraceful transport drop fires the will with its forwardable properties intact, and a truncated will-property block - which precedes the will topic on the wire, so a parse defect would corrupt everything after it - answers CONNACK `0x81`. The UNSUBACK test proves the surviving subscription is still live before asserting `00 11` for one-removed-plus-one-unknown.

`tests/test_iot_mqtt5.py` + `tests/test_iot.py`: 384 passed, 0 failed. Three MQTT 5 tests live in `test_iot.py` so they also run in CI's default matrix. End-to-end, on a merge of this branch with our mTLS-listener branch: a real AWS IoT Device SDK v2 (C++) binary connects, completes the jobs handshake, updates its shadow and disconnects cleanly.

## References

- [MQTT 5.0 spec: CONNECT/CONNACK properties](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901033)
- Part of #1343


Nahuel990:
@iot-rocket I am merging this but I found a preexisting issue (we currently have Maximum Packet Size at 16MB instead of 128 like AWS) so I will be chaning that to match AWS behaviour, also I will move the test of the certificate inside the iot

Nahuel990 (approved):

PR #1399, opened 2026-08-15
feat(iot): connectivity.* terms in the fleet index

iot-rocket:
## Summary

`SearchIndex` now answers `connectivity.connected`, `connectivity.timestamp` and `connectivity.disconnectReason`, and every hit carries the `connectivity` group. Gated as on AWS: the group appears, and its terms become queryable, only with `thingConnectivityIndexingMode: STATUS` in the indexing configuration. `DescribeIndex`'s schema string reflects the mode.

## Issue

Part of #1343 (§2.7). On AWS, "which of my things are online right now" is a first-class fleet-index query once thing indexing is configured with connectivity status. There was no local equivalent, although the broker already holds the truth: it tracks live sessions anyway, to evict duplicate client ids.

## Fix

`connected` is derived from the live session registry on every query, not stored; only the transition timestamp and the last disconnect reason are kept. Deriving means index and broker cannot disagree, and a restart cannot resurrect a stale `connected: true`, because there is no flag to resurrect. A thing counts as connected while a session whose client id equals the thing name is live, which is how AWS ties the two; the docstring states this explicitly.

Deliberate restrictions:

- Only the three disconnect reasons this broker can actually distinguish are emitted: `CLIENT_INITIATED_DISCONNECT`, `DUPLICATE_CLIENTID`, `CONNECTION_LOST`. AWS's keep-alive, throttle and credential-expiry reasons would be guesses, so they are not emitted.
- The group is closed, as on AWS: an unknown leaf answers `InvalidQueryException` instead of silently matching nothing.
- A thing that never connected reports `{"connected": false}` with no timestamp, rather than some epoch implying a disconnect that never happened.

After a takeover the index shows `connected: true` with `disconnectReason: DUPLICATE_CLIENTID` for the previous disconnect, and an ordinary reconnect clears the old reason. That matches what AWS's connectivity index reports.

One separate bug is fixed on the way: `_deregister_client` popped unconditionally, so when a takeover's losing session was later driven through cleanup by its socket close, it evicted the *winning* session from the registry, and a reconnected client silently stopped being tracked. That registry-integrity bug exists on main today; connectivity tracking is just what made it observable.

## Validation

`tests/test_iot.py` + `tests/test_iot_data.py`: 402 passed, also green in random order. The new tests include live MQTT-over-WebSocket cases: found while connected and lost on disconnect, the timestamp advancing, connectivity tied to the client id, a dropped transport reported, a real WS takeover observed as `DUPLICATE_CLIENTID` through the public `SearchIndex` surface, an ordinary reconnect clearing the reason, cross-account and cross-region isolation, connectivity terms answering 400 while the gate is off, and connectivity absent from the persistence snapshot.

## References

- [Fleet indexing: thing connectivity](https://docs.aws.amazon.com/iot/latest/developerguide/managing-fleet-index.html)
- Part of #1343 (§2.7)


iot-rocket:
The failing shard is deterministic for this diff, not a flake: the sink Lambdas in tests/test_iot_data.py run on the 3-second default timeout, and this PR's shard plan seats them right after the k3s-starting EKS tests (server log: 'timed out after 3s - killing worker' on every sink). #1403 is the two-line fix; I'll refresh this branch once it lands.

Nahuel990 (approved):

PR #1402, opened 2026-08-15
feat(bedrock-runtime): add named OrcaRouter proxy preset

XiaoHuo888-hue:
Adds a named [OrcaRouter](https://www.orcarouter.ai) preset to the Bedrock proxy, mirroring how the existing `MINISTACK_BEDROCK_PROXY_URL` OpenAI-compatible path is wired, so the docs and env-var table stay consistent.

Setting `MINISTACK_ORCAROUTER_API_KEY` routes Bedrock `Converse` / `InvokeModel` through OrcaRouter's `https://api.orcarouter.ai/v1/chat/completions` with bearer auth (keys start with `sk-orca-`), defaulting to OrcaRouter's smart router `orcarouter/auto` (`MINISTACK_ORCAROUTER_MODEL` to pin a specific model). Prompts are translated to OpenAI shape, forwarded, and translated back to Bedrock shape exactly like the generic proxy, and it falls back to the deterministic mock on connection error. OrcaRouter is an OpenAI-compatible gateway: one API key unlocks 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen, MiniMax and xAI behind a single endpoint. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

**Verification**

- `pytest tests/test_bedrock.py` — **111 passed** (108 existing + 3 new covering the preset: bearer auth header, `orcarouter/auto` default, `MINISTACK_ORCAROUTER_MODEL` override, and no-auth for the generic path).
- `ruff check` on the changed files — clean.
- L3 live test: with a real key, the Bedrock proxy returns text from `https://api.orcarouter.ai/v1/chat/completions`; an invalid key returns `None` and falls back to the mock silently.

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,6 +739,18 @@ resp = bedrock.converse(
 print(resp["output"]["message"]["content"][0]["text"])
 ```
 
+### Bedrock with OrcaRouter
+
+For a hosted, OpenAI-compatible backend, set `MINISTACK_ORCAROUTER_API_KEY` and MiniStack routes the Bedrock proxy through [OrcaRouter](https://www.orcarouter.ai) (`https://api.orcarouter.ai/v1`) with bearer auth — no `MINISTACK_BEDROCK_PROXY_URL` needed. Prompts are translated to OpenAI shape, forwarded, and translated back to Bedrock shape, exactly like the generic proxy. OrcaRouter also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.
+
+```bash
+docker run -p 4566:4566 \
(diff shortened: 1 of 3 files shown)


Nahuel990:
@XiaoHuo888-hue Thanks for this, but we hold to AWS parity and Bedrock has no third-party router in its API (its routing is first-party prompt routers)... The generic MINISTACK_BEDROCK_PROXY_URL already reaches OrcaRouter anyway. Once Bedrock lands OrcaRouter as a first-party resource, we will happily add it.

PR #1403, opened 2026-08-15
fix(tests): give the topic-rule sink Lambdas a real timeout

iot-rocket:
## Summary

Two-line test fix: `_make_sink_lambda` in `tests/test_iot_data.py` never set a `Timeout`, so every topic-rule sink function ran on Lambda's 3-second default.

## Issue

Five topic-rule tests failed the `parallel 2/3` shard on #1399 twice in a row, each with an empty sink (`assert None == {...}`). The uploaded server log shows the actual cause: every sink invocation died with `timed out after 3s - killing worker`, in the same window where the EKS tests were starting k3s containers on the machine. A worker cold-start cannot reliably beat 3 seconds on a runner under that load. And because the shard plan is derived from the diff, the same neighbors line up on every rerun - the failure is deterministic for that PR, not a flake you can retrigger away.

## Fix

`Timeout=30`, matching the authorizer test helpers, which set it for exactly this reason. A passing run is not one bit slower; only the kill threshold moves.

## Validation

`tests/test_iot_data.py` against a local server: 31 passed, including the five tests from the #1399 failure set.


iot-rocket:
The red shard here is the appsync publish-ack flake that #1401 fixes - the two PRs cover each other's failure modes (this one fixes the sink timeouts behind #1399's red check). Merging #1401 first and rerunning this makes both deterministic; happy to rebase if you prefer them squashed into one.

Nahuel990:
@iot-rocket thanks for this, I will close it as I moved this test to _SERIAL_TEST to avoid the 30s timeout (it solves on cold start), If it fails I will refactor to get one shared warm sink lambda to avoid the race

PR #1405, opened 2026-08-15
fix(cloudwatch): compute extended-statistic percentiles instead of al…

MGSousa:
`GetMetricData` _stat_value had a stub for extended-statistic percentiles that unconditionally returned `Average`, regardless of which percentile was requested. Querying the same dataset with the `pNN` stats, returned identical metrics, but without any errors.

The same gap existed one level up in alarm evaluation: `_evaluate_alarm` only ever read an alarm's Statistic field, never `ExtendedStatistic`, so a `PutMetricAlarm` configured with a percentile threshold was silently evaluated against `Average` instead.

Now will be possible to query properly the following stats: `p50`, `p95`, and `p5`.

### Fix

- Added function `_percentile(values, pct)` to sort ascending and linearly interpolate between the two nearest ranks (matches e.g. numpy's default linear interpolation).
- `_stat_value` now takes the raw per-period sample values and computes real percentiles via _percentile, falling back to Average only when raw values aren't available.
- Fixed both call sites (GetMetricData's per-bucket loop, and _evaluate_alarm) to pass the raw sample list through.
- `_evaluate_alarm` now prefers `ExtendedStatistic` over Statistic when set, instead of ignoring it.

### Tests

- Computed expected percentiles for a known dataset ([1, 2, 3, 4, 90, 95, 98, 99]) and confirmed the fixed implementation matches exactly: `p50 = 47.0`, `p95 = 98.65`, `p5 = 1.35`, `p99.9 = 98.993`.
- Rebuilt the Docker image from this branch and tested directly to confirm end-to-end funcionality.
- Added function `test_cloudwatch_get_metric_data_extended_percentile` to `tests/test_cloudwatch.py`, asserting those hand-computed values and that p95 != p5, since they collapsed to the same value before this fix.
- Added more tests to `test_cloudwatch.py` suite: 32/32 pass (31 pre-existing + 1 new), no regressions.
- Ran `ruff check ministack/` cleanly.


Nahuel990:
@MGSousa thanks for this, yes it returns an average as stub because AWS doesn't disclose how is calculated... where did you get this formula?

MGSousa:
> @MGSousa thanks for this, yes it returns an average as stub because AWS doesn't disclose how is calculated... where did you get this formula?

Hey @Nahuel990,
What I did was apply linear interpolation between the two nearest ranks, as we have for NumPy's default percentile; hence, the tests I ran locally with simulated data with this formula were also very similar to AWS CloudWatch's real metrics.

Nahuel990:
@MGSousa I am merging this becuase is a way to have those records but this is not AWS shaped either. I will update the label before release (still saying average) Thanks!

Nahuel990 (approved):

PR #1406, opened 2026-08-16
fix(cfn): apply AWS::IoT::Policy updates instead of rolling the stack…

maximoosemine:
### What

**CloudFormation — `AWS::IoT::Policy` updates apply instead of rolling the stack back.** The type shipped with a create and a delete handler but no update handler, so a stack update fell through to `_update_resource`'s create fallback, called `CreatePolicy` for a policy the same stack had already provisioned, and got `ResourceAlreadyExistsException` back. The stack reached `UPDATE_ROLLBACK_COMPLETE` with the policy document unchanged — an IoT policy could be deployed once and then never edited, which is the ordinary way authorization gets tightened as a device fleet grows.

A changed `PolicyDocument` is now the no-interruption update AWS documents: the document is stored through `CreatePolicyVersion` with `setAsDefault`, so the policy keeps its name and ARN, `Ref` keeps resolving to the same value, `GetPolicy` returns the new document under a bumped `defaultVersionId`, and the previous version stays listed in `ListPolicyVersions` the way it does on AWS. A changed `PolicyName` is a replacement: the policy is created under the new name and the old one removed, so a rename does not leave an orphan behind.

The handler is registered with `update_with_logical_id` so an auto-named policy (no `PolicyName` in the template) derives the same deterministic `_physical_name(stack_name, logical_id)` on update that it did on create, rather than being seen as a rename on every update.

### Testing Plan

- Two tests added to `tests/test_cfn.py`: a `PolicyDocument` change asserting `UPDATE_COMPLETE`, an unchanged `Ref`, the new document readable through `GetPolicy`, `defaultVersionId` of `2`, and both versions listed; and a `PolicyName` change asserting the new name is live, `Ref` follows it, and the old name answers `ResourceNotFoundException`.
- Reproduced the failure first against `main` — the update reached `UPDATE_ROLLBACK_COMPLETE` with the document still on version 1 — and confirmed both tests fail there and pass with the fix.
- `pytest tests/test_cfn.py tests/test_iot.py` passes (535 tests, 2 skipped).
- `ruff check ministack/` passes with no new findings.


iot-rocket:
Can confirm this one from our side. `AWS::IoT::Policy` was part of what we
reported in #1345: back then the type was rejected outright, and since #1373
creates work, but a stack update still fell through to create and rolled the
stack back. That has kept us deploying IoT policies through a CLI seed script
next to the stack instead of inside it, since a policy document you can never
update again is not usable for iterating on device permissions.

Reproduced both sides on 1.4.18: pristine image rolls back with
`AWS::IoT::Policy create failed: ResourceAlreadyExistsException` and the
document stays on version 1; with this branch's `provisioners.py` the same
update reaches `UPDATE_COMPLETE`, `GetPolicy` returns the new document with
`defaultVersionId` 2, and `Ref` keeps the name. The rename-as-replacement
behaviour also matches what CloudFormation does on AWS.

One observation, not a blocker: on real AWS, `CreatePolicyVersion` fails with
`VersionsLimitExceeded` once a policy has 5 versions, and the CloudFormation
handler deletes the oldest non-default version first to stay under the cap.
MiniStack's `_create_policy_version` doesn't enforce the cap today, so
repeated updates keep working here. If the cap ever gets implemented, this
update path would start failing on the 5th document change unless it prunes
the same way. Might be worth a code comment or a follow-up.

`AWS::IoT::ThingType` has the same shape by the way: create and delete
handlers but no update, so a changed thing type property will hit the same
create-fallback rollback. Different resource, so not something for this PR,
just noting it while looking at the table.

Thanks for fixing this.

maximoosemine:
@iot-rocket Thanks for the feedback. I’ll take a look at your additional concerns.

maximoosemine:
@iot-rocket Your feedback should be addressed in the additional commit on this branch.

Nahuel990 (approved):

PR #1407, opened 2026-08-16
fix(cognito): build federation endpoints from the user pool domain

tema-mazy:
## Problem

  `CreateUserPoolDomain` accepts `CustomDomainConfig` and discards it, always expanding
  `Domain` to `{Domain}.auth.{region}.amazoncognito.com`. Separately, `_acs_url()` and
  `_oidc_callback_url()` ignore the pool's domain entirely and are built from
  `MINISTACK_HOST`/`GATEWAY_PORT`.

  So federated sign-in hands external IdPs
  `redirect_uri=http://localhost:4566/oauth2/idpresponse` — unreachable from the browser,
  and rejected by Google and other public IdPs for the scheme alone. The flow dies at the
  hop to the IdP with no way to correct it.

  ## Solution
  
  This is the design @Nahuel990 specified in #982: keyed off `CustomDomainConfig`, built
  from the configured domain — no env var, no Host-header derivation.

  `CustomDomainConfig` now distinguishes the two domain kinds:
  
  | | stored as | serves | CloudFrontDomain |
  |---|---|---|---|
  | with `CustomDomainConfig` | `CustomDomain` | the FQDN as-is | `d{hash}.cloudfront.net` |
  | without | `Domain` | `{prefix}.auth.{region}.amazoncognito.com` | regional host |

  `/saml2/idpresponse` and `/oauth2/idpresponse` are then derived from whichever the pool
  has, at all three sites: the `AssertionConsumerServiceURL` in the generated SAML
  `AuthnRequest`, the OIDC authorize `redirect_uri`, and the `redirect_uri` replayed on the
  back-channel token exchange (the last two must match or the IdP rejects the exchange).

  Configuring one is the same call you would make against AWS:

  ```bash
  aws cognito-idp create-user-pool-domain --user-pool-id $POOL \
    --domain auth.example.com \
    --custom-domain-config CertificateArn=arn:aws:acm:...
  ```

  The certificate is recorded, never served — TLS for a custom domain is terminated by
  whatever fronts it (nginx, an ALB, CloudFront) and proxied to the gateway, which is the
  topology the domain exists to describe.

  `DescribeUserPoolDomain` returns the real `CustomDomainConfig` back, and
  `DescribeUserPool` reports `CustomDomain` as its own field alongside `Domain`, matching
  the AWS shape.

  ## Backward compatibility

  A pool with no domain still falls back to the local gateway, so existing setups are
  byte-for-byte unchanged. A pool holds one domain of each kind independently — deleting
  one leaves the other. A `CustomDomainConfig` with no `CertificateArn` is an
  `InvalidParameterException`.

  ## Tests

  7 new tests in `tests/test_cognito.py`: custom create/describe, prefix-domain regression,
  missing `CertificateArn`, prefix+custom independence, URL derivation across all three
  states, plus end-to-end guards that the OIDC `redirect_uri` and the ACS URL in the decoded
  SAML `AuthnRequest` carry the custom domain.

  236 Cognito tests pass, plus 510 across cfn/persistence/tagging/package.
  
  Closes #982.


Nahuel990 (approved):

Nahuel990:
@tema-mazy thanks for this! the shape looks correct but I will modify a few things after merge and before release (aws never returns amazoncognito.com domain)

tema-mazy:
@Nahuel990 thank you. 

you're right: 

Issuer / Metadata URL: https://cognito-idp.{region}.amazonaws.com/{userPoolId}
OpenID Configuration: https://cognito-idp.{region}.amazonaws.com/{userPoolId}/.well-known/openid-configuration
JSON Web Keys (JWKS): https://cognito-idp.{region}.amazonaws.com/{userPoolId}/.well-known/jwks.json
Hosted UI Login URL: https://{your-cognito-domain}/oauth2/authorize
Token Endpoint: https://{your-cognito-domain}/oauth2/token
Logout Endpoint: https://{your-cognito-domain}/logout

PR #1408, opened 2026-08-16
fix(readme): Fix BedRock proxy URL example

JoseExposito:
The documentation example for MINISTACK_BEDROCK_PROXY_URL uses a URL finishing in "/v1".
However, the service, appends "/v1/chat/completions" to the proxy URL, making the documented URL return 404.

Update the documentation to reflect the expected value.


Nahuel990 (approved):

Nahuel990:
Thanks

PR #1422, opened 2026-08-16
feat(iot): opt-in native mTLS MQTT listener on a TCP port

iot-rocket:
## Summary

Setting `IOT_MTLS_PORT` binds a TCP listener that requires a client certificate, so real device binaries connect unmodified. It is disabled unless the variable is set; nothing changes for existing WebSocket clients.

## Issue

Part of #1343 (§2.6). The broker was MQTT-over-WebSocket only, but device SDKs connect with mutual TLS on 8883 and cannot use a WebSocket transport. Device-side integration tests therefore could not point at MiniStack directly; we had to bridge through a local mosquitto instead, which is exactly the kind of moving part an emulator should make unnecessary.

## Fix

New `ministack/services/iot_mtls.py`, started from the app lifespan like the Transfer Family SFTP listener: lazy import, warn-and-continue on bind failure, opt-in via env. The server certificate is minted from the existing Local CA (already served at `GET /_ministack/iot/ca.pem`) with SANs for localhost, `MINISTACK_HOST` and `IOT_MTLS_EXTRA_SANS`; `CERT_REQUIRED`, TLS 1.2 minimum.

The peer certificate is mapped to its `certificateId` (SHA-256 of the DER, the same derivation `RegisterCertificate` uses) and from there to the account/region that registered it. The tenant is decided by identity, not by a credential in a URL. An unknown or non-ACTIVE certificate gets MQTT CONNACK `0x05` and the connection is closed. That deliberately diverges from AWS's silent close: an emulator should say why.

The existing session and broker machinery is reused unchanged; the only transport-specific part is a small adapter translating the two send shapes. Pub/sub, retained messages, LWT, QoS 1 and topic rules behave identically on both transports.

## Scope

Trust anchors are the Local CA plus whatever the IoT CA registry holds, read on every handshake through `trust_ca()`, so a CA registered after the listener bound is trusted without a restart. A certificate onboarded via `RegisterCertificateWithoutCA` cannot complete the handshake, because there is no issuer to chain to. `DescribeEndpoint` still returns the gateway address; devices point at `<host>:<IOT_MTLS_PORT>` explicitly.

## Validation

New `tests/test_iot_mtls.py`, 15 passed, each on a private server (the `test_tls.py` pattern): disabled by default, chain and SAN verification against the served CA, cross-transport publish/subscribe, `CERT_REQUIRED` enforcement, unknown and inactive certificates refused with `0x05`, account-scoped delivery, garbage bytes tolerated, duplicate client id evicted, and survival of `reset()`. The registered-CA chain case runs here too now that #1397 is on main: a leaf under a CA registered after the listener bound connects, and the same leaf under an INACTIVE CA is refused below MQTT.

Details that took real debugging: shutdown and reset are bounded because the listener socket closes synchronously and live sessions are `abort()`ed; a closing TLS transport otherwise waits for a `close_notify` that a stuck device never sends. `reset()` with a device attached completes in about 0.2 s, pinned by a test. Certificate-to-tenant mapping requires exactly one ACTIVE registration: the same PEM registered in two accounts is refused with `0x05` and a warning naming the ambiguous scopes, instead of lexicographically awarding the session to the lowest account id.

Every new test was first verified to fail on the unfixed listener. The MQTT packets in the tests are hand-rolled from the stdlib, so there is no new dependency.

## References

- [AWS IoT device connection (mutual TLS, port 8883)](https://docs.aws.amazon.com/iot/latest/developerguide/protocols.html)
- Part of #1343 (§2.6)


Nahuel990:
@iot-rocket thanks, but this one is out of shape... MiniStack accepts every credential and validates none, on purpose. IoT policies are stored but not enforced, there is no SigV4 anywhere, and the WS broker accepts any client. This PR ask for CERT_REQUIRED, refuse any cert that is not registered and ACTVE with a tenant decided by cert identity

Also AWS exposes no API for iot_mtls, so there is nothing to emulate here in the botocore sense, and the change adds an env var for what is transport-layer infrastructure. Support of MQTT over WebSocket was added only because it is the minimum needed to exercise the IoT pub/sub data plane, rules, shadows and fleet index, from test code, all on the gateway port. A native mTLS listener on its own TCP port is device-binary infra, not an API gap

iot-rocket:
@Nahuel990 
Fair point on the policies. The listener does not enforce IoT policy documents and I am
not proposing that it should.

Two things I would like to check before I drop this.

The pattern exists in the tree already. Transfer Family runs a real SFTP listener on its
own TCP port, configured by `SFTP_PORT`, `SFTP_HOST`, `SFTP_PORT_PER_SERVER` and
`SFTP_BASE_PORT`, and it is default-on when asyncssh is installed. It also authenticates:
`validate_public_key` in `transfer.py` scans every user across every server and account
and returns False for a key it does not find. The README lists that listener as a feature.
There is no botocore API for the SSH wire protocol either. On SigV4, `s3.py` has
`_verify_presigned_sigv4`, which recomputes the HMAC and answers 403
`SignatureDoesNotMatch`, and it runs on every S3 request rather than behind a flag. IoT
also validates X.509 already: `RegisterCertificate` refuses a leaf that its claimed CA did
not sign. The ACTIVE/INACTIVE lifecycle my listener reads was there before me.

Second, the IoT Phase 1 entry in the changelog (1.3.43) lists the deferred work as "Device
Shadows, mTLS on 8883, ListRetainedMessages queries, Rules Engine, Jobs, Fleet
Provisioning". Shadows and the rules engine have shipped since, Jobs is open as #1400, and
this is the mTLS item from the same sentence. The same entry says plain TCP 1883 is
deliberately not exposed because real AWS IoT requires TLS or SigV4 on every connection,
which I read as an argument for the TLS port rather than against it. And `app.py` already
documents `GET /_ministack/iot/ca.pem` as being there so IoT SDKs can configure trust for
mTLS connections to the local broker. Right now there is nothing on the other end of that.

Your point about strictness is fair on its own terms though. The WS path takes any client
and falls back to `MINISTACK_ACCOUNT_ID`, and my listener does not. I can bring it in
line: `CERT_OPTIONAL` instead of `CERT_REQUIRED`, no trust anchor enforcement, and an
unknown or ambiguous certificate falling back to `MINISTACK_ACCOUNT_ID` exactly like
`_ws_resolve_iot_account_id` does. The strict behaviour would move behind an opt-in flag,
same shape as `APPSYNC_EVENTS_ENFORCE_AUTH` and `MINISTACK_IMDS_V2_REQUIRED`. Then it is
pure transport: same broker, same semantics, same tenancy rule as WebSocket, just
reachable by a device binary.

Would you take it in that shape? If the answer is still no I will close it and we carry it
downstream. Either way the IoT work so far has been worth it.

Nahuel990:
@iot-rocket about this... AWS uses one API (botocore) to expose the control plane of all their services (if you deserialize their API you'll see iot is composed by 3 main services (iot, iot-data and iot-jobs-data) and several other iot-family services... Ministack keeps 1 file per service exposed in botocre + 1 test file per service (and a few other test files when the test is tranversal to Ministack itself) the design is not impressive but it helps us locate issues in an easy way and prevents us from people who just wants to have a PR with their name create random PRs without base (long files consume long context in many AI models so it forces contributors to read what they are touching) 

About the SigV4 you are correct about the S3 its a fix we introduced a few weeks ago, and it's the only enforced because the user can presign the URL with whatever they want without an IAM (it's just decode). I've been considering adding IAM for core services but is not something will land for at least 1 month until I can draft which services to cover first as the original intended scope was to replace LS Community only, we don't want to claim IAM Auth and just validate shapes (and preventing again random contributors adding PRs without knowleedge in AWS and just trusting in AI) 


On the env vars side, that is something I don't like, emulator should be easy to handle wihtout a learning curve (the ideal goal is click and deploy from MS to AWS), we do have knobs to support migration from Localstack but the main idea is to reduce the use of them.


So again my original message wasn't a rejection on the feature itself but asking more to shape it like mqtt, inside it corresponging service, thinking if the env vars are really needed or if the values will be a default and you just added them as a preventive measure.

iot-rocket:
@Nahuel990 Thanks, that clears it up. I read "out of shape" as a no to the feature, not
to its packaging. Your last comment made that clear to me.

One file per service: agreed, and done. `iot_mtls.py` is gone, the listener
sits at the bottom of `iot.py` next to the WebSocket path and shares
`_WSSession` and the framing helper, so both transports keep identical packet
semantics. Same for the tests: they moved into `test_iot_data.py`, where the
broker's transport tests already are, and reuse its MQTT codec instead of
carrying a second copy, the way you folded the MQTT5 tests in #1413.

On the env vars: agreed, and this is the part I got wrong. `IOT_MTLS_HOST`
and `IOT_MTLS_EXTRA_SANS` are deleted, the broker certificate derives its
names itself now.

For the two that are left I followed the SFTP listener, because the WebSocket
transport has nothing to copy: it has no port of its own, it is an upgrade on
the gateway port, and it cannot be turned off. So the listener is on whenever
`cryptography` is importable, `IOT_MTLS_ENABLED=0` turns it off the way
`SFTP_ENABLED` does, and `IOT_MTLS_PORT` moves it off 8883. A port that
cannot be bound degrades to a log line instead of failing the boot, which
matters here because 8883 is a popular port and a local broker may already
hold it. I first had the port variable double as the off switch, but that is
not a convention this repo has, and 0 already means "pick an ephemeral port"
to anyone reading a socket call.

Everything the two transports can share, they do share: the same
`_WSSession`, the same framing helper, the same `IOT_WS_FRAME_MAX_BYTES` cap,
and the same tenancy fallback for a client that presents no credential.

Two more things from a second read of my own code. A refused MQTT 5 client now
gets a real v5 CONNACK (reason code 0x87 plus the property block) instead of
the two-byte 3.1.1 form its decoder would choke on — the accepted path already
negotiated the version per connection, the refusal path now does too, and a
test pins it. And the server-certificate minting (RSA keygen) runs in a worker
thread now instead of on the event loop during startup; plain
`asyncio.to_thread`, since it is a one-shot job — the unbounded-thread
primitive #1386 adds is for re-entrant Lambda invocations and would be the
wrong tool here.

On SigV4 and IAM: understood, and I am not asking for either. No IoT policy is
evaluated and no credential is verified.

That leaves the client certificate, and there I would like to keep one piece
of strictness, because I think it sits on a different layer than the one your
objection is about. A certificate's status is not IAM and not SigV4. SigV4 is
another credential type, IoT policies are authorization that MiniStack stores
and does not evaluate, and ACTIVE/INACTIVE is the X.509 identity lifecycle,
which this service already models and already enforces on the control plane:
deleting an ACTIVE certificate answers CertificateStateException 406, same for
a CA.

So the rule is now the same one `s3.py` applies to a presigned URL, which you
described yourself: a caller that presents no credential is served, and a
caller that presents one that does not hold up gets refused. Concretely, a
client with no certificate connects and runs under `MINISTACK_ACCOUNT_ID`,
exactly like an unsigned WebSocket upgrade, so nothing is required of anyone.
A client that does present a certificate has it read against the registry:
ACTIVE in exactly one account serves that account and region, and unknown, not
ACTIVE, or ACTIVE in two accounts gets a "not authorized" CONNACK (0x05, or
0x87 to an MQTT 5 client). The practical reason is
that `UpdateCertificate` to INACTIVE should mean something on the wire,
otherwise deactivating a device's certificate leaves it publishing, just into
a different account. Real AWS IoT refuses those connections too. The double
registration cannot be attributed at all, since both registrations see
byte-identical bytes, so awarding it would just hand the session to whoever
registered the copy; it is refused and the log names both scopes.

If you would rather have no refusals at all on this transport, say so and I
will make the unattributable cases fall back to the default account as well.
It is a small change, and it needs no flag either way.

Head is 3ed0007 on top of v1.4.20. Is that the shape you had in mind?

Nahuel990 (approved):

Nahuel990:
Yes looks great now!

PR #1423, opened 2026-08-17
fix(apigateway): route on the resource+method pair, not the resource alone

iot-rocket:
## Summary

A REST API answers `403 Missing Authentication Token` for any request whose verb the matched resource does not declare, even when a `{proxy+}` elsewhere in the tree is wired to serve it. Only an exact resource+method match should keep a request on the specific resource.

## Issue

Part of #1345 (§1). This reverses the 1.4.16 change that made a methodless resource answer `403`, so it needs the measurement up front. That change was made on our report, and our report was too narrow: we described a methodless intermediate because that is what we hit, and assumed a declared resource otherwise wins.

Measured on real API Gateway, eu-central-1, MOCK integrations, no authorizer, `authorizationType NONE`. Tree: `/{proxy+}` `ANY`, `/jobs` with no methods, `/jobs/{operation_type}` `GET`, `/cors_only` `OPTIONS`.

```
GET  /jobs/reboot   200 {"served_by":"jobs-op"}   exact resource+verb match wins
GET  /jobs          200 {"served_by":"proxy"}     node with no methods
POST /jobs          200 {"served_by":"proxy"}     any verb
POST /jobs/reboot   200 {"served_by":"proxy"}     declared verb mismatch
GET  /cors_only     200 {"served_by":"proxy"}     OPTIONS-only node
```

So resolution is on the pair. The middle case matters most in practice: CDK and Serverless put a CORS `OPTIONS` preflight on every intermediate, so the node is rarely method-less, it simply does not serve the verb.

There is a consequence worth stating rather than discovering later. Routing precedes authorization, so a guarded method does not hold the path: with `/admin POST` at `AWS_IAM` and an open `{proxy+} ANY`, an unsigned `GET /admin` is served by the proxy. Measured, not assumed, because the instinct is to "fix" it. MiniStack currently refuses that request, which is safer than AWS and less faithful.

## Fix

`_match_greedy_fallback` runs whenever the matched resource has no method for the verb, not only when it declares none, and picks the most specific `{proxy+}` whose prefix matches and that serves the method. Literal prefix segments must match exactly, `{param}` segments capture. When nothing serves the verb the `403` stands.

## Validation

- `tests/test_apigatewayv1.py`: 123 passed. Three cases cover the methodless node, the CORS-preflight-only node and the guarded sibling; all three fail on the current resolver.
- `tests/test_apigatewayv2.py`, `tests/test_cfn.py`, `tests/test_persistence.py`: 554 passed, since the resolver is shared.
- The probe script that produced the table above is reproducible against any account; happy to attach it if useful.

## References

- Part of #1345 (§1)
- [AWS: API Gateway method request and greedy path variables](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-request.html)


iot-rocket:
The red shard is not this diff. It failed on `tests/test_dsql.py::TestClusterLifecycle::test_list_clusters_max_results_out_of_range` with `TimeoutError` after 10s, waiting for response headers on a raw urllib call to `/cluster?max-results=0`. That endpoint answers 400 in about 30ms against an idle server, and this PR only touches `apigateway_v1.py` and its tests.

Shard 2 of that run was the container-heavy one: two RDS postgres instances (11.8s and 8.8s), a CFN stack with S3, Lambda and DynamoDB (16.1s), a DynamoDB stream to Lambda, ECS tasks. That looks like #1278, where a synchronous Docker call freezes the event loop and every other request waits behind it, including ones that touch no containers at all. The DSQL test just happened to ask during the freeze.

Since the shard plan follows the diff, this lands on whichever PR happens to co-schedule those neighbours. I will retrigger, but flagging it so the red check is not read as this change.

Nahuel990:
@iot-rocket Sorry but I will need to check this against AWS myself, you introduced that change originally. I dont want to waste time

iot-rocket:
@Nahuel990 
fair enough that you want to verify it yourself.

And part of the mixup is
ours anyway: the strict 403 from 1.4.16 goes back to our #1345 report.

We wrote there that aws_iam methods "should reject requests without a SigV4
Authorization header - currently they're served", and that was too broad.

It is correct for a request to the guarded method itself. For any other verb on
the same path, real API Gateway routes to the sibling {proxy+} and serves it.

That bullet came from the emulator symptom plus the routing docs, before we
had a real AWS account to measure against. And it turns out the docs already
answer the routing question, just more precisely than either of us quoted
them. The actual sentence is: "a method request against a specific resource
takes precedence over a method request against a generic resource at the same
level of the resource hierarchy"
(https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-request.html#api-gateway-proxy-resource).
Precedence is defined per method request, not per resource, and the routing
table on that page keys every row by verb plus path. A methodless /jobs has
no method request, so there is nothing to take precedence. We both shortened
that sentence to "a defined resource takes precedence" at some point, and
that shortening is the whole disagreement.

So we measured instead of debating:
The attached script builds a throwaway
REST API on MOCK integrations, no Lambda involved: /jobs/{operation_type}
GET, which makes /jobs one of those auto-created methodless nodes, an
OPTIONS-only /cors_only, a POST /admin guarded with AWS_IAM, and an open
{proxy+} ANY. Then six unauthenticated probes, and it deletes the API again.
Re-run today against a clean account in eu-central-1:

    GET  /jobs/reboot -> 200 {"served_by":"jobs-op"}   (control)
    GET  /jobs        -> 200 {"served_by":"proxy"}     (methodless node)
    POST /jobs        -> 200 {"served_by":"proxy"}
    POST /jobs/reboot -> 200 {"served_by":"proxy"}     (leaf, undeclared verb)
    GET  /cors_only   -> 200 {"served_by":"proxy"}     (OPTIONS-only node)
    GET  /admin       -> 200 {"served_by":"proxy"}     (sibling POST is AWS_IAM)

That covers all three shapes this PR implements.

The last probe also pins the part that surprised us: routing happens before authorization, so an IAM guard
on a sibling method does not hold the path. The PR body says so. The script
needs nothing but aws-cli credentials and refuses to run against an emulator
endpoint.
[proxy-fallthrough.sh](https://github.com/user-attachments/files/31203203/proxy-fallthrough.sh)

Nahuel990:
GET  /jobs/reboot     -> 200   {"served_by":"jobs-op"}   [sanity: exact match, expect 200 jobs-op]
GET  /jobs            -> 200   {"served_by":"proxy","path":"/prod/jobs"}   [methodless node]
POST /jobs            -> 200   {"served_by":"proxy","path":"/prod/jobs"}   [methodless node, any verb]
POST /jobs/reboot     -> 200   {"served_by":"proxy","path":"/prod/jobs/reboot"}   [declared GET-only, verb mismatch]
GET  /cors_only       -> 200   {"served_by":"proxy","path":"/prod/cors_only"}   [OPTIONS-only node]

Yup same result, thanks!

PR #1424, opened 2026-08-17
fix(ec2): emit the public address under the real wire tags, and generate valid ones

iot-rocket:
## Summary

`DescribeInstances` and `RunInstances` write the public address under `publicDnsName` / `publicIpAddress`, but the EC2 query protocol names those members `dnsName` / `ipAddress`. SDKs drop members they cannot map, so both fields are missing from every response even though the server generated them.

## Issue

botocore's `ec2-2016-11-15` model is explicit:

```
PublicDnsName    locationName = dnsName
PublicIpAddress  locationName = ipAddress
```

Against a stock server, `run_instances` followed by `describe_instances`:

```
PublicIpAddress present: False
PublicDnsName   present: False
PrivateIpAddress: 10.0.212.225      <- correctly tagged, comes through
```

MiniStack already generates the address (`_random_ip("54.")` on `RunInstances`), so nothing is missing except the ability to read it. Code that waits for an instance address, or feeds one to SSH or a health check, has nothing to wait for.

Fixing the tags made a second defect reachable, so this PR fixes both rather than shipping one into view. `_random_ip` appends exactly two octets regardless of prefix:

```
_random_ip("10.0.") -> 10.0.104.37     four octets
_random_ip("52.")   -> 52.55.218       three
```

`AllocateAddress` returns that shape as `PublicIp` today, where it is already visible: `ipaddress.ip_address("52.55.218")` raises. The generator now completes any prefix to four octets.

## Fix

- the two tags become `dnsName` and `ipAddress`
- `_random_ip` fills to four octets instead of appending a fixed two

## Validation

`tests/test_ec2.py`, 155 passed. Two new cases: the SDK round-trip for both instance fields, parsed with `ipaddress` rather than merely asserted present, and an `AllocateAddress` / `DescribeAddresses` validity check. Both fail on current `main`, the first on the missing field and the second on the malformed one.

## References

- botocore `botocore/data/ec2/2016-11-15/service-2.json`, `Instance` shape

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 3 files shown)


Nahuel990 (approved):

Nahuel990:
Thanks!

Issue #1344, opened 2026-08-11
EC2: launch real containers from a user-supplied image (custom AMI + docker network)


iot-rocket:
The EC2 control plane is broad and works well for metadata: security groups,
key pairs, VPC/subnet objects, and instance records all behave, which is enough
for code that only *manipulates* infrastructure. What we're missing is the other
half - `RunInstances` starting something we can actually talk to.

```bash
aws ec2 run-instances --image-id ami-c0ffee0c0ffee --instance-type t3.micro
# returns a well-formed instance record; no container is started, nothing listens
```

## Use case

We emulate an **SSH jump-host appliance**: a Debian 13 box running systemd and
`sshd`, which devices open reverse tunnels to and which our Lambdas reach over
the same Docker network. Locally this used to be an EC2 instance backed by a
container, so the same code path (`RunInstances` → SSH to the instance IP) ran
in dev and in production. It is a common shape - anything that tests
"infrastructure + a real server on it" needs it.

## What we'd love to see

**1. A custom AMI backed by a local Docker image.** The convention we already
have tooling for (and that other migrating users will too) is *tagging*: an
image tagged `<prefix>/<name>:<ami-id>` is registered as that AMI, so a locally
built image becomes launchable without any registry round-trip.

```bash
docker build -t localstack-ec2/remote-server-debian13:ami-c0ffee0c0ffee .
aws ec2 run-instances --image-id ami-c0ffee0c0ffee ...   # starts THAT image
```

Any equivalent mechanism would work for us - an env var pointing at a catalog
file, or an `ImportImage`-style API. The important part is *user-supplied
images*, since the interesting appliances (systemd + sshd + our own packages)
can't come from a built-in catalog.

**2. systemd guests.** Starting the container with `/sbin/init` as PID 1 plus
the mounts/flags systemd needs (cgroup access, `/run` + `/tmp` tmpfs,
optionally privileged). floci models this nicely with a per-image
`guestRuntime: systemd` flag in its `ec2/image-catalog.yaml` - that shape would
be a good fit here too.

**3. Instances on a user-specified Docker network.** For an instance to be
useful it has to be reachable from the other containers in the developer's
compose project - ideally by container name, and able to reach the gateway at
its in-network alias. A `EC2_DOCKER_NETWORK` (or reusing the existing
`DOCKER_NETWORK`) would cover it.

**4. Already-standard extras**, in rough priority order: `user-data` executed on
[...]

PR #1426, opened 2026-08-17
feat(ec2): opt-in Docker VM manager (EC2_VM_MANAGER=docker)

iot-rocket:
Draft, for #1344. Not asking for a merge, asking whether the shape is one you would take.

## Summary

`RunInstances` returns a well-formed record and boots nothing, which is the right default and stays the default. With `EC2_VM_MANAGER=docker`, an AMI id that resolves to a locally tagged image boots as a container instead, and the instance record carries that container's address.

## Why it is opt-in

You said on #1344 that a real box is a dev-runtime capability rather than an AWS-fidelity gap. That is the reason it hangs off a flag and not a default. With the flag unset the code path is not reached at all: one predicate, `EC2_VM_MANAGER == "docker" and _get_docker() is not None`, read at three places. Flag unset with a Docker daemon present: `tests/test_ec2.py` 170 passed, `RunInstances` created no containers, `DescribeImages` returned the same 3 stub AMIs as before.

The pattern follows `DSQL_STRICT` and the RDS engine containers: default off, and the emulator behaves exactly as it does today.

## What it does when enabled

Tag any local image so the tag is an AMI id:

```
docker tag my-appliance ministack-ec2/my-appliance:ami-c0ffee0c0ffee
```

- `DescribeImages` lists it with `ec2_vm_manager=docker`, so a caller can detect the capability instead of guessing
- `RunInstances` boots `ministack-ec2-{account}-{region}-{instance-id}`, labelled `ministack=ec2`, on the detected network
- the container's IP replaces the synthetic private and public addresses
- stop, start, reboot and terminate map to the container lifecycle; user-data runs on boot; `KeyName` injects root `authorized_keys`
- an AMI id with no local image, or no daemon, stays metadata-only and logs why

Measured on this branch: `RunInstances` booted the container, `DescribeInstances` reported its `172.17.0.2`, `TerminateInstances` removed it.

## Tests

`tests/test_ec2.py`, 173 passed. The unit tier drives a docker-py stand-in and needs no daemon. Three integration cases tag a local alpine as an AMI, boot it, assert the container and its address, and clean up. Their gate probes for a reachable daemon, so they run on your runners rather than skipping.

## What it does not do

A container that dies on its own leaves the instance reported `running`. Reconciliation runs on state restore, not live, because the read path would have to call Docker, and those calls block the event loop today (#1278). It belongs after #1277, and wants a real `InstanceInitiatedShutdownBehavior` in front of it.

Also worth saying: we do not need this merged. Our own jump host runs as a plain container next to the emulator. We carry the branch either way, and if you would rather not own this surface, closing the draft is a fine answer. The reason to send it is that the question on #1344 was what the thing looks like, and this is easier to judge than a description of it.


iot-rocket:
@Nahuel990 Two updates on this draft.

First, the red shard is not this branch. Both failures are
`tests/test_iam.py::test_iam_account_authorization_details_*`, and the server log
artifact has the real error: `GetAccountAuthorizationDetails` answers 500 with
`unhashable type: 'dict'`. The CFN role provisioner stores `ManagedPolicyArns`
entries as `{"PolicyName", "PolicyArn"}` dicts while `AttachRolePolicy` stores plain
ARN strings in the same `AttachedPolicies` list, so the account-wide scan crashes as
soon as any CFN-provisioned role with a managed policy is in the store. This branch
touches neither `iam.py` nor the provisioner (both byte-identical to main), and the
two tests pass against a server built from it when run alone. On this PR's shard
plan they share a worker with the CFN/SAM role tests, and since the shard plan is
derived from the diff, reruns keep the same neighbours and stay red. Filed as
#1461 with a two-command repro. The fix is to normalise the provisioner to the
string the API path stores.

Second, since your concurrency branch may land first, this branch is rebased onto
`poc/concurrency-1277-1377` and the PR's base is switched to it, so the diff stays
this one commit (GitHub retargets to main when the concurrency branch merges). The
rework uses that machinery instead of working around it: instance containers carry
`container_reaper.own_labels("ec2")`, the reset sweep matches `ministack.instance`
instead of every `ministack=ec2` container on the daemon and goes through
`drop_containers`, a `register_live_ids("ec2", ...)` provider vouches for every
container an instance record still owns (a stopped instance keeps its exited
container for `StartInstances`), and the docker-touching actions plus
`CreateKeyPair`'s openssl subprocess run through `run_offloop`. The launched
containers run the AMI's own workload, nothing wired back to 4566, so they cannot
re-enter. That also closes the "What it does not do" paragraph in the description:
a container that dies out-of-band now reports its instance `stopped` on the next
`DescribeInstances` (downgrade only, `StartInstances` is the way back), instead of
the record claiming `running` until a state restore. A daemon error during the
check is treated as no evidence, not as a dead box. `tests/test_ec2.py` is at 183
passed; the new cases pin the ownership labels, the instance-scoped reset against a
foreign instance's decoy container, the exited/dead/gone/daemon-down distinctions,
a partial two-instance boot failure backing out the first container, and an
out-of-band kill in the live tier.

Nahuel990:
@iot-rocket sorry I haven't been ignoring this, I am trying to make the puzzle work with ECS (and potentially EKS). Ideally this will ship with 1.5.0 (along with concurrency, which I am still generating random sets of tests for to make sure we are not breaking anything)

PR #1427, opened 2026-08-17
feat(cloudformation): provision AWS::SES::ConfigurationSet and event destinations

ryan-bennett:
## What

Adds CloudFormation provisioning for two SES resource types that previously failed with `Unsupported resource type`:

- `AWS::SES::ConfigurationSet`
- `AWS::SES::ConfigurationSetEventDestination`

## Why

A stack that wires SES delivery tracking (a configuration set plus its CloudWatch/SNS event destination, exactly what the CDK `ses.ConfigurationSet` construct emits) could not be deployed against MiniStack. Provisioning failed on the first unsupported resource and rolled the stack back.

## How

No custom resource. Both provisioners delegate to the existing in-memory SES stores:

- The configuration set is registered with **both** the classic (v1) and v2 SES stores, so `DescribeConfigurationSet` (Query API) and `GetConfigurationSet` (v2 REST API) both resolve it. `Ref` returns the configuration set name; delete removes it from both stores.
- The event destination is recorded against its configuration set so the set round-trips with its destinations.

This is CloudFormation provisioning fidelity only: emails remain stored in-memory and SES events are not delivered, consistent with the existing SES emulation.

## Tests

- New end-to-end test in `tests/test_cfn.py` deploys a stack with an SNS topic, a configuration set, and an event destination; asserts `CREATE_COMPLETE`, the `Ref` output, visibility through both the v1 and v2 SES APIs, and clean deletion.
- `pytest tests/test_cfn.py` (180) and `pytest tests/test_ses.py tests/test_ses_v2.py` (88) all pass.
- `ruff check ministack/` clean.

## Checklist

- [x] Tests added and passing
- [x] `CHANGELOG.md` updated
- [x] README CloudFormation resource table updated


Nahuel990:
ConfigurationSet doesn't support GetAtt
    return name, {"Id": name}
    https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-ses-configurationset.html

Nahuel990:
I'll remove the GetAtt before release

PR #1428, opened 2026-08-17
rds: opt-in Aurora PostgreSQL readers with real streaming replication

Kiran01bm:
## Summary

Slice 2 of #1325: with `MINISTACK_RDS_PG_CLUSTER_REPLICATION=1`, the second and later members of an Aurora PostgreSQL cluster launch their own PostgreSQL containers running as real hot standbys that stream WAL from the writer. Off by default; Aurora MySQL and flag-off behavior are unchanged.

## Why

The shared-container Aurora model makes every cluster member an alias of one database process, so the `ReaderEndpoint` is read/write and there is no way to locally rehearse the things a reader actually changes: read/write connection routing, "read-only transaction" (`25006`) error handling, replica-lag tolerance, and reader-endpoint pooling. Slice 1 (#1339) carved the ownership/enumeration/endpoint seams; this slice puts a real replicating reader behind them.

## What

- `MINISTACK_RDS_PG_CLUSTER_REPLICATION=1` (documented in the README env table) enables the behavior for `aurora-postgresql` only; requires `DOCKER_NETWORK`.
- The first cluster member remains the writer on the existing shared container; each later member gets a dedicated container cloned via `pg_basebackup --write-recovery-conf` and started as a hot standby (via `gosu`, which the official `postgres:*-alpine` images ship).
- A reader is genuinely read-only: `pg_is_in_recovery()` is true, writes fail with SQLSTATE `25006`, and writer data becomes visible on the reader.
- `ReaderEndpoint` resolves to an available standby, falling back to the writer/shared container when none is available.
- Reader bootstrap failure marks the instance failed and cleans up its partial container/volume; deleting a reader removes only reader-owned compute.
- Not yet modeled (follow-up slices of #1325): `StopDBCluster` with replicating standbys is rejected with `InvalidDBClusterStateFault`, and a persisted standby demotes to a shared-container alias on warm boot instead of restoring a dead endpoint.
- Without `DOCKER_NETWORK`, the flag logs a warning and falls back to the coherent shared-alias model rather than fabricating a fake reader.

```text
Before (flag off — unchanged default)          After (MINISTACK_RDS_PG_CLUSTER_REPLICATION=1)

  writer ──┐                                     writer ─────────── pg (primary)
  reader ──┼── one shared pg container           reader-1 ── pg hot standby ◀─ WAL ─┐
  reader ──┘   (ReaderEndpoint = writer,         reader-2 ── pg hot standby ◀─ WAL ─┤
               read/write)                                                          │
                                                 ReaderEndpoint → available standby ┘
                                                 (read-only: 25006 on writes)
```

Tests: fake-Docker coverage for the flag gate, reader launch/lifecycle/deletion, endpoint resolution, stop rejection, and restore-time demotion; plus a live Docker test (gated on `DOCKER_NETWORK` + the flag) proving distinct containers, recovery mode, WAL streaming, read-only rejection, and reader-endpoint routing.


Nahuel990:
there is an mqtt test failing (flaky condition on lambda cold start, please ignore it in case it fails... I will be fixing it during the day)

Nahuel990 (approved):

PR #1429, opened 2026-08-17
fix(cfn): preserve the uniqueness suffix when truncating auto-generated physical names

ryan-bennett:
## Bug

`_physical_name()` builds the full `"{stack}-{logicalId}-{suffix}"` string and only then slices it to `max_len` from the end. For a deeply-nested CDK stack, the nested stack's own generated name (parent name + nested-stack logical id + a CloudFormation-assigned suffix) can already exceed a downstream resource's own name-length limit on its own — e.g. Lambda's 64-char `FunctionName` cap.

When that happens, every resource in that nested stack truncates to an identical string regardless of `logical_id`, since the part that would have disambiguated them (logical_id, then the hash) never survives the slice.

Concretely: every Lambda function inside such a nested stack collapses onto one physical function, so only one of them ever actually runs regardless of which the caller invokes — surfaces as every route on an `HttpApi` returning the same handler's response, no matter the path or method.

Repro: a CDK app with a `NestedStack` containing an `HttpApi` with several Lambda integrations (typical for a construct like `aws-cdk-lib/aws-apigatewayv2` route fan-out inside a nested stack). Once the nested stack's own auto-generated name is long enough (easy to hit — CDK's own nested-stack naming convention is `{ParentId}NestedStack{ParentId}NestedStackResource{hash}`, already 40+ characters before ministack appends its own `{parent}-{that}-{uuid}` wrapper), every Lambda in it collapses onto one physical function.

## Fix

Reserve room for the suffix and truncate the `"{stack}-{logicalId}-"` prefix instead, so the one part that actually guarantees uniqueness (a hash of `stack_name` *and* `logical_id`) always survives:

```diff
     suffix = hashlib.sha256(f"{stack_name}:{logical_id}".encode()).hexdigest()[:13].upper()
-    base = f"{stack_name}-{logical_id}-{suffix}"
+    prefix = f"{stack_name}-{logical_id}-"
+    available = max(max_len - len(suffix), 0)
+    base = prefix[:available] + suffix
     if lowercase:
         base = base.lower()
     return base[:max_len]
```

No change in shape for the common case (stack_name + logical_id already under `max_len`).

## Test plan

- Added `test_cfn_nested_stack_long_name_lambda_functions_get_distinct_physical_names` to `tests/test_cfn.py`: two real Lambda functions inside a nested stack whose own generated name deliberately exceeds 64 characters, asserting they get distinct physical names/ARNs and are independently invocable. Confirmed this test fails against the unfixed code (`first_arn != second_arn` trips — they were identical) and passes against the fix.
- Ran `tests/test_cfn.py` (179), `tests/test_apigatewayv2.py` (127), and `tests/test_lambda.py` (360) — 667 passed, 5 pre-existing unrelated skips, no regressions.


Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
Awesome thanks!

Nahuel990 (approved):

Nahuel990 (approved):

Issue #1322, opened 2026-08-08
S3: ceph/s3-tests conformance run — 353/961 passing, 11 reproducible defects


gaul:
Ran the full [ceph/s3-tests](https://github.com/ceph/s3-tests) suite (@ `5522d1c`) against
MiniStack 1.4.14 (`1590804`) to get a baseline on S3 wire conformance.

**353 of 961 executed tests pass** (1053 collected, 92 the suite skips itself). Restricted to the
core S3 data plane — `test_s3.py` minus the subsystems MiniStack does not implement at all — it is
**259/491 (52.7%)**.

Most of the gap is expected: SSE, bucket logging, lifecycle processing, bucket-policy evaluation
and most of IAM aren't implemented, and per `SECURITY.md` request signatures are deliberately not
verified (that stance alone accounts for 107 failures asserting that a bad request is rejected).
Those are not what this issue is about.

Underneath them sit **11 defects in code that *is* implemented**. Each was reproduced directly
against a stock server, outside the test suite. I'm happy to split these into one issue apiece if
that's preferred — filing together first so the conformance baseline lives in one place.

## Setup

```
ministack 1.4.14, PERSIST_STATE=0, path-style, SigV4, boto3 1.43.67, py3.11
main/alt/tenant given 12-digit access keys (000000000000 / 111111111111 / 222222222222)
so each identity lands in its own MiniStack account.
```

The suite was also re-run with every client sharing the server's default `test` secret, in case the
credential choice skewed the result — it moved the total from 353 to 350.

---

## 1. POST Object rejects any form whose fields carry a `filename`

`_post_object` classifies a multipart part as the object body when
`name == "file" or filename is not None` (`ministack/services/s3.py:2564`). Browsers and HTTP
libraries routinely set `filename=` on ordinary form fields — Python `requests` does it for every
field passed via `files=`. Every field then looks like the body, no `key` field survives, and the
upload is rejected.

```python
import boto3, requests
from collections import OrderedDict
c = boto3.client("s3", endpoint_url="http://localhost:4566",
                 aws_access_key_id="test", aws_secret_access_key="test")
c.create_bucket(Bucket="b")

# every part carries filename= -> 400
r = requests.post("http://localhost:4566/b",
                  files=OrderedDict([("key", "a.txt"), ("file", ("bar"))]))
print(r.status_code, r.text)
# 400 <Error><Code>InvalidArgument</Code>
#     <Message>Bucket POST must contain a field named 'key'.</Message>

[...]

Issue #1058, opened 2026-07-09
VersionId Missing from S3 Object Metadata When Payload Persisted to Volume


adzcodemi:
### Affected Service

S3

### What happened?

When Ministack is configured with a persistent volume for S3 payloads (.WithVolume(name: "ministack-s3-payloads", target: "/tmp/ministack-data/s3")), the versionId field is missing from S3 object metadata after the data is stored on or retrieved from the volume.

### Reproduction steps

```python
1. Start the MiniStack container.
2. Upload an object to the S3 bucket.
3. Compare the raw object metadata between StackPort and the Docker container volume. Confirm that the versionId exists in the StackPort metadata but is missing from the metadata stored in the Docker container volume.
```

### Expected behavior

The versionId should be present in the S3 object metadata response, consistent with behavior when no volume is used.

### MiniStack version

latest

Nahuel990:
@adzcodemi this should be fixed in v1.4.1

PR #1434, opened 2026-08-17
fix(s3): honor canned ACLs on buckets and key object ACLs by version

gaul:
SDKs send `put-bucket-acl --acl public-read` as an x-amz-acl header with an empty body, and PutBucketAcl read only the body, so the canned grant vanished and the bucket read back owner-only.  A caller that gates anonymous access on the reported grants -- S3Proxy's aws-s3 backend among them -- then treated every bucket as private.  CreateBucket dropped the header the same way.  Both now validate the canned value and store the grants it implies through the object-side table (#1322), log-delivery-write included; an XML body must be well-formed (MalformedACLError) and a request with neither header nor body answers MissingSecurityHeader, as on AWS.

Object ACLs are now keyed per version like tags (#1058): ?versionId= reads or writes that version's ACL, a fresh version starts from the default policy instead of inheriting its predecessor's, and the canned value on PutObject sticks to the version that PUT created.  Version and bucket deletes drop the stored entries.  The browser-POST form names its field `acl` rather than `x-amz-acl`, so it never matched the synthetic header list and was discarded; it now applies to the uploaded object.

This is thirteen fewer failures on S3Proxy's s3-tests lane: the twelve ACL-cluster tests pass, and test_expected_bucket_owner returns to being an honest expected failure instead of passing by accident off the dropped grant.


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1439, opened 2026-08-17
fix(s3): state, validate, and enforce server-side encryption

gaul:
SSE-S3, SSE-KMS and SSE-C headers all returned 200 and were then forgotten: nothing was echoed on the PUT reply or on later reads, an object written under a customer key read back without one, and incoherent requests -- a wrong key MD5, key material without the algorithm, SSE-C combined with KMS -- were accepted.

MiniStack still does not encrypt at rest.  Like the ACL plane, SSE is contract state now: validated on write (InvalidArgument on a bad algorithm, a non-256-bit key, a missing or mismatched key MD5, mutually exclusive modes together, or aws:kms naming no key -- there is no implicit account aws/s3 key to fall back to), persisted with the object so HEAD and GET echo it, and enforced on reads for SSE-C, whose whole point is that the object is unreadable without the key: keyless, wrong-key, and key-offered-for-a-plain-object reads are 400s, per ceph/s3-tests.  The state rides preserved_headers, so it follows versions, copies and multipart completes through the existing plumbing, and the customer key itself is never stored -- only its MD5, which is all AWS keeps visible either.

A bucket's default encryption now applies to the object rather than only stamping the PUT reply.  Multipart uploads echo their encryption on initiate, require the initiating SSE-C key on every part and on UploadPartCopy, and require it once more on CompleteMultipartUpload -- as AWS does and ceph/s3-tests pins -- completing into an object that carries it.  Copies never inherit the source's encryption: the destination's SSE comes from the copy request or the destination bucket's default, an SSE-C source demands its key as x-amz-copy-source-... parameters, and every response echoes what applied.  The POST form's SSE fields work like the headers.

This fixes 102 tests on S3Proxy's s3-tests lane (16 of them jointly with an S3Proxy-side fix: its post-copy metadata read now presents the destination's customer key).  The four ?encryption-subresource tests expecting the pre-2023 ServerSideEncryptionConfigurationNotFoundError stay red deliberately: MiniStack models the modern always-on default, as LocalStack does, which those tests are already marked for on that backend.


Nahuel990:
@gaul can you share the snippet you use to test this? I'm getting a 403 Access Denied in S3 (eu-west-1) for the wrong MD5 I'll review the tests tomorrow

gaul:
Sure — snippet below.  But the 403 you hit is almost certainly not the MD5 check: since **April 2026, S3 disables SSE-C by default** on new general-purpose buckets (and existing buckets in accounts with no SSE-C objects).  On a blocked bucket every write that carries SSE-C headers answers `403 AccessDenied` before the key material is looked at — you'll get the identical 403 with the *correct* MD5.  `get-bucket-encryption` will show `"BlockedEncryptionTypes"`, and you can unblock with:

```
aws s3api put-bucket-encryption --bucket $BUCKET \
  --server-side-encryption-configuration '{"Rules":[{
    "ApplyServerSideEncryptionByDefault": {"SSEAlgorithm": "AES256"},
    "BlockedEncryptionTypes": {"EncryptionType": ["NONE"]}}]}'
```

(Docs: https://docs.aws.amazon.com/AmazonS3/latest/userguide/blocking-unblocking-s3-c-encryption-gpb.html)

Here's what I test with; it runs identically against MiniStack and against S3 once SSE-C is unblocked.  One boto3 trap: pass `SSECustomerKeyMD5` explicitly — if you omit it, boto3 computes it from the key, so a wrong MD5 never reaches the wire.

```python
import boto3

KEY_A = 'pO3upElrwuEXSoFwCfnZPdSsmt/xWeFa0N9KgDijwVs='   # base64 of 32 random bytes
MD5_A = 'DWygnHRtgiJ77HCm+1rvHw=='                       # base64 MD5 of those raw bytes
KEY_B = '6b+WOZ1T3cqZMxgThRcXAQBrS5mXKdDUphvpxptl9/4='
MD5_B = 'arxBvwY2V4SiOne6yppVPQ=='
A = dict(SSECustomerAlgorithm='AES256', SSECustomerKey=KEY_A, SSECustomerKeyMD5=MD5_A)
B = dict(SSECustomerAlgorithm='AES256', SSECustomerKey=KEY_B, SSECustomerKeyMD5=MD5_B)

s3 = boto3.client('s3', endpoint_url='http://127.0.0.1:4566')  # drop endpoint_url for AWS
BUCKET = 'sse-c-probe-CHANGEME'
s3.create_bucket(Bucket=BUCKET)

def show(label, fn):
    try:
        print(label, '->', fn()['ResponseMetadata']['HTTPStatusCode'])
    except Exception as e:
        err = e.response['Error']
        print(label, '->', e.response['ResponseMetadata']['HTTPStatusCode'],
              err['Code'], '--', err['Message'])

show('PUT wrong MD5', lambda: s3.put_object(Bucket=BUCKET, Key='o', Body=b'x',
     SSECustomerAlgorithm='AES256', SSECustomerKey=KEY_A, SSECustomerKeyMD5=MD5_B))
show('PUT good     ', lambda: s3.put_object(Bucket=BUCKET, Key='o', Body=b'x', **A))
show('GET keyless  ', lambda: s3.get_object(Bucket=BUCKET, Key='o'))
show('GET wrong key', lambda: s3.get_object(Bucket=BUCKET, Key='o', **B))
show('GET right key', lambda: s3.get_object(Bucket=BUCKET, Key='o', **A))
```

Expected, and what this branch now answers:

```
PUT wrong MD5 -> 400 InvalidArgument -- The calculated MD5 hash of the key did not match the hash that was provided.
PUT good      -> 200
GET keyless   -> 400 InvalidRequest -- The object was stored using a form of Server Side Encryption. ...
GET wrong key -> 403 AccessDenied -- Requests specifying Server Side Encryption with Customer provided keys must provide the correct secret key.
GET right key -> 200
```

And your probe caught something real regardless: I had the wrong-**key** read answering 400 because ceph/s3-tests asserts 400 there — that turns out to be radosgw's answer, not S3's.  Amended and force-pushed: a read presenting the wrong key is `403 AccessDenied` now, while parts of an SSE-C multipart upload stay 400 with AWS's part-specific messages, which is how AWS splits it.  Beyond the PR's pytest cases, the branch runs ceph/s3-tests through S3Proxy against MiniStack — the encryption family plus both copy-encryption matrices pass, with the two tests that pin radosgw's 400 marked as expected divergences.

One question your experiment raises for a follow-up: whether MiniStack should model the new default block itself (SSE-C writes 403 until `PutBucketEncryption` unblocks).  LocalStack doesn't yet, and it would fail the whole s3-tests SSE-C family out of the box, so this PR leaves SSE-C available by default — happy to file an issue if you want the 2026 behavior.

Nahuel990 (approved):

Nahuel990:
Thanks!

Issue #1432, opened 2026-08-17
DynamoDB: key attribute type not enforced on PutItem, Query key conditions, or UpdateTable


iot-rocket:
## Affected Service

dynamodb

## What happened?

If a key attribute is declared `S`, real DynamoDB refuses any item, key or key condition that uses
`N` for it. MiniStack refuses this on most paths but not on all of them, and the two paths that
are missing are exactly the ones you use in normal application code.

Measured, table with `device_name` (HASH, S) and `timestamp` (RANGE, S):

| Operation | MiniStack 1.4.19 | Real AWS |
| --- | --- | --- |
| `PutItem`, item has `timestamp` as N | **accepted, stored as N** | `ValidationException: One or more parameter values were invalid: Type mismatch for key timestamp expected: S actual: N` |
| `Query`, key condition value is N | **accepted, returns the row** | `ValidationException: One or more parameter values were invalid: Condition parameter type does not match schema type` |
| `UpdateItem`, key has N | rejected, correct | `ValidationException: The provided key element does not match the schema` |
| `BatchWriteItem`, key has N | rejected, correct | same |
| `TransactWriteItems`, item has N | rejected, correct, with the exact AWS wording | same |
| `GetItem`, key has N | rejected, correct | same |
| `UpdateTable` with only `--attribute-definitions` | **accepted, key type changed in place** | `ValidationException: At least one of ProvisionedThroughput, BillingMode, ... is required`, and no API can change a key attribute type |

So the validator is present and already wired into `UpdateItem`, `BatchWriteItem`,
`TransactWriteItems` and `GetItem`. `TransactWriteItems` even produces the exact AWS message
("Type mismatch for key timestamp expected: S actual: N"), so the correct string exists in the
code base. Only `PutItem` and the `Query` key condition path do not call it.

There is a second effect that follows from the first. Once a row is stored with the wrong type,
key comparison ignores the type:

* `GetItem` with `{"timestamp":{"N":"1782028800"}}` is rejected as not matching the schema
* `GetItem` with `{"timestamp":{"S":"1782028800"}}` returns that same row, whose stored value is
  `{"N":"1782028800"}`

So `GetItem` refuses the key the item actually has, and accepts a key the item does not have.

The direction does not matter. With `ts` declared `N`, `PutItem` with `{"ts":{"S":"123"}}` is
accepted in the same way.

What it costs a user: the table quietly fills with items real DynamoDB would never accept, and the
[...]

Issue #1416, opened 2026-08-16
CloudFormation: AWS::SSM::Parameter writes the store directly, so it overwrites existing parameters


iot-rocket:
## Affected Service

CloudFormation

## What happened?

Not the same thing as #1415, even though both say SSM and CloudFormation: that one is
about template parameters of type `AWS::SSM::Parameter::Value<...>`, this one is about
the `AWS::SSM::Parameter` resource. Both came out of the same first deploy into a real
account, together with #1414.

The CloudFormation provisioner for `AWS::SSM::Parameter` writes into the SSM service's
parameter dict directly (`cloudformation/provisioners.py:1020`) instead of going through
`put_parameter`. The SSM API path itself is fine, so the two doors into the same store
behave differently.

What follows from it:

- A stack can create a parameter that already exists. Real CloudFormation fails the
  create, MiniStack overwrites it without a word. The API path has the right error
  already (`ParameterAlreadyExists`, `services/ssm.py:271-280`).
- `Version` stays 1 on every overwrite, while the API path increments it (`ssm.py:282`).
  A parameter written by CloudFormation keeps version 1 forever, so anything watching the
  version for changes never sees one.
- Nothing is written to the history, so `GetParameterHistory` still returns a single
  entry after several stack updates.
- A `SecureString` ends up in plain text. `put_parameter` stores the encrypted form plus
  `OriginalValue`, and `_param_out` (`ssm.py:672-676`) hands out the plain value only for
  `WithDecryption=true`. A parameter created by CloudFormation has no `OriginalValue` and
  a plain `Value`, so `GetParameter` without `WithDecryption` returns the secret.

For us this was useful at first, because it is what lets us pre-seed a placeholder
parameter locally that a stack then creates over. On real AWS that does not work. So it
is no blocker on our side, but the SecureString read looked worth reporting.

## Reproduction steps

On a plain `ministackorg/ministack:1.4.18`:

```bash
aws ssm put-parameter --name /probe/manual --type String --value made-by-hand

cat > t3.yaml <<'YAML'
Resources:
  Param:
    Type: AWS::SSM::Parameter
    Properties:
      Name: /probe/manual
      Type: String
      Value: made-by-cloudformation
  Secret:
    Type: AWS::SSM::Parameter
    Properties:
      Name: /probe/secret
      Type: SecureString
      Value: hunter2
YAML
aws cloudformation create-stack --stack-name probe-collide --template-body file://t3.yaml
```

Result:

```
[...]

Issue #1415, opened 2026-08-16
CloudFormation: AWS::SSM::Parameter::Value<...> is not re-resolved on UpdateStack


iot-rocket:
## Affected Service

CloudFormation

## What happened?

Follow-up to section 4 of #1345. The create path is fixed since v1.4.16 and we confirmed
that in the thread: a template parameter of type `AWS::SSM::Parameter::Value<String>`
now resolves against the SSM store when the stack is created. On a stack update the old
value is kept, even when the SSM parameter has a new value in the meantime.

What we see in the code: `_resolve_parameters` takes the previous deployment's map and,
for a parameter sent with `UsePreviousValue=true`, uses the stored value and sets
`already_resolved = True` (`cloudformation/engine.py:169-174`), which then skips the SSM
lookup at `:189`. The stored value is the one from the last deployment, because
UpdateStack writes the resolved value back as `ParameterValue`
(`cloudformation/handlers.py:528`). So the SSM parameter name is gone after the first
create, and there is nothing left to re-resolve.

This matters for us because `cdk deploy` sends `UsePreviousValue=true` for every
parameter it does not override, which is the normal case. Our concrete symptom: two
stacks publish their API id to SSM and a third one reads both. On a fresh environment
the consumer deploys against a placeholder value, and no later deploy of that stack ever
picks up the real ids, so we repair the Lambda environment variables with a post-deploy
script.

## Reproduction steps

On a plain `ministackorg/ministack:1.4.18`:

```bash
aws ssm put-parameter --name /probe/api-id --type String --value FIRST

cat > t1.yaml <<'YAML'
Parameters:
  ApiId:
    Type: AWS::SSM::Parameter::Value<String>
    Default: /probe/api-id
Resources:
  Bucket:
    Type: AWS::S3::Bucket
    Properties:
      BucketName: probe-resolve-bucket
Outputs:
  SeenApiId:
    Value: !Ref ApiId
YAML
sed 's/SeenApiId/SeenApiId2/' t1.yaml > t2.yaml   # any template change, so the update runs

aws cloudformation create-stack --stack-name probe-resolve --template-body file://t1.yaml
aws ssm put-parameter --name /probe/api-id --type String --value SECOND --overwrite
aws cloudformation update-stack --stack-name probe-resolve --template-body file://t2.yaml \
  --parameters ParameterKey=ApiId,UsePreviousValue=true
```

Result:

```
[...]

Issue #1411, opened 2026-08-16
IAM: CreateRole accepts Description values that real AWS rejects (charset not validated)


iot-rocket:
## Affected Service

iam

## What happened?

We have been deploying our CDK stacks against ministack locally and in CI for
months. This week the same stacks went to a real AWS account for the first
time, and one of them failed on CreateRole:

```
Resource handler returned message: "1 validation error detected: Value at
'description' failed to satisfy constraint: Member must satisfy regular
expression pattern: [\u0009\u000A\u000D\u0020-\u007E\u00A1-\u00FF]* (Service: Iam, Status Code: 400)"
```

The role description contained an em dash (U+2014). Real IAM restricts the
description field to tab/CR/LF plus printable ASCII and the Latin-1
supplement. MiniStack stores the string verbatim, so the invalid template
deployed fine locally the whole time and the error only showed up in the real
account.

As far as I can see there is no validation on this path at all: in
`ministack/services/iam.py`, `create_role` and `update_role_description` both
store `_p(p, "Description")` as-is, and the file has no charset check
anywhere (checked on current main).

## Reproduction steps

```
$ aws --endpoint-url http://localhost:4566 iam create-role \
    --role-name repro \
    --description 'Test role — with an em dash' \
    --assume-role-policy-document file://trust.json
# ministack: succeeds, GetRole returns the description verbatim
```

The same call against real AWS:

```
An error occurred (ValidationError) when calling the CreateRole operation:
1 validation error detected: Value at 'description' failed to satisfy
constraint: Member must satisfy regular expression pattern: [\u0009\u000A\u000D\u0020-\u007E\u00A1-\u00FF]*
```

## Expected behavior

CreateRole and UpdateRoleDescription reject a description with characters
outside `[\u0009\u000A\u000D\u0020-\u007E\u00A1-\u00FF]` with the same 400 ValidationError as real AWS.

Related, in case you want it in the same fix: RoleName and Path have their own
documented patterns which are not checked either. I kept this report scoped to
the description because that is the one that bit us. We can send a small PR
for this if you want.

## MiniStack version

v1.4.17 (fork build with some of our open PRs on top; `services/iam.py` on
current main shows the same behavior)

## SDK / tool used

AWS CDK 2.1128.0 / CloudFormation on the real-AWS side, AWS CLI v2 for the
isolated repro.

Nahuel990:
[...]

PR #1444, opened 2026-08-18
fix(cfn): SSM fix + iam charset

Nahuel990:


iot-rocket:
@Nahuel990 
We ran a server built from this branch against the exact repros we filed with
the issues, rather than reviewing by eye, and re-checked everything against
main after the merge. All four check out:

- **#1411**: `CreateRole` with an em dash in `Description` now answers
  `ValidationError` with the constraint message. Confirmed.
- **#1416**: a template `AWS::SSM::Parameter` over an existing name now fails
  the stack and leaves the store value untouched, a fresh parameter starts at
  `Version: 1`, and a stack update bumps it to `Version: 2` through the real
  `put_parameter` path. Confirmed.
- **#1415**: `AWS::SSM::Parameter::Value<...>` with `UsePreviousValue` now
  re-resolves on `UpdateStack` - we flipped the underlying parameter between
  create and update and the dependent resource picked up the new value.
  Confirmed.
- **#1432**: `PutItem` with the wrong type for a key attribute and a `Query`
  key condition with a mismatched type both answer `ValidationException`.
  That part landed via #1445 - confirmed there too, and thanks for taking
  that one as well.

One thing we noticed while validating: the branch also restored most of the
SSE wiring that went missing from #1439's merge (it went in via #1445). On
main before that, the SSE functions all existed but `PutObject` neither
stored nor echoed the resolved headers, which is why seven of #1439's own
tests were failing on the tip; with #1445 in, all ten SSE tests pass on main
again. Four call sites from the original are still unwired, all on paths the
tests do not pin: `UploadPartCopy` skips both SSE-C checks (the part key and
the copy-source key), and the `UploadPartCopy` / `CompleteMultipartUpload`
responses do not echo the stored algorithm. We have those restored on a
small follow-up branch rebased onto current main, with two new tests that
pin them (both fail on main today): opening it as a PR right after posting
this.

PR #1449, opened 2026-08-19
fix(s3): restore the SSE-C checks and echoes still missing after #1444

iot-rocket:
Follow-up promised in the #1444 thread. The #1439 squash lost six SSE call
sites in its merge resolution; the #1445 half of the #1444 branch restored
the PutObject store+echo, the POST-form fields and the complete-time SSE-C
key check. Four call sites stayed unwired, all on paths no test pinned:

- UploadPartCopy accepted a part for an SSE-C upload without the create-time
  key, and read an SSE-C copy source without the source key (the check also
  covers the versionId-qualified source path, which did not exist when #1439
  was written).
- The UploadPartCopy response did not echo SSECustomerAlgorithm /
  SSECustomerKeyMD5, though the wire shape models both.
- The CompleteMultipartUpload response did not echo the stored
  ServerSideEncryption.

All four now behave as #1439 originally wrote them, and two new tests pin
them; both fail on current main. tests/test_s3.py: 254 passed. One CHANGELOG
entry under Unreleased/Fixed.


Nahuel990:
@iot-rocket I just checked that PR and this is new behaviour, correct but new... nothing got lost in any merge

Nahuel990:
I haven't checked multi-part SSE-C but this makes sense, I will release 1.4.20, merge this after and release it during Aug19th/20th

iot-rocket:
@Nahuel990 Fine either way, the label changes nothing for this PR. For the record
though: the four call sites are in #1439's first commit, gaul's 2f3d5db, in
_upload_part_copy and _complete_multipart_upload:

https://github.com/ministackorg/ministack/blob/2f3d5db2c890de97e0d83ed4ad5e7ec282d88ed9/ministack/services/s3.py#L4947
https://github.com/ministackorg/ministack/blob/2f3d5db2c890de97e0d83ed4ad5e7ec282d88ed9/ministack/services/s3.py#L5133

The second commit on that branch (7cdc2ab, the merge of main) removed them.
The main-side parent had no SSE code at all, so the lines came from the PR
side and went missing there too. And the Files changed tab is misleading
here: it renders final head against base, where you do see a
_check_sse_c_part(headers, upload) line, but that one sits in _upload_part.
Same text, different function. The copies in _upload_part_copy are the ones
that vanished, and a head-vs-base view cannot show a line that is in
neither.

The outcome is the same either way: this PR changes only those four paths,
the two new tests pin them, and everything else stays byte-for-byte what
main does today.

Nahuel990:
@iot-rocket yes sorry for that I f*** it in the merge conflict, took the incoming instead for some reason github still displaying it, thanks for catching... I'm merging it now

image

iot-rocket:
It happens even to the best of us ;)

PR #1451, opened 2026-08-19
fix(s3): answer a conditional delete of an absent key 204

gaul:
DeleteObject read a missing key as a failed precondition and answered 412, where S3 counts deleting a key that is not there as done whatever condition the request carries -- so a caller retrying a delete it had already made got an error for work that was complete.

If-Match: * was wrong the other way: compared against the ETag as a literal it failed every object, where it asks only that some object be there.


Nahuel990 (approved):

Nahuel990:
@gaul I will merge this one but I will change the code to 412 before release. Please ensure the changes are accurate. If you cannot validate them, feel free to open an Issue not a PR with the snippet that triggers your error and I can apply the fix

PR #1452, opened 2026-08-19
fix(s3): honour the size and last-modified-time delete conditions

gaul:
DeleteObject took x-amz-if-match-size and
x-amz-if-match-last-modified-time and DeleteObjects took the Size and LastModifiedTime fields, and all four were read off the wire and dropped -- so a caller guarding a delete against a concurrent write had the object removed anyway, which is the one outcome the condition exists to prevent.

They now narrow the delete the way an ETag does: a mismatch is 412 on the single delete and a PreconditionFailed entry under Errors in a batch, an unparseable value never holds, and a key that is not there holds every condition.  A request naming a version conditions on that version rather than on whatever is current.

### DIFF
--- a/ministack/services/s3.py
+++ b/ministack/services/s3.py
@@ -790,6 +790,67 @@ def _check_read_preconditions(headers: dict, obj: dict, resp_headers: dict):
     return None
 
 
+def _delete_condition_target(bucket: dict, bucket_name: str, key: str,
+                             version_id: str) -> dict | None:
+    """The object record a delete's conditions are evaluated against.
+
+    Without a version id that is the current object; with one it is the
+    version addressed, since that is what the delete would remove.  A delete
+    marker, or a version that is not there, reads as nothing to condition on.
+    """
+    if not version_id:
+        return bucket["objects"].get(key)
+    for v in _object_versions.get((bucket_name, key), []):
+        if v["version_id"] == version_id:
+            if v.get("is_delete_marker"):
+                return None
+            return _object_record_from_version(v)
+    current = bucket["objects"].get(key)
+    if version_id == "null" and current is not None and not current.get(
+            "version_id"):
+        return current
+    return None
+
+
+def _delete_conditions_hold(obj: dict | None, size: str | None,
+                            last_modified: str | None) -> bool:
+    """Whether a delete's size and last-modified-time conditions hold.
+
+    These ride alongside If-Match on a conditional delete and narrow it the
+    same way: the object goes only if it is still the one the caller last
+    read.  A key that is not there holds every condition, since deleting one
+    that is already gone is a success whatever the request asked for.
+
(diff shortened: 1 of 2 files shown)


gaul:
Closing this — I measured it against S3 in `eu-west-1` and the premise doesn't hold.

This PR was written the same way as #1451 and #1458: reasoned about rather than measured. @Nahuel990 caught those two in #1458, and checking this one the same way says it's wrong too.

The stated hazard was that the four conditions are "read off the wire and dropped -- so a caller guarding a delete against a concurrent write had the object removed anyway." A caller can't reach that state on real S3, because S3 refuses the request before any delete happens.

**`DeleteObject`** with `x-amz-if-match-size` / `x-amz-if-match-last-modified-time` on a general purpose bucket (`scripts/probe_delete_size_mtime.py`):

```
case                          status  error            object survived
if-match-size, wrong (99)     501     NotImplemented   yes
if-match-size, right (5)      501     NotImplemented   yes
if-match-mtime, wrong (2001)  501     NotImplemented   yes
if-match-mtime, right         501     NotImplemented   yes
right etag + wrong size       501     NotImplemented   yes
```

**`DeleteObjects`** with `Size` / `LastModifiedTime` in the XML body, same bucket:

```
case                           outcome                                     survived
Size wrong (99)                400 MalformedXML                            yes
Size right (5)                 400 MalformedXML                            yes
LastModifiedTime wrong (2001)  400 MalformedXML                            yes
right ETag + wrong Size        400 MalformedXML                            yes
right ETag alone               200 deleted=1 errors=[]                     no
wrong ETag alone               200 deleted=0 errors=['PreconditionFailed'] yes
```

Two things worth pulling out of the second table:

- `ETag` in the batch body already works on general purpose buckets and ministack already honours it. Only `Size` and `LastModifiedTime` are refused, with `MalformedXML` — the general purpose schema has no such elements.
- `right ETag + wrong Size` is a 400, so a valid `ETag` does not carry an invalid sibling field through.

The API reference says the same once read carefully. Both headers are marked "only supported for directory buckets", and `DeleteObjects` has *separate* conditional-delete examples per bucket type: the directory bucket one carries `<LastModifiedTime>` and `<Size>`, the two general purpose ones use `<ETag>` alone.

There's nothing to narrow this to — excluding the ETag path that already works leaves nothing correct — so closing rather than reworking.

If ministack wants to model this faithfully the change is the opposite of this PR: answer the two headers 501 `NotImplemented` and the two XML elements 400 `MalformedXML`, so directory-bucket code written against ministack fails the way it will against S3. Happy to raise that separately if it's wanted.

Both probes are on `s3-conditional-delete-404` in my fork alongside the #1458 replacement, so the evidence is reproducible: they print status and error code per case rather than asserting, and run against S3 or ministack alike.

PR #1463, opened 2026-08-20
Ship MySQL IAM plugins in slim image

Areson:
## Why

The MySQL IAM auth plugin artifacts are compiled in `Dockerfile.full` and ship only in the `-full` image flavor; the slim image has no `/opt/ministack/mysql-plugins` tree, so IAM auth support silently disables itself there. An earlier revision of this PR compiled the plugins in the slim Dockerfile as well, which adds several minutes of compile time to every branch build. Per review feedback, this revision copies the artifact tree from the published full image instead — the plugins are release-stable, so recompiling them per branch buys nothing.

## What

- `Dockerfile` (slim): adds a global `PLUGIN_DONOR_IMAGE` build arg (default `ghcr.io/ministackorg/ministack:full`), a `plugin-donor` stage, and a final-stage `COPY --from=plugin-donor /opt/ministack/mysql-plugins /opt/ministack/mysql-plugins`. No compile stages are added.
- `.github/workflows/docker-publish.yml`: the full image now builds and pushes before the regular one, and the regular build pins `PLUGIN_DONOR_IMAGE` to the just-pushed full image digest. Release-tagged slim and full images therefore always carry the same release's plugin tree (the digest is a manifest list, so each architecture copies its own tree).
- `.github/workflows/docker-publish-on-pr.yml`: unchanged. PR previews use the build-arg default — the latest published full image — so ordinary branch builds skip plugin compilation entirely.
- `contrib/mysql-iam-plugin/README.md`: flavor claims updated to match.

## Behavior changes / risk

- Runtime behavior is unchanged: plugin delivery keys solely on artifact presence at the fixed lookup path, which the slim image now satisfies.
- Slim PR previews carry the latest release's plugin binaries rather than binaries built from the PR's source; PRs that change plugin source are validated via the `full-preview` label image, which compiles from source.
- The full-before-regular ordering and the digest handoff only run in the release workflow; the first tag push after merge exercises them for real.

## Validation

* `docker buildx build --check .` — clean
* `docker build --target plugin-donor .` — donor stage resolves and builds
* Copied plugin tree verified byte-identical (per-file sha256) to the tree carried by the published full image, for both artifact series

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -60,20 +60,6 @@ jobs:
             type=semver,pattern={{major}}.{{minor}}
             type=raw,value=latest
 
-      - name: Build and push (regular)
-        uses: docker/build-push-action@v5
-        with:
-          context: .
-          file: Dockerfile
-          platforms: linux/amd64,linux/arm64
-          push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
-          build-args: |
-            MINISTACK_VERSION=${{ steps.meta.outputs.version }}
-          cache-from: type=gha,scope=regular
-          cache-to: type=gha,mode=max,scope=regular
-
       # ── Full (Debian/glibc) image — :full, :{version}-full, :{major}.{minor}-full ──
       # Adds DuckDB (Athena engine), psycopg2-binary, pymysql. Larger
       # image (~360 MB vs ~110 MB) but all manylinux wheels install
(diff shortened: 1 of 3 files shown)


Nahuel990:
@Areson this will add roughly 8 minutes to the build time per branch. What we can do instead is copy the plugin from the Dockerfull once it’s generated. If no Dockerfull is generated (usual PRs), it will copy the latest cached artifact.

Just a heads-up: we also trimmed the Dockerfull by around 100 MB in the last release

Nahuel990 (approved):

PR #1464, opened 2026-08-21
fix(dynamodb): resolve UpdateExpression aliases before the key-attribute check

ppettitau:
## Summary

An `UpdateExpression` that targets a non-key attribute through an `ExpressionAttributeNames` alias was rejected as though it targeted the table's partition key.

Against a table whose partition key is named `pk`, this fails with `ValidationException: One or more parameter values were invalid: Cannot update attribute pk. This attribute is part of the key`, even though the update targets `gsi-1-pk`:

```js
new UpdateItemCommand({
  TableName: "my-table",
  Key: marshall({ pk: "item-1", sk: "detail" }),
  ReturnValues: "UPDATED_NEW",
  UpdateExpression: "set #pk = :pk",
  ExpressionAttributeNames: { "#pk": "gsi-1-pk" },
  ExpressionAttributeValues: marshall({ ":pk": "ARCHIVED#2026-01-01" }),
})
```

Writing a GSI partition key attribute whose alias is named after the table key is a common pattern in single-table designs, so this blocks a realistic workload. Real DynamoDB accepts the call.

## Root cause

`_update_item` pre-rejected updates that target a key attribute before the expression was evaluated. That check matched the key name as raw text against the expression:

```python
re.search(rf"(?i)\b(SET|ADD|REMOVE|DELETE)\b[^,]*?\b{re.escape(_kn)}\b", update_expr)
```

`#` and `:` are non-word characters, so `\bpk\b` matches inside the alias `#pk` and inside the value reference `:pk`. A second branch attempted alias resolution, but the raw-text branch fired first. The defect was confined to this pre-check; the downstream evaluator resolves aliases correctly via `_parse_path_from_tokens`.

## Fix

The pre-check is removed rather than replaced. `_apply_update_expression` already returns `updated_attrs`, the alias-resolved root of every path it touches, which is exactly the set the check needs. The key names are now tested against that set, so the expression is never scanned a second time.

This is cheaper than the code it replaces, which compiled and ran a regex once per key name plus once per alias.

- `_key_attribute_update_error(table, updated_attrs)` holds the check, with the error code, message text and 400 status unchanged.
- `updated_attrs` is owned by `_update_item` and passed into `_apply_update_expression`, so targets resolved before a raise are still available. AWS reports the key rejection ahead of the runtime operand-type errors the evaluator raises, and `ADD pk :v` hits both.
- `_apply_add` and `_apply_delete` record the path root immediately after resolving it, rather than after their operand-type checks, for the same ordering reason.

Net effect on `ministack/services/dynamodb.py` is 25 insertions and 26 deletions.

No other occurrence of the raw-text pattern exists in the file. `TransactWriteItems` validates through `_apply_update_expression`, so it needs no change.

## Tests

New `key_guard_table` fixture (partition key `part`, sort key `sort`, non-key attribute `other-part`) and 7 tests in `tests/test_dynamodb.py`:

- an alias named after the partition key resolving to a non-key attribute succeeds and leaves the keys untouched
- a value reference named after a key is not treated as a target
- direct `SET part` and `SET sort` are still rejected
- an alias genuinely resolving to the partition key is still rejected
- `REMOVE` and `ADD` against a key attribute are still rejected

The two positive tests fail against unmodified code with `Cannot update attribute sort. This attribute is part of the key`; the five rejection tests pass before and after, confirming no regression in the intended behaviour. The `ADD` rejection test also pins the error ordering described above.

## Verification

Run against a container built from this branch.

- `pytest tests/test_dynamodb.py` — 322 passed
- `pytest tests/ -n 4 --dist=loadfile -m "not serial"` — 6144 passed, 40 failed
- `pytest tests/ -m serial` — 78 passed, 9 failed
- `ruff check ministack/` — clean

No DynamoDB failures. The full-suite failures are host-reachability cases in my local environment (lambda proxy, EventBridge and Cognito lambda containers, OIDC callbacks, ALB dataplane, Athena) and reproduce identically on a build with this change stashed.

## Checklist

Adapted from `CONTRIBUTING.md`. This is a bug fix to an existing service, so the new-service steps do not apply.

- [x] Tests added and passing
- [ ] ~~New service file in `ministack/services/`~~ — n/a
- [ ] ~~Registered in `SERVICE_REGISTRY`~~ — n/a
- [ ] ~~Detection patterns in `router.py`~~ — n/a
- [ ] ~~Service added to the README table~~ — n/a


Nahuel990 (approved):

Nahuel990:
Thanks!

PR #1468, opened 2026-08-21
fix(rds): honor DescribeDBInstances filters

jayjanssen:
## Summary

- parse the AWS Query filter shape emitted by boto3/AWS CLI (`Filters.Filter.N.Name` and `Filters.Filter.N.Values.Value.N`)
- preserve MiniStack's JSON-flattened `Filters.member.N` compatibility path
- add an SDK-level regression test with instances in two Aurora clusters, covering one-value selection and multi-value OR semantics

## Root cause

`_apply_instance_filters` already implements `db-cluster-id`, but `_parse_filters` only recognized MiniStack's internal `Filters.member.N` representation. AWS Query SDKs serialize the same request as `Filters.Filter.N` / `Values.Value.N`, so parsing returned no filters and `DescribeDBInstances` returned every instance.

Reproduced against the released `ministackorg/ministack:1.4.20-full` image (source commit `c4ba0e1a630b75b576b7e233300d3dba758abf6a`) with two clusters: filtering for the first returned both instances before this change.

## Validation

- `pytest -q tests/test_rds.py::test_rds_describe_instances_v2 tests/test_rds.py::test_rds_describe_instances_filters_by_cluster tests/test_rds.py::test_rds_flatten_json_request_params` — 3 passed
- `ruff check ministack/` — passed
- direct parser check for both AWS Query and internal member forms — passed


jayjanssen:
Could a maintainer add the `full-preview` label? The downstream reproducer currently pins the `-full` image for its RDS IAM plugin artifacts, so the full preview would let it validate this fix before the next release. I attempted to add the label but external contributors do not have label permissions.

Nahuel990 (approved):

Issue #1461, opened 2026-08-20
IAM: a CFN-provisioned role with ManagedPolicyArns breaks GetAccountAuthorizationDetails (500, unhashable dict)


iot-rocket:
## Affected Service

IAM

## What happened?

`AttachedPolicies` on a role record holds two different shapes, depending on who wrote it:

- `AttachRolePolicy` (`services/iam.py`, `_attach_role_policy`) appends the plain ARN
  string;
- the CloudFormation `AWS::IAM::Role` provisioner
  (`services/cloudformation/provisioners.py`, "ManagedPolicyArns") appends
  `{"PolicyName": ..., "PolicyArn": ...}` dicts.

`GetAccountAuthorizationDetails` iterates every role's `AttachedPolicies`, assumes
strings, and calls `_lookup_policy(arn)`. A dict ends up in `_policies.get(<dict>)`,
which raises `TypeError: unhashable type: 'dict'`. The request answers 500 and
botocore reports it as "invalid XML received". One CFN-provisioned role with a
managed policy anywhere in the account is enough, and the call stays broken until
that role is gone.

The mismatch reaches further than that one call:

- `ListAttachedRolePolicies` walks the same list through `_lookup_policy` and
  `arn.rsplit(...)`, so it fails the same way on a CFN-provisioned role.
- `ListEntitiesForPolicy` compares with `if arn in r.get("AttachedPolicies", [])`,
  which never matches a dict entry, so a CFN attachment is invisible there.
- Adjacent, not part of the shape problem: the provisioner also skips the
  `AttachmentCount` bookkeeping the API path does, so `DeletePolicy` does not see a
  CFN attachment either.

No consumer reads the dict shape as a dict. The provisioner writes it and nothing
expects it, which is what makes the fix safe.

We hit this on a PR's CI shard where the two `test_iam_account_authorization_details_*`
tests shared a worker with CFN/SAM role templates. The tests pass in isolation, which
makes the failure look like a flake. It is not.

## Reproduction steps

```bash
docker run --rm -d -p 4566:4566 --name ms ministackorg/ministack:1.4.21-full
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1
A="aws --endpoint-url http://localhost:4566"

cat > role.json <<'JSON'
{"Resources":{"R":{"Type":"AWS::IAM::Role","Properties":{
 "RoleName":"managed-arn-probe-role",
[...]

PR #1472, opened 2026-08-22
feat(ec2): opt-in Docker VM manager (EC2_VM_MANAGER=docker)

iot-rocket:
Re-files #1426. That one closed itself when `poc/concurrency-1277-1377` was deleted
after #1386 merged: with the base branch gone, GitHub closed the PR instead of
retargeting it, and nothing was merged from it. This is the same commit rebased onto
`e4b2266`, and `git range-diff` against the head you reviewed reports no change. The
review history is on #1426. Part of #1344.

`RunInstances` returns a well-formed `running` record without booting anything, so a
workload that needs a reachable box gets a phantom instance. With
`EC2_VM_MANAGER=docker`, an AMI id that resolves to a locally tagged image
(`<EC2_DOCKER_IMAGE_PREFIX>/<name>:<ami-id>`) starts as a container on
`DOCKER_NETWORK`, and the container IP becomes the instance address.
`StopInstances`, `StartInstances`, `RebootInstances` and `TerminateInstances` map to
the container lifecycle, user data runs on boot, and `KeyName` writes root
`authorized_keys`. `ImportKeyPair` now stores the material and `CreateKeyPair`
generates keys through the `openssl` CLI, so this adds no dependency. Unknown AMI ids
and a missing Docker daemon keep the metadata-only behaviour, so the default path is
unchanged. A failed boot fails the call and backs the records out, which follows
`LAMBDA_STRICT` rather than the degrade-silently norm.

The rework uses the concurrency machinery from #1386 instead of working around it.
Instance containers carry `container_reaper.own_labels("ec2")`. The reset sweep
matches `ministack.instance` rather than every `ministack=ec2` container on the
daemon and goes through `drop_containers`. A `register_live_ids("ec2", ...)` provider
vouches for every container an instance record still owns, so a stopped instance
keeps its exited container for `StartInstances`. The Docker calls and the `openssl`
subprocess run through `run_offloop`. The containers run the AMI's own workload and
nothing points back at 4566, so they cannot re-enter. One behavioural consequence: a
container that dies out of band now reports its instance `stopped` on the next
`DescribeInstances`. That is a downgrade only, `StartInstances` is the way back, and
a daemon error during the check counts as no evidence rather than as a dead box.

`tests/test_ec2.py` is at 184 passed. The unit tier drives a docker-py stand-in, so
it runs without a daemon; the integration cases tag a local alpine as an AMI, boot
it, assert the container and its address, and clean up after themselves. Their gate
probes for a reachable daemon instead of reading `DOCKER_NETWORK`, so they run
wherever Docker exists. The new cases cover the ownership labels, the
instance-scoped reset against a decoy container belonging to a foreign instance, the
exited/dead/gone/daemon-down distinctions, a partial two-instance boot failure that
backs out the first container, and an out-of-band kill in the live tier.

Four things are deliberately not covered: the stop-failure path, the reboot IP
refresh, the belt-and-braces `_cleanup_terminated`, and the off-loop dispatch gate.

This stays a draft. As on #1426, the question is whether the shape is one you want,
not whether it should merge today. You mentioned 1.5.0 together with concurrency,
which suits us.

One note on CI: the red shard on #1426 did not come from this branch. Both failures
were `tests/test_iam.py::test_iam_account_authorization_details_*`, where
`GetAccountAuthorizationDetails` answers 500 with `unhashable type: 'dict'`, because
the CFN role provisioner stores `ManagedPolicyArns` entries as dicts while
`AttachRolePolicy` stores plain ARN strings in the same `AttachedPolicies` list. This
branch touches neither `iam.py` nor the provisioner; the shard plan simply puts those
two tests next to the CFN/SAM role tests. Reported in #1461.


Nahuel990:
@iot-rocket I had the chance to review this with a few external engineers from AWS. The way to make it work with parity for everyone, is to let you register your own container as an AMI.

So from now on you publish your image wherever (ghrc, docker, ecs, etc) whatever your daemon can pull, and then register it:

    aws ec2 register-image --name my-box --image-location ghcr.io/iot-rocket/my-box:1.0

That returns an ami- id, and from there you use it exactly the way you would use any EC2 image. DescribeImages lists it, RunInstances boots it, the instance reports the container's address, TerminateInstances removes it. Nothing is configured on our side

That also settles the cases you raised. If you need a different image per workload, register more than one. If you need --privileged, an sshd endpoint, systemd, or any particular config, you build that into your own image and we treat it the same as any other. The shapes you need are yours to make, which is the right boundary: on AWS the contents of the AMI are from the customer, not the control plaen

We already do the same thing for Lambda where people who want PHP or Bun attach their own container image and run it on MiniStack-provided infrastructure (BYO-Lambda)

A container has no persistent disk, so these register as "instance store" backed, which means StopInstances and StartInstances are refused exactly as AWS refuses them. And we inject nothing into the guest, so keys and startup belong in your image

iot-rocket:
@Nahuel990 
You were faster than our answer. The shape is right, and it
removes both the tag convention and the injection our PR carried. Instance-store
semantics fit us as well, since we call neither `StopInstances` nor `StartInstances`.

Two findings from testing `fa4b1f1`.

1. Local images work, and the README does not say so. Registering a tag that exists
only on the daemon and was never pushed returns an id and launches with no registry
involved, because `containers.run` pulls only when the image is missing. "Any image
your Docker or Podman daemon can pull" reads as if a registry were required. One
sentence would help: the build, run, change one line, run again loop stays local, and
so does CI without registry credentials.

2. A systemd guest cannot boot, because the launch options are fixed. Here is a
repro that does not need our image. The shim is there because systemd insists on
being PID 1, while the launch adds docker-init in front and a keepalive command
behind:

```bash
mkdir /tmp/systemd-box && cd /tmp/systemd-box
cat > init-shim <<'EOF'
#!/bin/sh
[ "$$" -eq 1 ] && exec /sbin/init
exec unshare --pid --mount-proc --fork --kill-child /sbin/init
EOF
cat > Dockerfile <<'EOF'
FROM debian:13-slim
RUN apt-get update && apt-get install -y --no-install-recommends systemd systemd-sysv util-linux \
 && rm -rf /var/lib/apt/lists/* \
 && systemctl mask getty.target console-getty.service systemd-logind.service systemd-udevd.service
COPY init-shim /usr/local/sbin/init-shim
RUN chmod +x /usr/local/sbin/init-shim
STOPSIGNAL SIGRTMIN+3
ENTRYPOINT ["/usr/local/sbin/init-shim"]
EOF
docker build -t systemd-box:1 .

AMI=$(aws --endpoint-url=http://localhost:4566 ec2 register-image \
        --name systemd-box --image-location systemd-box:1 --query ImageId --output text)
IID=$(aws --endpoint-url=http://localhost:4566 ec2 run-instances --image-id "$AMI" \
        --count 1 --instance-type t3.micro --query 'Instances[0].InstanceId' --output text)
sleep 5
aws --endpoint-url=http://localhost:4566 ec2 describe-instances --instance-ids "$IID" \
  --query 'Reservations[].Instances[].State.Name' --output text     # terminated
docker logs "$(docker ps -a --format '{{.Names}}' | grep "$IID")"   # see below
```

The log line is

```
unshare: unshare failed: Operation not permitted
```

and `terminated` is the right answer for a container that died. The cause is the
options, not the image. Same image, same keepalive command, one flag added:

```bash
docker run -d --init --privileged --name control systemd-box:1 sleep infinity
docker top control     # /sbin/init and systemd-journald are up
```

I looked for something weaker and did not find it. `--cap-add SYS_ADMIN` gets past
`unshare` and then fails on `Failed to set RLIMIT_CORE: Operation not permitted`, and
adding `SYS_RESOURCE`, `--ulimit core=-1`, `seccomp=unconfined`, `apparmor=unconfined`,
a writable cgroup mount, the `--cgroupns` variants, or dropping the appended init so
the entrypoint is PID 1 did not produce a booting systemd either.

So the question is whether you want a knob for the launch options: a field on
`RegisterImage`, an environment variable, or something per image. The default stays as
it is, and it would apply only to an image the user registered, which is the same
trust boundary as the image contents. Without it an instance can run a container but
not an operating system, which leaves out jump hosts and anything running services
under systemd.

Nahuel990:
@iot-rocket yes, sorry for that a fix with a listener to ENTRYPOINT will land soon and should fix that case

Nahuel990:
@iot-rocket from v1.5.0 you can use   ENTRYPOINT ["/usr/sbin/sshd", "-D", "-e"] (or reequired flags) and it should work

iot-rocket:
@Nahuel990 Measured on `1.5.0-full` before answering. The ENTRYPOINT handling works
as you describe: our image declares its own ENTRYPOINT, the launch no longer
appends `sleep infinity`, and a variant of the image with
`ENTRYPOINT ["/bin/sh","-c","ssh-keygen -A; exec /usr/sbin/sshd -D -e"]` comes up
`running` with sshd answering on port 22 at the instance address. A container that
runs one service as PID 1 needs nothing else now. Thanks, that part is done.

The systemd case is unchanged, because it never was about the command. Re-run on
1.5.0 with the `systemd-box:1` image from my earlier comment (debian:13-slim +
systemd + the init shim):

```bash
AMI=$(aws --endpoint-url=http://localhost:4566 ec2 register-image \
        --name systemd-box --image-location systemd-box:1 --query ImageId --output text)
IID=$(aws --endpoint-url=http://localhost:4566 ec2 run-instances --image-id "$AMI" \
        --count 1 --instance-type t3.micro --query 'Instances[0].InstanceId' --output text)
sleep 5
aws --endpoint-url=http://localhost:4566 ec2 describe-instances --instance-ids "$IID" \
  --query 'Reservations[].Instances[].State.Name' --output text   # terminated
docker logs "$(docker ps -a --format '{{.Names}}' | grep "$IID")"
# unshare: unshare failed: Operation not permitted

docker run -d --init --privileged --name control systemd-box:1   # control: boots,
docker top control                                               # systemd is up
```

The container is unprivileged and systemd cannot set up its namespaces there; the
plain-docker control with `--privileged` boots the same image fine. Running the
services directly as ENTRYPOINT would sidestep it, but our image exists to run the
production units the way the real box runs them (sshd, rsyslog, two systemd
timers), and that needs systemd as init.

So the remaining ask from last time stands: an opt-in way to pass launch options
for a registered image. We carry it as one commit on v1.5.0 meanwhile: an
`EC2_DOCKER_FLAGS` environment variable taking a docker-CLI-style string
(`--privileged`, `--cap-add`, `-e`, `-v`, `--tmpfs`, `--add-host`, `-m`,
`--shm-size`), parsed into `containers.run` kwargs, `--init` refused, unset
changes nothing. With `EC2_DOCKER_FLAGS=--privileged` the same register/run
sequence boots our systemd box: `running`, sshd on 22, `docker top` shows systemd.
Parser plus one `kwargs.update` line plus two tests; say the word and it comes as
a PR. A per-image field on `RegisterImage` would be fine by us too, if you prefer
the knob there.

Closing this PR either way: #1473 plus the 1.5.0 ENTRYPOINT handling covers
everything else it carried.

iot-rocket:
@Nahuel990 I said I would wait for a word, but a concrete diff is easier to judge
than a description, so it is filed as #1489: one commit, the parser and its tests,
default untouched. Your CI's preview image for it already boots our systemd box
with `EC2_DOCKER_FLAGS=--privileged` (checked against
`ministack-preview-build:pr-1489-397a568d`). If you would rather not carry the
knob, closing it costs nothing and we keep it on the fork.

PR #1483, opened 2026-08-24
fix(s3): answer the versioning edges the way S3 answers them

gaul:
Four places where a versioned bucket answered something other than what AWS says, all of them found by s3proxy's conformance tests:

- DeleteBucket emptied a bucket that still held versions.  Deleting an object under versioning writes a delete marker rather than removing anything, so a bucket whose listing is empty may still hold every version ever written to it; AWS refuses it BucketNotEmpty until they go by version id, and dropping them silently loses data the caller could otherwise still read.

- Reading a delete marker by its version id returned 200 with an empty body on GET, and NoSuchVersion on HEAD.  The marker is there and has no content, which AWS says with 405 and the marker's own headers; neither of the old answers told the caller which of the two it was.

- An ACL or tag op naming a version that never existed, or that has since been deleted, read back the default policy as though the version were merely unpermissioned.  A version id addresses a version the way a key addresses an object: what is not there answers NoSuchVersion.  All five subresource ops ask the one helper.

- ListObjectVersions ignored the delimiter and listed every key flat, so a client walking a versioned bucket by folder saw no groups at all.  It now rolls up exactly as ListObjects does.


Nahuel990:
@gaul You are inferrinf these from AI. To avoid the back-and-forth that happened in 1451, please first raise an issue with the repro steps showing where Ministack drifted from AWS. Once that’s identified, we can fix it directly. I'm closing these ones

PR #1491, opened 2026-08-24
fix(rds): support PostgreSQL major engine selectors

jayjanssen:
## Summary

- treat a major-only Aurora PostgreSQL `EngineVersion` such as `16` as a dot-boundary selector for every advertised minor in that major
- honor `DefaultOnly` and keep major defaults explicit, matching AWS's current `16` default of `16.11` rather than deriving it from the newest catalog entry
- preserve exact-version selection

## Root cause

`DescribeDBEngineVersions` compared `EngineVersion` to each catalog version with exact string equality. The catalog contains concrete versions such as `16.8` and `16.14`, so `EngineVersion=16` returned zero records.

Against AWS in `us-east-1`, `Engine=aurora-postgresql, EngineVersion=16` currently returns 13 available `aurora-postgresql16` records. Adding `DefaultOnly=true` returns exactly `16.11`. MiniStack v1.5.0 returned zero for both calls, while exact `16.14` already worked.

## Validation

- regression reproduced against `ministackorg/ministack:1.5.0-full`
- `pytest -q tests/test_rds.py::test_rds_describe_engine_versions_family tests/test_rds.py::test_rds_describe_aurora_mysql_engine_versions_by_family tests/test_rds.py::test_rds_describe_aurora_postgresql_engine_versions_by_family tests/test_rds.py::test_rds_describe_aurora_postgresql_engine_versions_by_major tests/test_rds.py::test_rds_describe_aurora_postgresql_engine_version_status` — 5 passed
- regression covers 13-version major listing, unrelated-major exclusion, `16` + `DefaultOnly` → `16.11`, exact `16.14`, family metadata, and `Status=available`
- `ruff check ministack/ tests/test_rds.py` — passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versioning follows [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- **RDS — Aurora PostgreSQL major-version selectors return the matching catalog** — `DescribeDBEngineVersions` treated `EngineVersion=16` as an exact version and returned nothing. Major-only selectors now return every advertised minor in that family, and `DefaultOnly=true` narrows the result to AWS's configured default minor for that major.
+
 ## [1.5.0] — 2026-08-23
 
 ### Added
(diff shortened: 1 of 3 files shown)


jayjanssen:
Could a maintainer add the `full-preview` label? The downstream consumer pins `1.5.0-full`, so a full preview would allow end-to-end validation against the same runtime flavor before the next release.

Nahuel990 (approved):

PR #1510, opened 2026-08-25
fix(stepfunctions): make CreateStateMachine idempotent

bandle:
`CreateStateMachine` throws `StateMachineAlreadyExists` on any second create of a name, where real Step Functions treats an identical call as idempotent and answers with the machine it already made. 

Our `terraform apply` bringing in the CI got flaky due to this issue. The resource exists and terraform has not recorded it, so the failed apply cannot simply be re-run either.

### Fix

A second create returns the existing ARN and its original `creationDate` when it describes the same machine. As documented but also measured against real AWS

- The check is on the machine: `definition`, `type`, `loggingConfiguration`, and whether the call publishes. A re-create that publishes answers with the version the first one made, and publishes no second version
- A differing `roleArn` or set of tags is **ignored**: it neither fails the call nor overwrites what is stored
- `StateMachineAlreadyExists` is left for the case it is meant for, a name reused for a different machine, which is the half that already worked

### Testing

`test_sfn_create_duplicate_name` asserted the old behaviour, so it is rewritten rather than joined by a second test: an identical re-create returns the same ARN and creation date, a differing `roleArn` is accepted without updating the stored one, and a different definition still raises.

The script below runs against MiniStack or real AWS:

```
                                          before                        this branch
an identical re-create returns the ARN    StateMachineAlreadyExists     same ARN
the answer keeps the first creation date  None                          2026-08-25 08:25:32
a differing roleArn is ignored, no write  StateMachineAlreadyExists     same ARN, role unchanged
a different definition still raises       StateMachineAlreadyExists     StateMachineAlreadyExists
```

<details>
<summary>check_sfn_create_idempotency.py</summary>

```python
#!/usr/bin/env python3
"""Check that CreateStateMachine is idempotent, the way Step Functions documents it.

AWS treats a second create of a name as idempotent when it describes the same machine -- name,
definition, type, logging configuration, and whether the call publishes: it returns the EXISTING
ARN and creation date rather than an error, and it ignores a differing roleArn or tags (without
updating them). StateMachineAlreadyExists is reserved for a name reused for a different machine.

That contract is what makes the call safe for an SDK to retry, and every SDK assumes it is: a
create that times out or 500s in transit is sent again, and under MiniStack the second attempt
fails the caller even though the first one succeeded. A `terraform apply` deploying several
machines dies there, with the resource created but absent from state, so the retry cannot be
resumed either.

Against MiniStack:
  python check_sfn_create_idempotency.py --endpoint http://localhost:4566 --region eu-central-1

  python check_sfn_create_idempotency.py --profile <profile> --region eu-central-1

Creates and deletes one state machine and, on AWS only, one role with no permissions. Nothing is
executed: the definition is a single Pass state. Pass --keep to leave them behind. Exit status is 0
when every check passes.
"""

import argparse
import json
import sys
import time
import uuid
from contextlib import suppress

import boto3

DUMMY_ROLE = "arn:aws:iam::000000000000:role/sfn-idempotency-check"
OTHER_ROLE = "arn:aws:iam::000000000000:role/sfn-idempotency-check-other"
TRUST_POLICY = {
    "Version": "2012-10-17",
    "Statement": [{
        "Effect": "Allow",
        "Principal": {"Service": "states.amazonaws.com"},
        "Action": "sts:AssumeRole",
    }],
}
DEFINITION = json.dumps({"StartAt": "Ok", "States": {"Ok": {"Type": "Pass", "End": True}}})
OTHER_DEFINITION = json.dumps({"StartAt": "Other", "States": {"Other": {"Type": "Pass", "End": True}}})


def options():
    parser.add_argument("--keep", action="store_true", help="Leave resources behind")
    return parser.parse_args()


def create(sfn, name, role_arn, definition):
    """(arn, creation_date, error_code). Exactly one of arn / error_code is set."""
    try:
        made = sfn.create_state_machine(name=name, definition=definition, roleArn=role_arn)
    except ClientError as exc:
        return None, None, exc.response["Error"]["Code"]
    return made["stateMachineArn"], made.get("creationDate"), ""


def main():
    args = options()
                                aws_secret_access_key="test")
    else:
    client_args = {"endpoint_url": args.endpoint} if args.endpoint else {}
    sfn = session.client("stepfunctions", **client_args)
    iam = session.client("iam", **client_args)

    name = f"sfn-idempotency-{uuid.uuid4().hex[:8]}"
    created = []
    results = []

    def check(label, ok, detail):
        print(f"{'PASS' if ok else 'FAIL'}  {label:48} {detail}")

    try:
        if args.endpoint:
            role_arn, other_role_arn = DUMMY_ROLE, OTHER_ROLE
        else:
            roles = []
            for suffix in ("role", "other"):
                role = iam.create_role(RoleName=f"{name}-{suffix}",
                                       AssumeRolePolicyDocument=json.dumps(TRUST_POLICY))["Role"]
                created.append(("role", role["RoleName"]))
                roles.append(role["Arn"])
            role_arn, other_role_arn = roles
            time.sleep(15)  # IAM propagation before Step Functions accepts the role

        first_arn, first_date, error = create(sfn, name, role_arn, DEFINITION)
        if not first_arn:
            check("the machine can be created at all", False, f"{error}")
            return 1
        created.append(("state_machine", first_arn))
        check("the machine can be created at all", True, first_arn.rsplit(":", 1)[-1])

        # 1. The retry case, and the only one that matters to a caller: byte-identical re-create.
        again_arn, again_date, error = create(sfn, name, role_arn, DEFINITION)
        check("an identical re-create returns the existing ARN", again_arn == first_arn,
              "same ARN" if again_arn == first_arn else f"raised {error}")

        # 2. Idempotent means returning the original, not making a second one: the creation date
        #    has to be the first one's. A build that re-created under the same name passes check 1
        #    and fails this.
        ok = again_date is not None and again_date == first_date
        check("the idempotent answer keeps the first creation date", ok,
              f"{first_date}" if ok else f"{first_date} -> {again_date}")

        # 3. AWS ignores a differing roleArn: it neither errors nor writes the new role over the
        #    stored one. Both halves are one property, and scoring them as one check keeps the
        #    denominator the same on a build that fails here as on one that passes.
        role_arn_result, _, error = create(sfn, name, other_role_arn, DEFINITION)
        live = (sfn.describe_state_machine(stateMachineArn=first_arn)["roleArn"]
                if role_arn_result == first_arn else None)
        ok = role_arn_result == first_arn and live == role_arn
        check("a differing roleArn is ignored and does not overwrite", ok,
              "same ARN, role unchanged" if ok
              else (f"raised {error}" if role_arn_result != first_arn else f"role became {live}"))

        # 4. The error is not gone, only reserved: a name reused with a DIFFERENT definition is
        #    still StateMachineAlreadyExists. A build that made every create succeed fails here.
        clash_arn, _, error = create(sfn, name, role_arn, OTHER_DEFINITION)
        check("a different definition still raises AlreadyExists",
              error == "StateMachineAlreadyExists",
              error if error else f"returned {clash_arn}")
    finally:
        if not args.keep:
            for kind, ident in reversed(created):
                with suppress(ClientError):
                    if kind == "state_machine":
                        sfn.delete_state_machine(stateMachineArn=ident)
                    elif kind == "role":
                        iam.delete_role(RoleName=ident)

    return 1 if failed else 0


if __name__ == "__main__":
    try:
        sys.exit(main())
    except ClientError as exc:
        sys.exit(2)
```

</details>


Nahuel990:
@bandle thanks for this... I checked against aws in eu-west-1 and DescribeStateMachine returns the stored record as-is, so the new createVersionArn bookkeeping key leaks into the response as a field real AWS doesnt have so I'm removing it before release

Nahuel990 (approved):

bandle:
@Nahuel990 oh sorry i didnt check this response thoroughly. thanks for fixing the unintended side-effects.

PR #1513, opened 2026-08-25
fix(app): inflate gzip-compressed request bodies

Lukasdoe:
AWS SDKs gzip request bodies above ~10 KB for operations carrying smithy's @requestCompression trait (Lambda CreateFunction, CloudWatch Logs, EC2) and send Content-Encoding: gzip. The aws-chunked decoder strips the chunk framing but leaves the gzip layer, so handlers parse compressed bytes and fail with SerializationException (Lambda, Logs) or InvalidAction (EC2, where the gzipped form body hides the Action parameter).

Inflate the body once the target service is known and drop the gzip token from Content-Encoding. S3 is excluded: Content-Encoding is object metadata there, so the stored bytes must stay exactly as sent.


Nahuel990:
@Lukasdoe Do you have a repro script for this? I don't have any source to back this PR

PR #1521, opened 2026-08-25
fix(pipes): deliver DynamoDB stream records to a Step Functions target

facuparedes:
## Problem

The Pipes poller delivered to SNS and to nothing else. A pipe whose target is a state machine reported `CurrentState: RUNNING`, advanced no stream position and moved no records, with no error and no log line:

```python
pipes.create_pipe(
    Name="orders-to-workflow",
    Source="arn:aws:dynamodb:us-east-1:000000000000:table/orders/stream/2024-01-01T00:00:00.000",
    Target="arn:aws:states:us-east-1:000000000000:stateMachine:process-order",
    RoleArn=role_arn,
)
# writes to `orders` land on the stream; `process-order` never starts
```

`_poll_once` skipped any pipe whose target ARN was not SNS, so any DynamoDB stream wired to a state machine was dropped on the floor:

```python
if _arn_service(source_arn) != "dynamodb" or _arn_service(target_arn) != "sns":
    continue
```

Separately, the position advanced on attempt rather than on delivery:

```python
for rec in batch:
    _publish_record_to_sns(target_arn, pipe, rec)
_positions[pipe["Arn"]] = pos + len(batch)
```

A batch that failed to publish was skipped permanently and in silence.

## Change

`_poll_once` keeps the loop and the account/region scoping. Everything below it moves into `_poll_pipe`, and delivery into `_deliver_batch`, which returns whether the whole batch reached its target.

A `states` target gets one `StartExecution` per batch, carrying the records as the JSON array `$states.input` reads. That is the shape Pipes delivers on AWS: the target receives the batch, not one event per record.

The stream position advances only once `_deliver_batch` returns true. A `StartExecution` that fails and a target this poller cannot serve both leave the records on the stream, so the next poll retries them until they age out of the retention window, and both log a warning naming the pipe.

`_start_execution` answers the `(status, headers, body)` triple every MiniStack handler answers, so the check is `status >= 400`.

## Testing

Three tests in `tests/test_pipes.py`, each driving `_pipes._poll_once()` in process so no timing is involved:

- `test_pipes_dynamodb_stream_starts_state_machine_with_the_whole_batch`: two writes to the table produce one `StartExecution` whose input is the two records as a JSON array, and the position advances by two.
- `test_pipes_holds_the_batch_when_the_state_machine_does_not_exist`: `StartExecution` answers 400, the position stays where it was, the warning names the pipe, and the next poll offers the same records again.
- `test_pipes_target_this_poller_cannot_deliver_to_says_so`: a target that is neither SNS nor Step Functions holds the batch and logs which service it is.

All three fail against `main`'s poller. Restoring only that file and rerunning:

```
$ git checkout main -- ministack/services/pipes.py && pytest tests/test_pipes.py -q
FAILED tests/test_pipes.py::test_pipes_dynamodb_stream_starts_state_machine_with_the_whole_batch
FAILED tests/test_pipes.py::test_pipes_holds_the_batch_when_the_state_machine_does_not_exist
FAILED tests/test_pipes.py::test_pipes_target_this_poller_cannot_deliver_to_says_so
3 failed, 26 passed in 0.08s
```

With the branch's poller back, and then the whole suite:

```
$ pytest tests/test_pipes.py -q
29 passed in 0.15s

$ pytest tests/ -q
2 failed, 6739 passed, 94 skipped in 816.13s
```

Both failures are environmental and neither touches `pipes.py`. `test_dsql.py::TestContainersE2E::test_env_flag_spins_up_real_backend` reproduces identically on `main` at 97b0a80: `dsql.py` hands a new cluster a port an earlier proxy still holds. `test_rds.py::test_rds_start_db_instance` is container timing under full-suite load, and `pytest tests/test_rds.py` on this same branch gives 277 passed, 14 skipped.

## References

- [AWS: Amazon EventBridge Pipes targets](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-event-target.html), which lists Step Functions state machines
- [AWS: Pipes batching and concurrency](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-batching-concurrency.html), for the batch-shaped payload
- [AWS: DynamoDB streams as a Pipes source](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-dynamodb.html)


Nahuel990 (approved):

Nahuel990 (approved):

Nahuel990:
Thanks!

Issue #1519, opened 2026-08-25
Step Functions JSONata: remaining features


facuparedes:
### Affected Service

Step Functions

### What happened?

AWS documents Step Functions as implementing the JSONata **2.0.6** specification, with a single
exclusion: *"All built-in JSONata functions and operators in the 2.0.6 specification are supported,
with one exception: `$eval` is not available — use `$parse` instead."*

I scored MiniStack 1.5.0 against the official JSONata conformance suite (the 1674 cases jsonata-js
ships, minus the `function-eval` group, which AWS does not implement either):

```
532/1674 cases pass (31%)
1/101 groups fully pass
```

The interesting part is *what* fails. Splitting the 1142 failures by whether the expression calls a
function the evaluator does not have, or uses syntax it cannot parse at all:

```
708  (61%)  syntax the evaluator cannot parse
434  (38%)  a function that is not registered
```

Sixty-one percent of the gap is not about functions. These all fail today, and none of them would be
fixed by registering another `$name`:

| Expression | Error |
|---|---|
| `$states.input.rows[status = "SETTLED"]` | `Unsupported JSONata path segment: rows[status = "SETTLED"]` |
| `$lookup($states.input.obj, "a").b` | `Unsupported JSONata expression: $lookup($states.input.obj, "a").b` |
| `$states.input.arr.({"v": $})` | `Unsupported JSONata path segment: ({"v": $})` |
| `$states.input.d.'a#b'` | `Unsupported JSONata path segment: 'a#b'` |
| `$states.input.a.b.%.c` | `Unsupported JSONata expression: .c` |

Filter predicates, navigating a function's result, object constructors over a sequence, quoted keys
and the parent operator. `$states.input.arr[0]` works, so array indexing is there; everything else
that goes inside `[ ]` or after a `)` is not.

This is the same wall #636, #645 and #669 kept hitting from the other side. #669 added 25 functions
and closed in a day, and the ceiling did not move, because the ceiling is structural: the evaluator
splits the expression string by operator precedence with regular expressions. There is no tokenizer
and no AST, so a filter predicate, a path continuation or a sequence has nowhere to attach. Every
new function makes the 38% smaller and leaves the 61% exactly where it was.

For completeness, the function side is still open too. 21 of the 2.0.6 builtins raise
`States.QueryEvaluationError`:

`$assert` `$decodeUrl` `$decodeUrlComponent` `$each` `$encodeUrl` `$encodeUrlComponent` `$filter`
[...]

Issue #669, opened 2026-05-19
More JSONata Expressions


youngkwangk:
### AWS Service name

Step Functions

### Operations needed

Ran into an unsupported JSONata expression again. Used LLM to test out other JSONata expressions & the following seems not supported at the moment.

Maybe a library like https://github.com/rayokota/jsonata-python can be utilized.

Not supported (Unsupported JSONata expression):**
| Function | Category |
|----------|----------|
| `$exists()` | Object |
| `$type()` | Type checking |
| `$now()` | Date/Time |
| `$millis()` | Date/Time |
| `$uppercase()` | String |
| `$lowercase()` | String |
| `$substring()` | String |
| `$trim()` | String |
| `$sum()` | Numeric aggregation |
| `$average()` | Numeric aggregation |
| `$max()` | Numeric aggregation |
| `$min()` | Numeric aggregation |
| `$abs()` | Numeric |
| `$floor()` | Numeric |
| `$ceil()` | Numeric |
| `$round()` | Numeric |
| `$power()` | Numeric |
| `$sqrt()` | Numeric |
| `$sort()` | Array |
| `$reverse()` | Array |
| `$distinct()` | Array |
| `$keys()` | Object |
| `$values()` | Object |
| `$boolean()` | Type conversion |
| `$uuid()` | Utility

### Use case

Testing Step Functions workflows that use JSONata built-in functions. Ministack v1.3.43 supports core JSONata expressions (arithmetic, comparisons, string concatenation, ternary operators, $count, $merge, $length, $string, $not) but rejects most built-in functions with:

  Error: States.QueryEvaluationError
  Cause: Unsupported JSONata expression: $functionName(...)

Functions confirmed unsupported:

  String:    $uppercase, $lowercase, $substring, $trim, $contains, $split, $join, $replace, $pad
  Numeric:   $sum, $average, $max, $min, $abs, $floor, $ceil, $round, $power, $sqrt
  Array:     $sort, $reverse, $distinct, $append
  Object:    $exists, $keys, $values, $lookup
  Type:      $type, $boolean, $number
  Date/Time: $now, $millis
  Utility:   $uuid, $base64encode, $base64decode, $formatNumber

Highest-priority functions for real-world workflows:
- $exists() — checking if optional fields are present before routing (Choice states)
- $uppercase/$lowercase — normalizing string input
- $substring — parsing/extracting identifiers
- $sum/$max/$min — aggregating numeric data
- $keys/$values — iterating over dynamic objects
[...]

PR #1522, opened 2026-08-25
fix(stepfunctions): run JSONata expressions on a real JSONata interpreter

facuparedes:
## Problem

`QueryLanguage: JSONata` was evaluated by an ad-hoc operator splitter inside `stepfunctions.py`, with no tokenizer and no AST. Against the official JSONata conformance suite it scored 532 of the 1674 cases it is asked to cover, and passed 1 of 101 groups.

Some expressions raised `States.QueryEvaluationError`. Others answered a wrong value the execution then carried forward, including the most ordinary path in the language, mapping a field over an array:

```json
{
  "Type": "Pass",
  "QueryLanguage": "JSONata",
  "Output": { "total": "{% $states.input.items.n %}" }
}
```

On `main` at 97b0a80, with `{"items": [{"n": 3}, {"n": 9}]}` as the execution input:

```
status: SUCCEEDED output: {"total": null}
```

A `Choice` then compares against nothing and an `Assign` stores nothing, with no error and no log line. A state machine that runs on AWS failed here, or quietly produced different data, for reasons its author could not reproduce against the real service. On this branch the same execution answers `{"total": [3, 9]}`.

Also rejected: predicate filters (`items[status = 'OPEN']`), lambdas handed to `$map` / `$filter` / `$reduce`, range predicates (`arr[[0..1]]`), variable bindings (`($x := 2; $x * 3)`), descendant search (`**`), grouping (`items{status: n}`), and most of the function library.

Reported in #1519.

## Change

A JSONata 2.0.6 interpreter in `ministack/core/jsonata/`: tokenizer, Pratt parser, evaluator, 69 builtins, and the XPath picture-string engines behind `$formatNumber`, `$formatInteger`, `$formatDateTime` and `$parseInteger`. It is a library rather than an AWS service, so it sits in `core/` next to `iam_evaluator.py`, and it adds no runtime dependency: standard library only.

`stepfunctions.py` loses 1342 lines and gains 159. The package's public surface is four names: `evaluate`, `stringify`, `UNDEFINED`, `JsonataError`.

Four state types now evaluate their JSONata fields, which they did not before: a Map's `Items` and `Output`, an `Assign` on Map or Parallel, a Wait's `Seconds` and `Timestamp`, a Fail's `Error` and `Cause`.

`$states.context` exposes the five fields AWS documents and nothing else, so a workflow can no longer read back the variables an `Assign` wrote. `$states.result` and `$states.errorOutput` are unavailable outside the states AWS binds them in, rather than reading as `null`. A variable assigned inside a Parallel branch goes out of scope when the state completes, matching Map.

## Breaking changes

`$values` is no longer callable. It is not a JSONata function and not one of the six Step Functions adds (`$partition`, `$range`, `$hash`, `$uuid`, `$parse`, `$random`), so an expression using it never ran on AWS.

Three ways an expression now fails the execution, all as `States.QueryEvaluationError`, which a state can `Retry` and `Catch`:

- it selects nothing, instead of writing a `null` nobody asked for
- its result is something JSON cannot represent (a function value, `1/0`, `0/0`), instead of handing the caller a document no JSON parser accepts
- it nests deeper than 200 levels, instead of exhausting the worker's Python stack

A JSONata error carries its specification code (`T0410`, `D3137`, and the rest) into the cause, so the failure names the rule it broke.

### Conformance suite

`tests/jsonata_suite.json` carries all 1686 cases of the jsonata-js suite, folded from its 1319 loose files. 1674 pass. The 12 skipped exercise `$eval`, the one function of the specification Step Functions does not offer. They carry `absent_from_step_functions` and are skipped rather than dropped, so every run says how much of the reference this interpreter is not asked to cover.

```
$ pytest tests/test_jsonata_*.py -q
2199 passed, 12 skipped in 4.73s
```

That number covers the suite plus five hand-written modules for what the suite barely touches: `test_jsonata_parser.py` (precedence, associativity, error positions), `test_jsonata_evaluator.py` (scoping, tail calls, signature validation), `test_jsonata_builtins.py` (the argument edges of all 69), `test_jsonata_picture.py` (the four picture-string engines), `test_jsonata_rendering.py` (JavaScript number text).

### Step Functions

15 tests in `tests/test_stepfunctions.py`, each running a real state machine end to end. They cover the four state types that gained evaluation, the scoping rules for `$states.*` and for a Parallel branch's variables, and the three new failure modes, including that each one is catchable.

### Whole suite

```
$ pytest tests/ -q
1 failed, 8951 passed, 106 skipped in 905.58s
```

The failure is `test_dsql.py::TestContainersE2E::test_env_flag_spins_up_real_backend`, which reproduces identically on `main` at 97b0a80: `dsql.py` hands a new cluster port 25434 while an earlier proxy is still listening there.

### Deliberate departures from jsonata-js

Each one is measured against the reference, and in each one copying it is not an option. Nothing wraps a JSONata evaluation in a timeout, and a JavaScript `RangeError` is not a `JsonataError`, so it would surface as an uncatchable `States.Runtime`.

| expression | jsonata-js | here |
| --- | --- | --- |
| `$formatNumber(1/0,'0.0e0')` | never returns | `"NaN.0e0"` |
| `$formatInteger(1e7,'I')` | `RangeError`, `code: undefined` | `D1001` |
| `$parseInteger('12','００')` | `RangeError: Invalid code point` | `NaN` |
| `$fromMillis(0,'[Q01]')` | the literal string `"undefined"` | `D3132` |

Where the reference does answer, this matches it, including the parts XPath leaves to the implementation: `$formatInteger(1000000,'I')` is 1000 `M` either side.

## References

- [JSONata 2.0.6 documentation](https://docs.jsonata.org/)
- [jsonata-js test suite](https://github.com/jsonata-js/jsonata/tree/master/test/test-suite) (MIT), the source of `tests/jsonata_suite.json`
- [XPath `format-integer` and `format-dateTime` picture strings](https://www.w3.org/TR/xpath-functions-31/#formatting-integers), which JSONata cites for the picture engines
- [AWS: Transforming data with JSONata in Step Functions](https://docs.aws.amazon.com/step-functions/latest/dg/transforming-data.html)
- [AWS: JSONata reserved variables and the Context Object](https://docs.aws.amazon.com/step-functions/latest/dg/jsonata-reserved-variables.html)
- Issue #1519


Nahuel990:
@facuparedes I’m closing this PR and won’t merge it. I’m going to replace the current ~1,400 lines with an implementation based on the `jsonata-python` library that will cover your use cases.

This PR does exactly what I asked you not to do.

facuparedes:
Hi @Nahuel990, sorry for the confusion here, and thanks for bearing with us on this one.

Part of it is just bad timing on our side. We were already building the interpreter before this thread got going: the branch was committed at 18:19 UTC and your first reply landed at 18:37, so the PR was something we were finishing rather than an answer to what you asked for. We should have checked in with you first, and that one is on us.

On why we did not reach for `jsonata-python`: it came up back in #669 and the fix shipped in 1.3.44 without it, and CONTRIBUTING asks for minimal dependencies, so writing it in-tree seemed like the option that fit the repo best. What made us comfortable proposing it at all is that jsonata-js publishes its own conformance suite, so we could measure the implementation against the reference rather than against our own reading of the spec.

We were curious how the existing libraries do on that same suite, so we ran both of them. Same 1674 cases, with `$eval` excluded since Step Functions does not offer it:

```
jsonatapy 2.2.8         1664 passed, 10 failed
jsonata-python 0.7.0    1590 passed, 84 failed
```

`jsonatapy` (txjmb/jsonata-core) is the stronger of the two by a good margin, and it ships `timeout` and `max_stack_depth` right in its API, which is exactly the pair of guardrails we ended up building by hand. It also validates against jsonata-js's suite, pulled in as a git submodule, so at least everyone is measuring against the same thing. The catch for MiniStack is packaging. It is Rust behind maturin and publishes wheels for macOS arm64, manylinux and Windows, but not for musl, and `Dockerfile:29` is `python:3.13-alpine`:

```
$ docker run --rm python:3.13-alpine pip install --only-binary=:all: jsonatapy
ERROR: Could not find a version that satisfies the requirement jsonatapy (from versions: none)
```

Adopting it would mean a Rust toolchain in the Alpine builder stage, which felt like a bigger ask than the problem we were solving.

`jsonata-python` installs there without any trouble, being pure Python. Most of its remaining gap is error handling: 60 of the 84 raise an exception carrying no JSONata code at all, usually a plain Python one (`IndexError`, `RuntimeError`, `ValueError`, `TypeError`). In Step Functions that difference reaches the user, since `States.QueryEvaluationError` is something a state can `Retry` and `Catch` and an internal traceback is not. One case also does not return at all:

```
( $inf := function(){$inf()}; $inf() )    # the suite expects U1001
```

Both projects are also young: `jsonata-python` has 104 stars, 13 releases and one contributor with 130 commits against 5 for the next one, and `jsonata-core` started this January with 9 stars. They are genuine ports and they clearly work, we were just a little nervous about putting a service's whole query language behind either one.

On keeping this much code alive over time, which is the fair worry here: the suite is jsonata-js's own rather than something we wrote, and it runs in CI on every push, so a regression shows up as a named failing case instead of as a bug report months later. That part holds whichever way you go, since we would wire the same suite in for a library too.

All that said, it is your repo and very much your call. If the size of the diff is more of a problem than the approach, we can split it in two so the review is smaller:

- `ministack/core/jsonata/` with its tests and the suite: +9,096 lines that nothing imports yet and that change no behaviour
- the wiring in `stepfunctions.py`: +583 / -1,351, so the part that actually changes what MiniStack does is a net removal of 768 lines

And if you would rather carry a dependency after all, just tell us which one and we will open a fresh PR built on it instead, with the same suite in CI so the gaps above stay visible. Any of these works for us.

Thanks again for taking the time on this one!

Issue #1528, opened 2026-08-26
Amazon Location Service: trackers + device positions (control plane + AWS::Location::Tracker)


iot-rocket:
## Which AWS service

Amazon Location Service, API namespace `geo` (`boto3.client("location")`,
signing name `geo`, endpoint prefix `geo.<region>`). Tracking only.

## Which operations

The tracker resource plus its device-position data plane:

- `CreateTracker`, `DescribeTracker`, `ListTrackers`, `UpdateTracker`,
  `DeleteTracker`
- `BatchUpdateDevicePosition`, `GetDevicePosition`, `BatchGetDevicePosition`,
  `GetDevicePositionHistory`
- CloudFormation: `AWS::Location::Tracker` (Ref = tracker name; GetAtt Arn,
  TrackerArn, CreateTime, UpdateTime)

That is the full surface our workload touches. Not asking for the rest of the
service.

## Real use case

We run an IoT fleet backend on MiniStack. One Lambda (an IoT topic rule
target) resolves a device's position estimate and writes it to a tracker via
`BatchUpdateDevicePosition`; an operations script reads positions back with
`GetDevicePosition` and handles `ResourceNotFoundException` for devices that
never reported. Today the service does not exist on MiniStack, so the
deployment carries three workarounds: the `AWS::Location::Tracker` resource is
condition-gated out of the template, the rule that triggers the Lambda deploys
disabled, and the Lambda's IAM statement cannot be scoped to the tracker
because `!GetAtt Tracker.Arn` cannot resolve. The whole geotracking chain is
dead in local and CI runs.

## Scope boundaries

- No geofencing, maps, places, or routes. Trackers only.
- No KMS integration, no EventBridge position events, no pricing plans, no
  consumer association.
- Position store: latest position per device plus a bounded history for
  `GetDevicePositionHistory`. `SampleTime`/`ReceivedTime` as timestamp shapes,
  `Position` as the `[lon, lat]` double list, `Accuracy` and
  `PositionProperties` stored and echoed.
- `BatchUpdateDevicePosition` returns the required `Errors` list (empty on
  success); `GetDevicePosition` answers `ResourceNotFoundException` for an
  unknown device or tracker.

Implementation is ready on our side (service file, router, CFN provisioner,
tests, README/CHANGELOG), with the error wordings and response defaults
pinned against the live service; the PR follows as soon as the scope here is
agreed. API references for the scoped surface:
[...]

PR #1540, opened 2026-08-26
feat(location): trackers and device positions (control plane + AWS::Location::Tracker)

iot-rocket:
Part of #1528 (the Location tracking issue)

## What

Adds Amazon Location Service to MiniStack: trackers only, as scoped in the
issue: `CreateTracker`, `DescribeTracker`, `ListTrackers`, `UpdateTracker`,
`DeleteTracker` on the control plane and `BatchUpdateDevicePosition`,
`GetDevicePosition`, `BatchGetDevicePosition`, `GetDevicePositionHistory` on
the device-position data plane, plus the `AWS::Location::Tracker`
CloudFormation resource.

## Why

We run an IoT fleet backend on MiniStack. One Lambda (an IoT topic rule
target) resolves a device's position estimate and writes it to a tracker via
`BatchUpdateDevicePosition`; an operations script reads positions back with
`GetDevicePosition` and handles `ResourceNotFoundException` for devices that
never reported. With no `location` service, the whole geotracking chain is
dead in local and CI runs, and the deployment carries three workarounds: the
`AWS::Location::Tracker` resource is condition-gated out of the template, the
rule that triggers the Lambda deploys disabled, and the Lambda's IAM statement
cannot be scoped to the tracker because `!GetAtt Tracker.Arn` cannot resolve.
This PR removes the need for all three.

## How

- **Routing.** The client is named `location` but signs with credential scope
  `geo` (botocore signingName), and the model's per-operation host prefixes
  (`cp.tracking.` control plane, `tracking.` data plane) sit in front of
  `geo.{region}`, so the host pattern matches `geo\.` unanchored, and under
  an endpoint override (where the host carries no `geo.` at all) a new
  `"geo" -> "location"` entry in the credential-scope map is the primary
  routing signal. Both paths are covered in `tests/test_router.py`.
- **Wire shapes**, verified against the botocore `location` 2020-11-19 model:
  `SampleTime`/`ReceivedTime`/`CreateTime`/`UpdateTime` are `timestamp` shapes
  with an explicit `timestampFormat: iso8601`, so responses emit ISO8601 UTC
  strings (epoch numbers are tolerated on input; the store keeps a canonical
  epoch float). `Position` is the modeled two-member `[longitude, latitude]`
  double list. A test asserts boto3 parses every timestamp member back into a
  datetime and that `SampleTime` survives the round-trip exactly.
- **Batch semantics.** `BatchUpdateDevicePosition` returns the required
  `Errors` list: empty on success, and a per-entry
  `{DeviceId, SampleTime, Error: {Code: "ValidationError", Message}}` for a
  malformed or out-of-range entry while well-formed siblings still land. The
  wording is the live service's ("latitude must be between -90 and 90
  degrees", measured 2026-08-26). botocore refuses to send a structurally
  broken entry, so that test goes over the raw wire.
  `BatchGetDevicePosition` mixes found positions with `ResourceNotFoundError`
  error entries for unknown devices as the API models them; a live call was
  observed simply omitting unknown devices instead, so the modeled behaviour
  is the stricter of the two.
- **Position store.** Latest position plus a bounded history per device; the
  newest 100 samples, oldest dropped first, documented in the module
  docstring. The real service bounds by time instead (30-day retention, and
  its default TimeBased filtering keeps at most one position per 30 seconds
  per device); we store every sample, which is the more useful behaviour for
  assertions and is named as a divergence. `GetDevicePositionHistory` returns
  ascending `SampleTime` order with `StartTimeInclusive` inclusive and
  `EndTimeExclusive` exclusive, defaulting to the documented last-24-hours
  window; a single page tops out at the documented `MaxResults` cap of 100.
- **Errors**, wordings from the live service: `ResourceNotFoundException`
  ("Tracker not found: X.") for an unknown tracker on every per-tracker
  operation and ("records not found for given deviceId") for an unknown
  device on `GetDevicePosition`; `ConflictException` ("Tracker already
  exists: X.") on duplicate `CreateTracker`; `DescribeTracker` reports the
  required `Description` member as an empty string when unset, plus the
  measured defaults (`PositionFiltering: TimeBased`,
  `EventBridgeEnabled: false`, `Tags: {}`).
- **Stored but inert**, named as divergences: `PricingPlan` (deprecated on
  AWS), `KmsKeyId`, `PositionFiltering`, `EventBridgeEnabled` and `Tags` are
  stored and echoed but drive no behavior (no KMS encryption, no EventBridge
  position events, no position filtering), and there are no tracker
  consumers, geofences, maps, places, or routes.
- **CloudFormation.** `AWS::Location::Tracker` provisions through the service
  module, so the resource is readable back through the real API. `Ref`
  returns the tracker name; `Fn::GetAtt` exposes `Arn`, its documented
  synonym `TrackerArn`, `CreateTime` and `UpdateTime`.
  `Description`/`PositionFiltering`/`EventBridgeEnabled` update in place; a
  `TrackerName` change is refused with a replacement error rather than
  silently discarding the tracker's device positions (the same protection the
  custom-named replacement guard gives a DynamoDB table); create adopts an
  existing same-name tracker so the update-fallthrough stays idempotent.
- **Persistence.** Region-scoped store (`AccountRegionScopedDict`), saved
  under `PERSIST_STATE=1` and restored at import; covered by the parametrized
  restore-path test and a dedicated round-trip test in
  `tests/test_persistence.py`.
- **Tests**: `tests/test_location.py` (tracker CRUD incl. name validation,
  `UpdateTracker`, batch update + get, out-of-range and malformed entries
  over the raw wire, unknown-device 404, ordered/windowed history with the
  100-sample bound), the router cases, the persistence round-trip, and a CFN
  deploy → describe-via-client → in-place update → delete test. README
  service + CFN table rows and a CHANGELOG entry under Unreleased.

## References

- CreateTracker API reference (PositionFiltering semantics): https://docs.aws.amazon.com/location/previous/APIReference/API_CreateTracker.html
- BatchUpdateDevicePosition API reference (Errors contract, 30-day retention): https://docs.aws.amazon.com/location/previous/APIReference/API_BatchUpdateDevicePosition.html
- GetDevicePositionHistory API reference (window defaults, MaxResults): https://docs.aws.amazon.com/location/previous/APIReference/API_GetDevicePositionHistory.html
- AWS::Location::Tracker CloudFormation reference (Ref/GetAtt, mutability): https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-location-tracker.html

(The `location/previous` tree is the one matching the botocore `location`
2020-11-19 model this implements; the `latest` tree describes the split
successor services.)


iot-rocket:
@Nahuel990
> There is a hard limit of 5 PRs per user to avoid this type of PR proliferation, unless we have agreed on a specific scope beforehand.

First of all: I am aware of the recently introduced unlimited->1-> 5 PRs per user and kept the unlimited draft-PR openings. Also I can understand, that it is always a goal to have the issues and PR count -> 0.

The limit of PRs usually also comes with some side effects to contributors: The PRs become bigger and not so easy to review, or the already done work has to get opened one by one over days/weeks. (Your activity in this project is btw. special and super fast! :) )
Also keep in mind there would be also a way around it: we are several people with several GH accounts... (in our case the PRs should go all through my account - that also explain the amount from my side. We prepare them in advance, which explains the created PR amount in short time)

I still try to split the PRs into small logical ones, so they should be also relatively easy to review. 

In that case I actively decided to create them anyway as a draft, so the ministack community sees already, what we are working on, so they can e.g. cherry pick them to their own project and the community can e.g. focus on other awesome features and/or other fixes.

The main reason is visibility and deduplication of work. Human and AI resources are still limited.
If we would just keep them hidden in one of the 400+ (wow!) forks, it would be a waste of resources - we think. Also we want to contribute to your amazing project and bring our work mainline(even it brings also some additional effort to us)
We had some internal discussion, that in case it would be our project, we would be generally welcoming that.
Of course it comes with review effort for the maintainers, especially for you in that case. But it will come anyway, if your project gains more and more attraction, we think.

***

> In this case, the PRs contain a mix of things that are outside the scope we are aiming for. We are not trying to become LocalStack PRs, and we have already discussed the expected shape of services and their corresponding test files.

I understand what you mean by that, but on the other hand, your README.md describes itself like:
> LocalStack recently moved its core services behind a paid plan. If you relied on LocalStack Community for local development and CI/CD pipelines, MiniStack is your free alternative.

Some of our previously merged PRs unlocked the next wave, therefore we created the next PR batch.

***

> Please keep contributions aligned with our existing standards and the agreed structure. We appreciate contributions, but we need to maintain consistency and control over the direction of the project. We have had to block contributors in the past when this wasn't respected.

We generally try to deliver high quality PRs which are aligned to this projects standards best, we can. For that:
- we have a huge aws stack with 1000s of unit, integration and e2e tests
- we test everything against a AWS sandbox account first
- we deploy everything via CFN
- we have multiple internal reviews(AI, Human) before opening those PRs

I hope you also recognized, after the start of our first PRs, the new ones are pretty high quality ones. (there may be exceptions, but in that case some hints will help and we try to align in that case)

Also we decided for your amazing project and not e.g. some other projects, because we like the shape of it and the techstack aligns with our proejcts.

I also recognized you closed some of our PRs of `niche services`, waiting for traction. I could understand that, in case you don't have a PR yet and you would need to implement that by yourself, but in case the PR is already existing?

I hope you can follow my arguments, but also I would be very interested and like to understand your point of view / arguments here, so we can even better contribute in the future.

PR #1682, opened 2026-09-10
Record direct-to-phone SNS publishes and serve them back

himangshuj:
## What this adds

SNS is the one messaging service MiniStack does not let you read back. SES keeps what it was sent and serves it at `/_ministack/ses/messages`; SQS does the same at `/_ministack/sqs/messages`. A direct-to-phone `Publish` returns a valid `MessageId`, logs a line, and stores nothing — so an SMS is the only thing you can send and then not observe.

This records those publishes and serves them back, following the two existing handlers rather than introducing a new shape.

**Endpoint:** `GET /_ministack/sns/sms-messages`, and — for the same reason `_HEALTH_PATHS` already answers `/_localstack/health` — also at `/_aws/sns/sms-messages`, in that endpoint's `{"sms_messages": {"<phone>": [...]}, "region": "..."}` shape. Filterable by account, like its siblings.

**Storage** is an `AccountRegionScopedDict`, so records are scoped the same way the rest of the service's state is. It participates in state save/load and is cleared by `/_ministack/reset`.

**Record fields** are those LocalStack returns — `PhoneNumber`, `TopicArn`, `SubscriptionArn`, `MessageId`, `Message`, `MessageAttributes`, `MessageStructure`, `Subject` — with `TopicArn` and `SubscriptionArn` null for a direct-to-phone send, and an absent `Subject` reading as null rather than an empty string. They are stored in the shape they are served in, so nothing is reshaped on the way out.

## Why the LocalStack-compatible path is included

Test suites that assert on SMS content read `/_aws/sns/sms-messages` today. Answering it alongside the `_ministack` path means those suites keep working unedited, which is the same compatibility the health endpoint already provides.

## Tests

Seven new tests in `tests/test_sns.py` cover the round trip, the account and region scoping, the phone-number grouping, the null fields, the `MessageAttributes` parsing, both paths returning the same body, and reset clearing the log.

The existing suites were run before and after: the set of failures is byte-identical to stock, so nothing here changes an existing outcome. `ruff check` is clean.

## Not included

Nothing about topic or subscription delivery changes — this is only the direct-to-phone path, which is the one that previously had nowhere to go.

### DIFF
--- a/ministack/app.py
+++ b/ministack/app.py
@@ -1097,6 +1097,100 @@ async def _handle_sqs_messages_request(method: str, path: str, headers: dict, qu
     return 200, {"Content-Type": "application/json"}, json.dumps(response).encode()
 
 
+_SNS_SMS_PATHS = ("/_ministack/sns/sms-messages", "/_aws/sns/sms-messages")
+
+
+async def _handle_sns_sms_messages_request(method: str, path: str, headers: dict, query_params: dict):
+    """Handle the SNS SMS log endpoint.
+
+    A `Publish` that carries a `PhoneNumber` and no `TopicArn` has nowhere to
+    deliver to, so sns.py records it and this serves the recording back. Same
+    read-only introspection as `/_ministack/ses/messages` and
+    `/_ministack/sqs/messages`.
+
+    Also answered at LocalStack's `/_aws/sns/sms-messages`, in LocalStack's
+    `{"sms_messages": {"<phone>": [...]}, "region": "<region>"}` shape, for the
+    same reason `/_localstack/health` is answered alongside
+    `/_ministack/health`: a suite that asserts an SMS was sent should not have
+    to be rewritten to run here.
+
+    Filters:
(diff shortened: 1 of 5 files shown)


ministack/app.py
```diff
@@ -1110,6 +1110,100 @@ async def _handle_sqs_messages_request(method: str, path: str, headers: dict, qu
     return 200, {"Content-Type": "application/json"}, json.dumps(response).encode()
 
 
+_SNS_SMS_PATHS = ("/_ministack/sns/sms-messages", "/_aws/sns/sms-messages")
```
Nahuel990: @himangshuj here is where you need to drop the _aws path which adds noise. Please follow current implementation for similar endpoints

Nahuel990 (approved):

Nahuel990:
I am merging this but I am striping a few things before release (like the aws test and that block of comments) please refrain from open new PRs without reviwing Contributing.md next time

Issue #1642, opened 2026-09-04
Lambda: durable invocations bypass the warm pool and time out on a loaded runner (flaky test_lambda_durable_* in CI)


iot-rocket:
## Summary

Three tests fail together on a loaded CI runner and pass on the next one:

```
FAILED tests/test_lambda.py::test_lambda_durable_runtime_env_vars_present - KeyError: 'arn'
FAILED tests/test_lambda.py::test_lambda_durable_chained_invoke_runs_child - AssertionError: expected ChainedInvokeSucceeded, got ['ExecutionStarted']
FAILED tests/test_lambda.py::test_lambda_durable_event_wrapped_with_sdk_fields - KeyError: 'keys'
```

Seen on runs 33925222515, 33925229002 and 33925245768 (three PRs filed within a minute, shard `parallel 1/3`), and on 33923392522 (`poc/lambda-enhance`). Two sibling PRs with the identical shard plan passed at the same time. In the `ministack.log` artifact of a red run every durable invocation ends the same way, 3 s after the function was created:

```
22:26:22 INFO [lambda] Lambda function created: durable-env-08eda277 (python3.12)
22:26:25 INFO [lambda] Lambda durable-env-08eda277 output:
Lambda timed out (no stderr/stdout captured).
```

18 of 18 durable invocations on that runner; the other 15 tests in the file only pass because they do not read the invoke result. The green runs have no timeout at all. Non-durable functions on the same red runner kept working (`Lambda worker spawned … cold start` lines continue through the window).

## Cause

[...]

PR #1703, opened 2026-09-12
fix(lambda): run durable invocations on the warm pool

iot-rocket:
Part of #1642 (durable invocations bypass the warm pool and time out on a loaded runner).

A durable invocation of a python or nodejs function does not go to the warm pool. `_execute_function_dispatch` sends it to `_execute_function_local` instead, because the three `AWS_LAMBDA_DURABLE_*` variables change on every call and a pooled worker's environment is fixed when it spawns. That executor starts a fresh interpreter for every invocation and its wrapper is prefixed with `INVOKE_DEPTH_BOOTSTRAP`, which imports `botocore.handlers` before the handler module, and all of it runs inside `subprocess.run(..., timeout=config.get("Timeout", 3))`. A durable function created without a `Timeout` therefore has 3 seconds for interpreter start, the botocore import, the handler import and the handler itself. On an idle host that is well under a second; on a two-vCPU runner with three xdist workers it crosses the budget, and then every durable invocation fails the same way until the load drops. That is the `test_lambda_durable_*` flake: in the `ministack.log` of a red run, 18 of 18 durable invocations end in `Lambda timed out (no stderr/stdout captured)` exactly 3 seconds after the function was created, while non-durable functions on the same runner keep working.

What changes:

- `ministack/core/lambda_runtime.py`: `DURABLE_CTX_EVENT_KEY` (`_ministack_durable_ctx`) and `DURABLE_ENV_VARS` next to `INVOKE_DEPTH_EVENT_KEY`, substituted into both worker bootstraps by the token helper, which is renamed `_sub_runtime_tokens` now that it carries more than the depth names. Both bootstraps pop the key off the event and move the three variables into `os.environ` / `process.env` before the handler runs, and delete them when the key is absent, so a worker reused for a non-durable invocation does not see the previous one's context. This is the channel the X-Ray trace header and the invoke-depth counter already use, for the same reason. `DURABLE_ENV_VARS` maps each variable to its key in the durable context, so the overlay that sets them and the bootstraps that clear them read one declaration: a fourth variable added to only one of the two would be set and never cleared, which is the leak the event carrier exists to avoid.
- `ministack/services/lambda_svc.py`: `_execute_function_warm` puts `_durable_env_overlay()` into the event when the invocation is durable, and the python and nodejs arm of the dispatch stops filtering durable invocations out of the pool, so a durable function is served like any other python or nodejs function: one cold start per worker, then warm. `provided.*` keeps the one-shot executor #1673 left it on (that arm was unconditional before #1673, which gave the non-durable half a warm pool and kept durable on the one shot), and the comment there now says why the two runtimes differ: a bootstrap process reads its environment once, at spawn. `_durable_env_overlay()` stays as it is for the docker, provided-runtime and one-shot paths. The injection sits in `_execute_function_warm` rather than in the dispatch, so the two Docker-unavailable fallbacks, which send python and nodejs to the same pool, carry the context as well; today they invoke the handler with the three variables unset. The durable SDK reads its context out of the event, which those fallbacks pass through, so nothing is lost there today; the variables are this emulator's own convenience and they belong wherever the handler runs.
- `README.md`: the sentence #1673 added under the warm-start section said durable invocations retain the one-shot executor because their execution context changes per call. That is still true for a custom runtime and no longer true for python and nodejs, so it now says which is which. #1673's own `[Unreleased]` bullet says the same thing unqualified, nine lines above this branch's, so it gets the `provided.*` qualifier too rather than shipping two contradictory sentences in one set of release notes.
- `tests/test_lambda.py`: #1636 had raised `Timeout` to 30 on `test_lambda_durable_chained_invoke_runs_child` (parent and child) and `test_lambda_durable_event_wrapped_with_sdk_fields` so a loaded runner's subprocess cold start would fit; with the pool serving durable calls those three go back to the default, and the comment that blamed the budget goes with them. `test_lambda_durable_runtime_env_vars_present` never got the bump and is the one of the three #1642 names that still showed the timeout on main.
- The one-shot executor is not touched. After this change no python or nodejs invocation reaches it any more (the two Docker-unavailable fallbacks route those runtimes to the pool first, and every other runtime gets its mock answer), so it is left as it is rather than changed on a path nothing runs.

Measured on a server run from the branch: `tests/test_lambda.py` and `tests/test_lambda_provided_dispatch.py`, 416 passed, 2 failed, 7 skipped; `main` on the same host, 407 passed, 2 failed, 7 skipped, and the two failures are the same two by name (`test_lambda_warm_start` and `test_apigwv2_nodejs_lambda_proxy`, both `urlopen error Name or service not known`, this host's name resolution and not either tree). #1673's own `tests/test_lambda_provided_dispatch.py` passes unchanged, including the case that pins durable `provided.*` to `_execute_function_provided`.

Three new tests, each run against the unpatched package and failing there. `test_lambda_durable_invocation_reuses_the_warm_worker` creates a durable python function with `Timeout: 1` and a module-level counter, invokes it twice and asserts the counter reaches 2 with two different execution ARNs and checkpoint tokens. `test_lambda_durable_invocation_reuses_the_warm_worker_nodejs` does the same for nodejs, with the variables read from `process.env`. On `main` both fail with `assert 1 == 2`, because every invocation is a new process and the module state is gone; on a slower host the first invocation times out inside the 1 second instead and the test fails on the missing key. `test_lambda_durable_context_is_dropped_when_the_next_invocation_is_not_durable` invokes once, switches `DurableConfig` off through `UpdateFunctionConfiguration`, invokes again and asserts that the same worker served it (counter at 2) with all three variables gone; on `main` it fails with `assert 1 == 2`, because the durable call ran in a one-shot process. `test_lambda_durable_event_wrapped_with_sdk_fields` now also asserts that no `_ministack` carrier key reaches the handler, a guard rather than a proof, since `main` injects no such key.

Two more in `tests/test_lambda_provided_dispatch.py`, next to #1673's own parametrized routing test, so both arms of the same `if/elif` are pinned in one file rather than one arm being pinned directly and the other inferred from a module counter. `test_python_and_nodejs_dispatch_to_the_warm_pool_durable_or_not` asserts the target executor and that no other executor is called, over python and nodejs, durable and not: on `main` the two durable cases fail with `_execute_function_local`, the two non-durable ones pass there and are guards. `test_docker_unavailable_fallback_carries_the_durable_context` drives both permissive fallbacks (no docker SDK, and an unreachable daemon) and asserts python reaches the pool rather than the one shot, with the overlay the context produces; it passes on `main` too, so it is a guard for the second half of the change rather than a proof of it.

Not covered: the durable-functions page describes a durable function as a regular function with a checkpoint/replay mechanism and documents no environment variable for it; the three variables are MiniStack's own, next to the event the SDK reads its inputs from. The per-invocation variable AWS itself documents is `_X_AMZN_TRACE_ID` ("changes with each invocation"), which the emulator already carries through the event the same way. The docker executor keeps `_durable_env_overlay()` in the container environment, which is per-container and correct there, so it is unchanged; the same for the provided-runtime executor. The resume path of a durable execution (`invoke_durable_resume`) now runs on the pool as well and has no test of its own. Nothing changes when `LAMBDA_EXECUTOR=docker` or in strict mode. The one-shot executor still charges its cold start against the function `Timeout`, which real Lambda does not (init has its own limit); nothing reaches it from this path any more, and it is not changed here.

References: https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html, https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html


Nahuel990 (approved):

tests/test_lambda_provided_dispatch.py
```diff
@@ -129,6 +129,68 @@ def test_provided_dispatch_preserves_other_executors(monkeypatch, mode, target):
         executors[target].assert_called_once_with(func, {}, "request-1")
```
Nahuel990, line 131: I'll remove this file before release as I stated in the unrleased PR that gave it origin

PR #1712, opened 2026-09-13
fix(lambda): honor cross-account layer grants and retain attached content

iot-rocket:
## Summary

`CreateFunction` and `UpdateFunctionConfiguration` refused every foreign-account layer version with `AccessDeniedException`, `AUTH` on or off; Lambda Insights is published under `580247275435`. `AddLayerVersionPermission` policies were stored but never read. Attachment, `GetLayerVersion` and `GetLayerVersionByArn` now evaluate that policy via a new `evaluate_resource_policy` in `iam_evaluator`, which keeps statements whose `AWS` principal names the caller and hands them to `evaluate`; `Service` principals are ignored, and `aws:PrincipalOrgID` resolves in the evaluator. `AWS::Lambda::Function` uses the same path. A deleted version is released once nothing references it, also after CloudFormation deletes and state restore.

## AWS behaviour (measured)

One account, eu-central-1.

| case | AWS | before | after |
| --- | --- | --- | --- |
| Ungranted foreign layer | `AccessDeniedException` `User: <caller ARN> is not authorized to perform: lambda:GetLayerVersion on resource: <layer version ARN> because no resource-based policy allows the lambda:GetLayerVersion action` | blanket denial | as AWS |
| Granted foreign layer | unmeasured | refused | attaches |
| Function after `DeleteLayerVersion` | invokes with the layer, config lists it | n/a | same, also cold start and revoke |
| Read deleted version | `ResourceNotFoundException` `The resource you requested does not exist.` | n/a | as AWS |
| Attach deleted or unpublished version | `InvalidParameterValueException` `Layer version <ARN> does not exist.` | n/a | as AWS |
| Own layer, other region | `Layers are not in the same region as the function. Layers are expected to be in region <region>.` | n/a | as AWS |
| Foreign layer, other region | `AccessDeniedException` | refused | as AWS |
| Layer ARN as `LayerName` | accepted | refused | accepted |
| Stored policy | `arn:aws:iam::<account>:root`; org grant as `Principal: "*"` plus `aws:PrincipalOrgID`, unknown id accepted | stored | same |

Deliberate deviation: a layer name never registered locally still attaches with `CodeSize` 0 and no content, so a stack referencing the public Insights layer deploys; AWS refuses it. A later publish does not add bytes. Say if you prefer a refusal.

`UpdateFunctionConfiguration` now checks the combined unzipped size first, like `CreateFunction`.

## Not covered

- A real cross-account grant and revocation (needs a second account).
- A real organization condition; local Organizations has no members, so an org grant reaches only its creator.
- The CloudFormation failure reason for a denied layer.

## Tests

11 test functions, 24 cases (`test_lambda.py` +455/-16, `test_cfn.py` +57); implementation +202/-93. Negative cases replace what the blanket denial guaranteed; name any to drop.

`tests/test_lambda.py` + `tests/test_iam_auth.py`: 820 passed / 7 skipped on the branch, 799 / 7 on `main` at `efddc4a8`. `tests/test_cfn.py -k "lambda or layer"`: 45 against 43. No failures. `ruff check ministack/` clean.


iot-rocket:
Maybe some more explanation here:
The branch now checks AddLayerVersionPermission policies with either AUTH setting. Both layer read APIs use the same check, including organization conditions.

I also fixed the execution path: granted foreign layer bytes reach the runtime, and existing functions retain them after revocation or deletion. Missing versions of locally known layers are denied. Tests cover cold starts, published versions, both read APIs, and denied updates.

Unknown external layer names still allow metadata-only attachment for local stack deployment. They cannot acquire private content published later without reattachment and a successful permission check. This fallback remains a deliberate difference from AWS.

Validation: all 63 layer cases pass, plus five HTTP scenarios with AUTH=true.

iot-rocket:
Follow-up to my comment from before, after another re-audit of the branch.

The policy was stored and served back, and neither the attachment path nor `GetLayerVersionByArn` read it. The branch evaluates it on both, with either `AUTH` setting. The evaluation now goes through a small public helper in `iam_evaluator.py` (`evaluate_resource_policy`, +43/-15) instead of the evaluator's private functions.

On size: the implementation is +119/-73 in `lambda_svc.py` plus that helper. The rest is tests, +561/-16, most of them parametrized cases: the negative cases the old blanket denial gave for free, retention after revoke and delete, and persistence. If you want fewer, tell me which and I take them out.

Retention after `DeleteLayerVersion`, the exact error sentences for a deleted, a missing, a cross-region and a no-grant layer, `GetLayerVersion` accepting the layer ARN as `LayerName`, and the download URL being fetchable without credentials. The branch answers the same sentences now, and a deleted version is released once nothing references it. What one account cannot measure is a real cross-account grant and an organization condition.

The one deliberate difference stays: a layer name never published here attaches with `CodeSize` 0, so a stack referencing the public Insights layer deploys locally. Say if you would rather have that refused and I remove it.

The branch is force-pushed as one commit on current main, and the description above is rewritten to match the code.

Nahuel990:
@iot-rocket  I am merging this but I'm dropping a few non AWS behaiour. Also you are adding 3 more things here than just cross account check. please review the description your AI generates matches what you do in the code. we talked before about hidding features as fixes

Nahuel990 (approved):

Issue #1140, opened 2026-07-22
EKS update-kubeconfig


StraggleCraft:
### Affected Service

EKS

### What happened?

I'm running ministack on a rootful podman instance on win10.  The EKS cluster is created successfully, however, I've noticed that aws eks update-kubeconfig --name eks-cluster and aws eks get-token --cluster-name eks-cluster populate cluster credentials that don't allow for connecting to the cluster.  I was able to pull the credentials directly from the sidecar, and manually put them into my kube config.  I'm unclear if this is a required step because the emulation is not able to pass through the authentication, or if it should be working out of the box.  The documentation and blog posts i've seen don't make mention of needing to manually pull the static client certificate data. 

### Reproduction steps

```python
aws eks update-kubeconfig --name eks-cluster
```

### Expected behavior

aws eks update-kubeconfig populates .kube\config with valid cluster authentication values

### MiniStack version

latest

### SDK / tool used

aws-cli/2.35.6

Nahuel990:
The CA certificate and endpoint in the kubeconfig are correct. We pull the actual k3s CA and expose a reachable endpoint.

The authentication flow is the part that's currently different. `aws eks get-token` generates an IAM-based token, but the k3s sidecar doesn't include `aws-iam-authenticator` (or an equivalent authentication webhook) to validate it, so `kubectl` can't authenticate using that token

Using the k3s client credentials directly, as you did, is the expected approach today. Supporting `aws eks update-kubeconfig` out of the box would require either injecting the k3s admin credentials into the generated kubeconfig or running an authentication webhook that can bypass IAM tokens (we don't validate them) 

Keeping this open as it's a nice to have

jgrumboe:
Revisiting this now that MiniStack has IAM authentication.

First, a correction on one of the two options above: injecting k3s admin credentials into the generated kubeconfig isn't reachable. `aws eks update-kubeconfig` builds the user entry as an unconditional `exec` stanza shelling out to `aws eks get-token` - there's no code path that emits a client cert or a static token - and it reads only `certificateAuthority.data`, `endpoint` and `arn` off the DescribeCluster response. MiniStack can influence the cluster entry, but nothing it returns can supply credentials. Making the token itself validate is the only server-side fix.

[...]

PR #1717, opened 2026-09-13
fix(eks): support AWS CLI kubeconfig authentication

jgrumboe:
`aws eks update-kubeconfig` generates an IAM exec-token configuration, but the k3s API server previously could not authenticate those tokens. This change connects k3s to a MiniStack TokenReview webhook so the standard AWS CLI kubeconfig works without extracting a container admin kubeconfig.

Implements #1140.

### Behavior

- Inject the TokenReview webhook configuration before k3s starts and preserve the cluster account and region in its callback URL.
- With `AUTH=false`, accept nonempty local bearer tokens with administrator access.
- With `AUTH=true`, validate the presigned STS token locally, including its signature, cluster binding, lifetime, and IAM or STS credentials. AWS EKS API operations retain IAM policy enforcement; Kubernetes access requires a creator grant or an Access Entry.
- Record the actual cluster creator, honor disabled bootstrap access, and match assumed-role sessions to IAM role paths. Cluster service roles and roots from other accounts receive no implicit administrator grant.
- Materialize the five general-purpose EKS managed access policies from the [AWS-published Kubernetes rules](https://docs.aws.amazon.com/eks/latest/userguide/access-policy-permissions.html):
  - `AmazonEKSClusterAdminPolicy`
  - `AmazonEKSAdminPolicy`
  - `AmazonEKSEditPolicy`
  - `AmazonEKSViewPolicy`
  - `AmazonEKSAdminViewPolicy`
- Support cluster and namespace scopes, namespace wildcard patterns, additive user-supplied Kubernetes groups, and automatic reconciliation for new namespaces, scope changes, disassociation, Access Entry deletion, and k3s restart.
- Keep all EKS tests in `tests/test_eks.py` and update the README EKS example to wait for the CA before generating kubeconfig.

### Commit structure

1. TokenReview authentication implementation
2. Authentication and mode tests
3. Kubeconfig documentation
4. Managed access-policy RBAC implementation
5. Access-policy authorization and revocation tests
6. Access-policy documentation

### Validation

- 40 focused EKS authentication and access-policy tests passed.
- Complete `tests/test_eks.py`: 86 passed; the only failure was the pre-existing OIDC JWKS test because the isolated light environment did not include the optional `cryptography` package.
- `ruff check ministack/services/eks.py tests/test_eks.py` and `git diff --check` passed.
- Live k3s checks passed for `kubectl get nodes` in both AUTH modes, account and STS role-path isolation, namespace-scoped View permissions, Secret denial, wildcard reconciliation for a newly created namespace, cross-namespace denial, and revocation after disassociation.

### Implementation note

Real EKS evaluates access policies in a private EKS authorizer and does not expose them as Kubernetes RBAC objects. k3s has no equivalent authorizer, so MiniStack creates internal managed ClusterRoles and bindings with the same published rules and scope outcomes. As a result, these permissions can appear in Kubernetes RBAC inspection commands even though they do not on AWS. Other, non-implemented EKS access policies continue to round-trip but do not grant automatic Kubernetes permissions. Existing clusters without a recorded creator need an explicit Access Entry. k3s may cache authentication results for up to five minutes; policy revocation removes the RBAC binding independently of that cache.


ministack/services/eks.py
```diff
@@ -75,6 +94,23 @@
 _port_counter = [EKS_BASE_PORT]
 _oidc_keypair_lock = threading.Lock()
 _oidc_keypair = None                  # (private_key, jwk_dict, kid)
+_rbac_reconcilers_lock = threading.Lock()
+_rbac_reconcilers = {}                 # (account, region, cluster) -> (stop, wake)
+_rbac_reconcile_locks = {}             # (account, region, cluster) -> Lock
+_EKS_RBAC_RECONCILE_INTERVAL = float(
+    os.environ.get("EKS_RBAC_RECONCILE_INTERVAL", "2")
```
Nahuel990, line 101: I don’t see this var being used a plain 2 sec should be enough (without the env var)...  The env var isn’t needed, but the wait maybe is

ministack/services/eks.py
```diff
@@ -402,6 +428,8 @@ def _k3s_run_kwargs(
         "--disable=traefik,metrics-server,servicelb",
         "--tls-san=0.0.0.0",
         "--https-listen-port=6443",
+        "--kube-apiserver-arg=authentication-token-webhook-config-file=/etc/rancher/k3s/eks-auth-webhook.yaml",
+        "--kube-apiserver-arg=authentication-token-webhook-cache-ttl=5m",
```
Nahuel990, line 436: the apiserver has 2m by default, here should be 2m too

ministack/services/eks.py
```diff
@@ -440,6 +468,55 @@ def _k3s_run_kwargs(
...
+    works for both a shared Docker network and a host-run gateway.
+    """
+    from ministack.core import tls as _tls
+
+    host = _k3s_gateway_host(client, ms_network)
+    port = os.environ.get("GATEWAY_PORT") or os.environ.get("EDGE_PORT") or "4566"
+    scheme = "https" if _tls.use_ssl_enabled() else "http"
+    account_id = account_id or get_account_id()
+    region = region or get_region()
+    server = f"{scheme}://{host}:{port}/eks-auth/{account_id}/{region}/{cluster_name}"
```
Nahuel990, line 503: EKS keeps this webhook inside the control plane

Nahuel990 (approved):

Issue #661, opened 2026-05-17
RFC: Unify service state contract — derive everything from `SERVICE_REGISTRY`


jgrumboe:
## Problem

Adding a new service to ministack currently requires touching multiple places in `app.py`:

1. `SERVICE_REGISTRY` — the service entry itself
2. `_state_map` — hand-maintained dict for persistence save
3. `_extra_reset_modules` — tuple for modules that need reset but aren't directly routable
4. The eager-import list in `_load_persisted_state()` — for sub-modules that self-restore at import time

Miss any of these and you get silent bugs: state not saved on shutdown, state not restored on warm-boot, or state not cleared on reset.

On top of that, `ses.py` has ~330 lines of inline v2 code that duplicates `ses_v2.py`, and `pipes` lives outside `SERVICE_REGISTRY` entirely despite being a stateful service. That's relevant further down.

## Proposal

Make `SERVICE_REGISTRY` the single source of truth. Everything else derives from it automatically. Two changes get us there:

### 1. Uniform state contract for every module

Every service module implements three functions:

```python
def get_state() -> dict:
    """Return current state for persistence save. Empty dict if stateless."""
    return {}

def load_persisted_state(data: dict):
    """Restore state from persisted data. Called by app at startup."""
    pass

def reset():
    """Wipe all in-memory state. Called on /_ministack/reset."""
    pass
```

Stateless modules (account, sts, batch, etc.) get no-op implementations. No exclusion lists needed — if a module returns `{}` from `get_state()`, nothing meaningful gets persisted.

### 2. Single restore path — no module-level side effects

Currently we have two restore patterns:
- **Central restore**: app calls `module.load_persisted_state(data)` — used by apigateway, apigateway_v1, servicediscovery
- **Self-restore at import time**: module calls `load_state()` at the top level — used by everything else (sqs, sns, pipes, ses_v2, etc.)

The self-restore pattern is why we need the eager-import hack: sub-modules like `ses_v2` and indirectly-routed services like `pipes` won't be imported until the first request arrives, so their state never gets restored on warm-boot unless we force-import them.

**Proposed: unify on central restore.** `_load_persisted_state()` becomes a single loop:

```python
def _load_persisted_state():
    for mod_name in _state_map:
        data = load_state(mod_name)
        if data:
            _get_module(mod_name).load_persisted_state(data)
```

No eager-import list. No module-level side effects. The app controls restore order. Modules are only imported if they actually have state on disk.

### Registry schema change

[...]

PR #1730, opened 2026-09-14
refactor(persistence): centralize restore through registry (Phase 3)

jgrumboe:
## Summary

Phase 3 of #661: RFC: Unify service state contract — derive everything from SERVICE_REGISTRY.

- Centralized persisted-state restoration through the registry and shared application restore path.
- Preserved service-specific restore behavior, including S3 metadata and lifecycle loader contracts.
- Separated restore from runtime resume and made restore helpers private behind public service loaders.
- Updated persistence coverage to exercise public loaders across registered services.

## Verification

- `git diff --check origin/main...HEAD` — passed.
- `uv run --extra test pytest -q tests/test_persistence.py` — passed.


jgrumboe:
@Nahuel990 how do you like it? Is it mergeable or do you think something is missing?

PR #1731, opened 2026-09-14
test: default RDS and ElastiCache control-plane tests to no-docker

jgrumboe:
## Summary

- Add request-scoped no-Docker handling for RDS and ElastiCache using ContextVars.
- Enable the mode dynamically per annotated test through a botocore request monkeypatch; no startup environment switch is required.
- Default RDS and ElastiCache control-plane tests to the no-Docker path.
- Keep live database/Redis data-plane, TCP connectivity, container lifecycle, and persistence tests Docker-backed.
- No DSQL, README, or test_ministack changes are included.

### RDS

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| RDS test-call time | 179.53s | 4.57s | **97.5% faster (~39×)** |
| RDS-heavy parallel shard | 225.08s | 144.31s | **35.9% faster** |
| Complete CI workflow | 4m54s | 3m37s | **1m17s faster (26%)** |

### ElastiCache

Matched local runs of 68 existing non-connectivity tests:

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| ElastiCache pytest time | 23.45s | 0.75s | **96.8% faster (~31×)** |
| End-to-end test command | 25.89s | 2.37s | **90.8% faster (~10.9×)** |
| Containers started in the control-plane sample | 41 | 0 | **100% fewer** |

The live Lambda-to-ElastiCache connectivity test was excluded from that sample and remains Docker-backed.

## Validation

- ElastiCache suite: **69 passed, 1 skipped** locally.
- Latest CI run: [34963936855](https://github.com/ministackorg/ministack/actions/runs/34963936855) — all jobs passed.
- `git diff --check` passed.


jgrumboe:
## ElastiCache performance

Matched local runs of the 68 existing non-connectivity ElastiCache tests (excluding the new request-scope regression test):

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| ElastiCache pytest time | 23.45s | 0.75s | **96.8% faster (~31×)** |
| End-to-end test command | 25.89s | 2.37s | **90.8% faster (~10.9×)** |
| Containers started in this sample | 41 | 0 | **100% fewer** |

The baseline was run from the parent revision with Docker-backed ElastiCache behavior. The PR run used the request-local no-Docker path.

The live test_elasticache_lambda_network_connectivity test was intentionally excluded from both runs and remains Docker-backed in CI, so the full ElastiCache suite still exercises real container connectivity.

iot-rocket:
EDIT:
@jgrumboe
I re-checked the head (`c3d4532`) after your note on #1742.

The split makes sense. On my dev server the 553 tests that get the no-Docker header (406 in `test_rds.py`, 64 in `test_rds_data.py`, 8 in `test_rds_iam_plugin.py`, 6 in `test_rds_mysql_compat.py`, 69 in `test_elasticache.py`) passed sequentially and with three xdist workers. I did not run the live container tests.

Four points:

1. ElastiCache restore. `_ensure_live_containers` runs at the top of every request. Inside a no-Docker request it discards each pending entry, `_spawn_redis_container` gets `None` from `_get_docker()` and returns the fallback `redis:6379`, and the cluster keeps `available` with no container. The next normal request finds nothing left to respawn. Reproduced in-process: restore one cluster, call `_ensure_live_containers` with the context set, and the pending set is empty while the node endpoint reads `redis:6379`. An early return in `_ensure_live_containers` when the context is set keeps the entries for the next Docker-enabled request.

2. In-process calls. `test_rds_handle_request_describe_with_json_body` calls `_create_db_instance()` directly, so the header never reaches it. With `docker.from_env` replaced by a recording fake, the test still called `containers.run` (name `ministack-rds-...-instance-inproc-json-...`, image `postgres:15-alpine`) from a `ministack-rds-start-*` thread inside the pytest process, and passed anyway because it reads the status before that thread finishes. Same as before the PR, but the marker now says the file runs without Docker. The other in-process tests in the file monkeypatch `_get_docker`; this one should do the same. Setting the context var from the autouse fixture instead would break the two new request-scope tests, which set `_docker` directly and expect `_get_docker()` to return it after the reset.

3. Live coverage. With the marker in place, 19 tests in the two suites stay on the Docker path. In CI, 18 of them skip: 17 need `DOCKER_NETWORK`, which no job sets, and the ElastiCache connectivity test skips the same way. That was also the case before the PR; the two Lambda connectivity tests have not run in CI so far. What changes is the rest: before, every RDS create in CI started a container, after, one test does (`test_rds_create_db_instance_returns_before_container_is_ready`). If that is fine for the default run, a separate job with `DOCKER_NETWORK` set and `.[test,full]` installed (the 17 live tests need psycopg2 and pymysql, which the `test` extra does not carry) would keep the container path covered, as a required check or nightly. I have not measured what `DOCKER_NETWORK` has to be for a server that runs on the runner host; `bridge` is my guess.

4. #1742. `test_rds_restore_state_respawns_docker_container` still needs its own fix. It waits until the fake `containers.run` was called and then asserts `available` at once; the status flips on the runner thread after the call returns. A 300 ms delay at that point fails it with `'creating' == 'available'` (`test_rds.py:5050`). The cluster variant polls for the final status and joins the threads; the single-instance test should do the same.

One question: the two headers are honoured for any client on any deployment. Is that intended, or should they only work when the test harness talks to the server?

jgrumboe:
Thanks for your feedback @iot-rocket.
This PR is AI generated and yes, there are for sure some tests that would need real containers but are marked incorrectly now; AI doesn't get everything right from the start 😉 

The last question is a good one that I haven't thought about. The current implementation came out of the idea to not introduce another env variable again and the wish to turn docker on/off per test via pytest marker.
I think it should be scoped to the test harness only and that  needs more investigation.

jgrumboe:
I'll close this PR as it was a good experiment but not the best design.
New PR with new strategy for CI tests and unnecessary docker-call optimization is coming.

PR #1735, opened 2026-09-15
Poc/startup time

Nahuel990:
Full AI generated PoC, waiting on proper validation


iot-rocket:
@Nahuel990 
Ran it against our stack. Short version: keep the lazy version lookup, the CloudFormation import behind the path check and the deferred listener start. Drop the P-256 switch. Give the listener a readiness signal. Numbers are from container boots.

The GIL claim in the docstrings does not hold on the cryptography the image ships (50.0.1 in both images). A worker thread generating RSA-2048 keys back to back leaves the main thread's 1 ms sleep loop at p99 1.8 ms, same as idle. `_mtls_start_locked` on main already mints in `asyncio.to_thread`, so with the listener start in a background task RSA keygen never touches the loop. P-256 saves about 80 ms of thread time on a first boot without persistence and nothing on a warm one. In exchange the Local CA and the broker certificate become ECDSA, while AWS presents RSA (ATS endpoints chain to Amazon Root CA 1): a device TLS config restricted to RSA fails here and passes on AWS. Both our clients (aws-crt on OpenSSL, and the v1 SDK the fleet runs) connected over the ECDSA chain, so this is a parity point, not a breakage report. If you want to keep the option: `generate_ca` stays at rsa2048 by default, the call site passes ec256, and the GIL sentence goes from both docstrings and the new test.

The deferred start needs a readiness signal. `/_ministack/health` and `/_ministack/ready` answer 200 before 8883 is bound, and consumers read a refused connect as "no listener" (our smoke does; that was the documented behaviour before 1.4.21). With a 5 ms poll I caught it in 1 of 10 boots of the full build (health 200, handshake refused, bound 31 ms later) and in 3 of 10 on the preview build under load. Suggestion: `iot_mtls: starting|listening|degraded|disabled` in the health JSON, and `/ready` waits for the task. Then `_wait_mqtt` in the tests can go and consumers keep one endpoint.

Boot numbers, 5 cold plus 5 warm boots per arm, 1.5.12-full against your branch built with Dockerfile.full so both are glibc full (the preview build is the alpine light edition and boots slower than either, so I left it out). Time to health, median: 1678 ms base vs 1736 ms PR cold, 1936 vs 1727 warm, ranges overlap. RSS at 10 s: 43 vs 45 MiB. The PR moves the listener behind the HTTP port; it does not move time to health or memory. What does: the image installs with `pip --no-compile` and /opt/ministack has no pyc, so every fresh container compiles the tree on import (`import ministack.app` 609 ms, 149 ms once a `__pycache__` exists). A layer with `python3 -m compileall -q /opt/ministack /usr/local/lib/python3.13/site-packages` on top of 1.5.12-full boots to health in 1047 ms cold, 1199 warm, RSS 32 MiB. That is the startup PR I would do.

Small things: `global _iot_mtls_task` sits inside an else branch, hoist it; title, commit messages and a CHANGELOG bullet are still in PoC state.

Validation: `tests/test_router.py`, `test_iot.py`, `test_iot_data.py` 1074 passed on the PR vs 1071 on efddc4a8, 0 failures either side, each arm on its own server from its own tree. Against the PR image built as full: our device mTLS smoke with the modern and the legacy client, the device full chain (binary, IoT rule, Lambda, DynamoDB) 7 passed plus one scenario that only runs on our canonical instance, backends smoke 11 passed.

PR #1753, opened 2026-09-16
fix(multitenancy/s3): S3 PostObject ignores the account in its form credentials

ihmpavel:
A presigned POST signs in the multipart form fields (`x-amz-credential`), but `extract_access_key_id` reads only the Authorization header and the query string, so the request is never scoped and `_post_object` drops the field. Browser form uploads therefore always run in 000000000000, while the SDK and presigned PUTs from the same caller follow the 12-digit key.

To simplify: `PostObject` fails with `404 NoSuchBucket` when using multitenancy


Nahuel990 (approved):

Nahuel990:
@ihmpavel thanks, I'm merging this, please next time review Contributing.md

PR #1759, opened 2026-09-17
ci: separate control-plane and data-plane tests

jgrumboe:
## Summary

- Add a centralized `MINISTACK_DOCKER_ENABLED` gate, enabled by default.
- Apply the gate to Docker-backed services, DSQL preflight, and startup/periodic cleanup.
- Mark live-container tests as `data_plane` and keep them out of control-plane shard planning.
- Add a dedicated Docker-enabled serial CI job with an isolated network and cleanup.
- Update test commands and environment-variable documentation.

## Verification

- `13 passed` — `tests/test_docker_gate.py`
- Strict marker collection and test-shard planning pass.
- Compilation, workflow YAML parsing, and `git diff --check` pass.

The live Docker data-plane suite is intended to run in CI.


jgrumboe:
@Nahuel990 What's your opinion about introducing a `MINISTACK_DOCKER_ENABLED` variable? Default to true doesn't break anything. Explicitly setting it in CI gives us a good separation between control-plane only and data-plane tests.
PR is not ready yet, still some tweaks here and there but I'd like to get early feedback from you about the strategy itself.

jgrumboe:
@iot-rocket Wdyt?

iot-rocket:
@jgrumboe yes, sure :)

So generally looks quite good to me!

2 things:
- RDS: with the gate off, no CI test do has the container full lifecycle anymore (stop/start/reboot/replica/failover). Only the Lambda network connectivity test has the `data_plane`, so create and connect are covered, but the others not. Saw the TODOs on the Aurora tests, but maybe a couple of basic marked lifecycle tests there?
- Tests are reading `MINISTACK_DOCKER_ENABLED` from their own env and probe their own Docker daemon to decide what the server can do -- If e.g. `/_ministack/health` could report the effective server capabilities, one conftest skip could replace `requires_docker` and the `LAMBDA_EXECUTOR`, `DOCKER_NETWORK` and `OPENSEARCH_DATAPLANE` skip conditions?

Some more minor points/ideas:
- The `PARALLEL_SHARDS`/`SERIAL_SHARDS` env are still there next to the hardcoded matrix. Bump one without the matrix and the planner builds a shard nobody runs, CI should stay green
- Gate is pasted into ten `_get_docker()` copies, eks/lambda_svc additionally at import time. One central `core.docker.get_client()` could e.g. cover that
- The two cleanup steps in the data-plane job can fail a green run if the server misses the 30s SIGTERM window or a container survives (`network rm` with an endpoint attached). OK if that is meant as a leak check ;) - otherwise a fallback cleanup of the containers on the job's network
- No `timeout-minutes` on any job, so a hung run should wait  for the GitHub default. Something like 20min could help?
- `OPENSEARCH_DATAPLANE=1` is around 26s for one test - the RDS replication flag has no marked test yet : `psycopg2`/`pymysql` should be already available via the `full`/`dev` extras
- Serial job is now the long part (3m06 vs 2m32). The planner reads `SERIAL_SHARDS`, it should just need a second matrix entry and both counts are getting bumped together

jgrumboe:
@iot-rocket 
Thanks — great review. I’m keeping this PR focused on the CI/Docker test-plane framework rather than expanding service coverage.
- I’ve intentionally deferred the extended RDS/Aurora lifecycle cases to a follow-up coverage PR.
- The control-plane matrix is static and each runner selects its own shard directly, so there is no longer a separate planning job or duplicate shard-count configuration to drift.
- Added job timeouts, split the serial lane into two shards, and removed the unused RDS replication flag.
- Data-plane cleanup now writes leak details to the workflow summary, force-removes attached containers, retries network cleanup, and does not fail an otherwise-green run.
- The repeated service-local Docker client construction and a health-capabilities/conftest skip model are worthwhile follow-ups, but broader than this PR.
- The small Lambda response-stream compatibility fix: it is needed for an already-existing data-plane test that otherwise fails with HTTP 502; no new test coverage was added.
Thanks again for the detailed notes.

iot-rocket:
@jgrumboe
trying by best ;)
Yeah, `/_ministack/health` was just an idea to prevent another not well-received environment variable

jgrumboe:
@Nahuel990 That would be my proposal to lift CI on the next level 😉

jgrumboe:
@Nahuel990, nothing more to add here. Can we get that merged?

Nahuel990 (approved):

PR #1769, opened 2026-09-17
Enable configured TLS for PostgreSQL containers

### LINKED ISSUE #1768: Support verified TLS for real-backed PostgreSQL
### Affected Service
RDS

### What happened?
Real-backed PostgreSQL starts with TLS disabled. Applications that require an encrypted connection cannot connect, including Lambda handlers that use PostgreSQL drivers with certificate and hostname verification.

The current RDS implementation selects `postgres:<major>-alpine` and supplies database credentials, but does not configure a server certificate or `ssl=on`. This differs from stock MySQL 8.0, which initializes its own TLS certificates. MySQL's image defaults do not establish a MiniStack-managed CA or hostname-verification contract.

### Reproduction steps
Create a PostgreSQL instance using the README's RDS example, then connect to its returned endpoint:

```python
psycopg2.connect(
    host=endpoint["Address"], port=endpoint["Port"],
    user="admin", password="password", dbname="appdb",
    sslmode="require",
)
```

The stock PostgreSQL backing image reports `ssl=off`; a TCP client with `sslmode=require` fails with `server does not support SSL, but SSL was required`. This was also reproduced in a fresh disposable `postgres:16-alpine` container using the same credentials-only startup configuration.

### Expected behavior and proposed scope
Provide an opt-in way to supply PostgreSQL server certificate/key files and enable TLS for standalone instances and Aurora PostgreSQL backing containers, including streaming readers. Copy material over the Docker API so MiniStack-in-Docker does not require the same paths on the Docker host. Keep the private key outside the database directory and restrict its permissions.

Document that certificates must cover the advertised endpoint, how clients receive the issuing CA, and how existing Lambda environment settings / `LAMBDA_DOCKER_FLAGS` can supply client trust. Test successful CA-and-hostname verification, rejection of wrong CA/hostname, restart and cleanup behavior, and unchanged MySQL behavior.

Non-goals: a new CA service, automatic root-store modification, production certificate management, IAM authentication, TLS enforcement for all existing clients, or changes to MySQL. Existing plaintext clients would remain compatible. No new dependencies or Docker image changes are planned; this issue also records scope before any infrastructure changes should they prove necessary under CONTRIBUTING.md.

### MiniStack version
v1.5.13 / main at 551c42860ce49246866ec57a5fc4257932a3d193 (source inspection); backing-image reproduction with PostgreSQL 16.15.

### SDK / tool used
Docker CLI and PostgreSQL psql 16.15

jayjanssen:
## Why

PostgreSQL backing containers start without TLS, so clients that require verified encryption cannot connect. Stock MySQL supplies its own TLS defaults; PostgreSQL needs explicit server material. Closes #1768.

## What

Add opt-in `MINISTACK_RDS_PG_SSL_CERT` / `MINISTACK_RDS_PG_SSL_KEY` settings for standalone, cluster, and streaming-reader PostgreSQL containers. Document hostname coverage and client/Lambda CA trust through existing settings.

## How

Validate the PEM pair, copy it over the Docker API before startup, and enable PostgreSQL TLS. Keep the private key outside database volumes with mode `0600`. Failed TLS launches do not fall back to plaintext or report standalone instances as available. No dependencies, Dockerfiles, or Lambda runtime changes.

## Risk

Unconfigured PostgreSQL and MySQL keep their existing launch path. Configured servers still accept plaintext clients. Retained containers need recreation for certificate rotation. Existing endpoint limitations remain: streaming readers advertise IPs, and cluster stop/start can replace the advertised DNS name with an IP; certificates must match the name or IP clients actually use.

## Testing

- Ran `MINISTACK_TEST_RDS_TLS_DOCKER=1 pytest tests/test_rds_tls.py -q` with `MINISTACK_TEST_RDS_TLS_IMAGE` set to `postgres:11-alpine`, `postgres:16-alpine`, and `postgres:18-alpine`. Verified TLS succeeds, wrong CA/hostname fail, plaintext remains compatible, and data/certificates survive restart. Tests own and clean up their Docker resources.
- In an isolated MiniStack gateway on a user-defined Docker network, created a cluster, issued a test certificate for its returned names, then created its first member. `psql` with `sslmode=verify-full` and the test CA reported TLS 1.3 via `pg_stat_ssl`. A local `provided.al2023` Lambda bootstrap written in Go also verified the server using default system roots plus the CA directory in `SSL_CERT_DIR`.
- A streaming reader negotiated verified TLS and reported `pg_is_in_recovery() = true` using a certificate-covered hostname with explicit `hostaddr`. This does not claim that the existing IP-only reader endpoint matches a DNS-only certificate.

Generated with Codex


jayjanssen:
CI triage: the only test failure is `tests/test_cognito.py::test_oauth2_full_flow_for_non_boot_region_pool`, which timed out after 10 seconds waiting for `/oauth2/token` in parallel shard 2. That shard otherwise had 2874 passing tests; the other parallel shards, serial shard, and lint passed. No Cognito or Lambda code changed in this PR. The exact failing test passed six consecutive isolated local runs against the PR source. This appears to be a parallel-load timeout, though a rerun is needed to confirm.

Could a maintainer rerun the failed CI jobs? `gh run rerun 35272108002 --failed` was refused because my account lacks repository admin rights. I have not changed unrelated tests or relaxed their timeout to conceal the failure.

jayjanssen:
Closing as superseded by the automatic RDS PostgreSQL TLS implementation in #1774, released in v1.5.14. Thanks for wiring this up, and for the clarification in https://github.com/ministackorg/ministack/issues/1768#issuecomment-5749344301.

We have now tested the published `1.5.14-full` image (manifest digest `sha256:29d91a1c766ec8fc61a8f5e5df3836b2576b27da0c30440f5a3533780fef69e8`) with our actual SQL gateway through both Data API and the Go Lambda executor. Both routes passed, including persisted data and stable DNS endpoints across RDS stop/start. The executor used hostname-verified TLS with the public CA from `/_ministack/rds/ca.pem` installed in its normal trust store; no TLS bypass was used.

The released automatic-certificate approach replaces the manual certificate/key configuration proposed here, so keeping this alternative implementation open is no longer necessary.

We did find a separate edge case in the replacement: using a generated endpoint longer than 64 bytes as the certificate common name raises the X.509 CN-length error (our normal infrastructure fixture produced an 80-byte endpoint). Shorter fixture names passed verified TLS. That remaining issue should be handled separately rather than reviving this superseded implementation. This validation does not claim gateway-process restart/CA persistence, streaming-replica, Docker Lambda, or Aurora coverage.

PR #1773, opened 2026-09-18
fix(rds-data): correct PostgreSQL SELECT row counts

jayjanssen:
## Why

PostgreSQL reports returned rows in `cursor.rowcount` for SELECT. MiniStack copies that value into `numberOfRecordsUpdated`, so a SELECT that returns one row incorrectly reports one updated record.

## What

Report zero updated records when the PostgreSQL command tag is SELECT. Keep returned records and column metadata unchanged.

## How

Check the first token of `cursor.statusmessage` for PostgreSQL engines only. Preserve other command counts, including INSERT/UPDATE/DELETE with RETURNING, and keep the negative-count clamp. Do not infer affected counts from SQL text or result columns.

## Risk

This is a narrow response correction; MySQL behavior stays unchanged. A SELECT tag is not a guarantee that a statement has no side effects, including writes inside CTEs.

## Testing

No live database testing. An independent read-only review found no blocking issues. Local regression validation loaded the unmodified handler from `git show HEAD:ministack/services/rds_data.py` before the fix was committed: four nonempty SELECT cases failed as expected. The fixed handler passed with socket connections blocked and `MINISTACK_ENDPOINT=http://127.0.0.1:1`.

Generated with Codex


jayjanssen:
CI triage: the only test failure is `tests/test_cognito.py::test_oauth2_full_flow_for_non_boot_region_pool`, which timed out after 10 seconds waiting for `/oauth2/token` in parallel shard 2. That shard otherwise had 2866 passing tests; all five new PostgreSQL row-count test groups passed, as did the other parallel shards, serial shard, and lint. No Cognito or Lambda code changed here.

This matches the failure reported on #1769 (https://github.com/ministackorg/ministack/pull/1769#issuecomment-5721179614), where a maintainer rerun passed on attempt 2. That is evidence of a similar intermittent failure, not proof of this run’s root cause.

Could a maintainer rerun the failed CI jobs for https://github.com/ministackorg/ministack/actions/runs/35361329905? I have not changed unrelated tests or relaxed their timeout.

Nahuel990:
@jayjanssen sorry the wrong test result is cached in github, I will take a look during the weekend, dont worry, if tests passed locally we will merge it if the shape is correct, we've been delayed reviewing some merged PRs that contained incorrect behaviour

Nahuel990 (approved):

PR #1775, opened 2026-09-19
fix(eventbridge): resolve dynamic API destination HTTP parameters

dmgarland:
API Destination `HttpParameters` currently treat JSON paths as literal strings. For example, a topic subscription target with `PathParameterValues: ["$.detail.fullDocument.payload.topicKey"]` sends that expression in the URL instead of the event's `venue:1231245` value.

This resolves whole JSON paths in path parameters, header values and query-string values against the original event envelope. It supports field access, array indexes and wildcards; preserves literal values and connection precedence; and builds new parameter containers so one delivery cannot change the saved target or a later delivery. Request-body transformation continues independently.

### AWS verification

Verified against AWS EventBridge in eu-west-1 using MongoDB Atlas outbox inserts → an existing partner bus → isolated API Destination targets → a temporary API Gateway/Lambda HTTPS recorder. AWS setup, inspection and cleanup used the AWS CLI. Test infrastructure and database records were removed afterwards.

Observed AWS behaviour captured in the tests:

- A dynamic topic path and `topicKeys[0]` resolve before an input transformer removes those fields from the body.
- Header and query values resolve from the same original event; connection values take precedence.
- Missing header/query fields render as empty strings; numeric, boolean and null values render as `42`, `false` and `null`.
- An object renders as `{key:value}` and an array wildcard as `[artist:6789,venue:123]`.

References: [dynamic path parameters](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-targets.html#dynamic-path-parameters), [HttpParameters](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_HttpParameters.html).

### Validation

- Six new HTTP capture regression cases fail on the unmodified implementation with the literal JSON-path URL.
- The same tests pass with the fix, covering direct/indexed paths across `Input`, `InputPath`, and `InputTransformer`, and two successive events per stored target.
- `python -m pytest tests/test_eventbridge.py --tb=short -q`: **249 passed**.
- `ruff check ministack/` and `ruff check tests/test_eventbridge.py`: passed.

A separate AWS observation: DELETE API Destination requests arrived without a body, whereas a direct DELETE control to the same recorder preserved it. HTTP method/body behaviour is outside this parameter-resolution fix.


Nahuel990:
@dmgarland can you state which sdk are you using?

dmgarland:
We use AWS SDK for JavaScript v3 (`@aws-sdk/client-eventbridge` 3.1111.0) to publish local events, and Python boto3 to configure the local rules and API destinations. The regression tests in this PR use the existing boto3 fixtures.

I also verified the dynamic HTTP parameter behaviour against real AWS using AWS CLI 1.22.34 (botocore 1.23.34) to configure the rules and API destinations. For that verification, events originated from a MongoDB Atlas trigger into an EventBridge partner bus.

Is there a particular client call or version you would like me to check?

Nahuel990:
@dmgarland no, it’s okay. We’ve been receiving a lot of PRs claiming “AWS verification” that aren’t actually verifying against AWS. Since this is your first contribution, I’ll check the PR against AWS myself to make sure the behavior matches what you’re claiming

dmgarland:
That's amazing thanks @Nahuel990 would be great to get a second opinion on this

Nahuel990 (approved):

dmgarland:
:tada:

